PR #6, opened 2018-10-19
change code structure

leonjia0112:
The the current structure
src/
├── common.rs
├── main.rs
└── tpm.rs
common and tpm are two parallel module that can access each other and can be accessed by the main function


leonjia0112:
It needs more test, but this skeleton is good for the first release.
Another crypto.rs will be included in parallel follow this structure as well.

frozencemetery:
Needs to pass Travis.  Also please clean up the commit history.

frozencemetery:
(Superseded by #7)

PR #9, opened 2018-10-30
Finished adding quote resquest handling functions

leonjia0112:


leonjia0112:
Why does it keep complaining the diff with no diff?
-                                    common::IMA_ML_STUB.to_string()
+                                    common::IMA_ML_STUB.to_string()

frozencemetery:
FWIW, I think it was probably a whitespace (tabs vs. spaces) issue, but glad you got it resolved.

PR #12, opened 2018-11-06
Add key derivation function binding and appropriate tests

mbestavros:
As the title suggests, this PR adds a binding for the PBKDF2-HMAC function in OpenSSL and configures it to match the [implementation in Python-Keylime](https://github.com/mit-ll/python-keylime/blob/2da18bd4814fb3c93d89175d2d5172b1167753c2/keylime/crypto.py#L83). An appropriate test is also added.

A few notes:

- This is also mentioned in the function documentation, but it bears repeating: The only reason this binding uses SHA-1 is because Python-Keylime uses it at the moment. PyCryptodome [defaults to SHA-1 as a PRF](https://pycryptodome.readthedocs.io/en/latest/src/protocol/kdf.html) if the `prf` parameter is not specified, and Python-Keylime does not provide that parameter in its KDF binding.

- The key the test compares against was generated from the Python-Keylime KDF binding using the same password and salt.

Tagging @frozencemetery and @leonjia0112 for review.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -2,17 +2,19 @@ extern crate hex;
 extern crate openssl;
 
 // use super::*;
+use openssl::error::ErrorStack;
 use openssl::hash::MessageDigest;
+use openssl::pkcs5;
 use openssl::pkey::PKey;
 use openssl::sign::Signer;
 use std::string::String;
 
 /*
- * Input: secret key
- *        message to signed
- * Output: signed mac result
+ * Inputs: secret key
+ *        message to sign
+ * Output: signed HMAC result
  *
- * Sign message and return hmac mac string
+ * Sign message and return HMAC result string
  */
 pub fn do_hmac(input_key: String, input_message: String) -> String {
     let key = PKey::hmac(input_key.as_bytes()).unwrap();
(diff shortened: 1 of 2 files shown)


leonjia0112:
@frozencemetery 
I think you can try to delete all white space and type in space again. Otherwise, you can try rustfmt, that would so solve this problem for you.

mbestavros:
@frozencemetery: Made a few small style changes; CI build looks like it's working. `cargo` is complaining that the call to `pbkdf2_hmac` isn't handling errors correctly, though it's just a warning at the moment. Not sure of the best way to handle that, though.

mbestavros:
@frozencemetery: Got it; per our offline discussion, I made `kdf` and `do_hmac` pass the `Result` returned from OpenSSL down the line until it can be appropriately handled. Figured the slight context-switch deserved its own commit.

Let me know what you think.

mbestavros:
@frozencemetery: Done, commits are properly separated now.

mbestavros:
@frozencemetery Fixed; the call to HMAC from within `main.rs` now should fail gracefully. I used the `match` pattern; if there's a better way, please let me know.

PR #13, opened 2018-11-07
Test tpm function and bug fix

leonjia0112:
I added a few test case to make sure the tpm function is working. Skipped quote function test, because I haven't figured out how to test this function. Also added some comments for each function to make clear what it does. Fixed some bug in tmp functions, now it is able to execute tpm command and getting the result back from tpm. 

But it won't pass the CI since CI doesn't contain tpm with it.

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -44,6 +44,7 @@ struct TpmData {
 ftpm_initialize.py
 Following are function from tpm_initialize.py program      
 *****************************************************************/
+
 /*
 input: content key in tpmdata
 return: deserlized json object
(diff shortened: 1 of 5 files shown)


frozencemetery:
Please post a comment once this is passing CI and ready for review.

leonjia0112:
Removed unnecessary changes on README.md file and test.json file.

src/tpm.rs
```diff
@@ -396,13 +468,51 @@ fn read_file_output_path(output_path: String) -> std::io::Result<String> {
...
+
+    #[test]
+    fn test_read_file_output_path() {
+        assert_eq!(
+            read_file_output_path("test_input.txt".to_string()).unwrap(),
+            "Hello World!\n"
+        );
+    }
 
+    //Comment out for tpm unit testing
```
frozencemetery: Actual TPM tests are good and we should have them.  Rather than commenting them out, please make running them conditional on whether we have the capability to do so.

leonjia0112: For the test condition, I found out there are two ways to do. One is to let the problem test the system capability for those command using the same "cargo test" command. The other one is let the user change the running condition. For example, if the system has the tpm run the test 'cargo test --feature 'tpm_ok' ', otherwise run 'cargo test', the tpm test case will not execute with normal cargo test.

This is the implementation of auto command detection. It is ready for review.

PR #15, opened 2018-11-15
Add secure mount function

leonjia0112:
1. contains unsafe libc function

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -20,6 +20,15 @@ pub static IMA_ML_STUB: &'static str =
 pub static IMA_ML: &'static str =
     "/sys/kernel/security/ima/ascii_runtime_measurements";
 pub static KEY: &'static str = "secret";
+pub static WORK_DIR: &'static str = "/tmp";
+
+/*
+ * Temporaray location for configuration parameters
+ */
+
+// cloud node
+pub static SECURE_SIZE: &'static str = "1m";
+pub static MOUNT_SECURE: bool = true;
 
 /*
  * Input: Response status code
@@ -97,6 +106,35 @@ pub fn get_restful_parameters(urlstring: &str) -> HashMap<&str, &str> {
         parameters.insert(list[x], list[x + 1]);
     }
     parameters
+}
+
+/*
+ * Input: path directory to be changed owner to root
+ * Return: Result contains execution result
+ *         - directory name for successful execution
+ *         - -1 code for failure execution.
+ *
+ * If privilege requirement is met, change the owner of the path to root
+ * This function is unsafely using libc. Result is returned indicating
+ * execution result.
+ */
+pub fn chownroot(path: String) -> Result<String, i32> {
+    unsafe {
+        // check privilege
+        if libc::geteuid() != 0 {
+            error!("Privilege level unable to change ownership to root for file: {}", path);
+            return Err(-1);
+        }
+
+        // change directory owner to root
+        if libc::chown(path.as_bytes().as_ptr() as *const i8, 0, 0) != 0 {
+            error!("Failed to change file {} owner.", path);
+            return Err(-1);
+        }
+
+        info!("Changed file {} owner to root.", path);
+        Ok(path)
+    }
 }
 
 // Unit Testing
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,6 +21,7 @@ extern crate tempfile;
 
 mod common;
 mod crypto;
+mod secure_mount;
 mod tpm;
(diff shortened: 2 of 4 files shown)


leonjia0112:
Added secure mount function in a separated secure mount file. Test passed, ready for a review.

src/crypto.rs
```diff
@@ -42,6 +42,7 @@ pub fn do_hmac(
  * PBKDF2 function defaults to SHA-1 unless otherwise specified, and
  * Python-Keylime uses this default.
  */
+#[allow(dead_code)]
```
frozencemetery: Please don't add these.

src/common.rs
```diff
@@ -20,6 +20,15 @@ pub static IMA_ML_STUB: &'static str =
 pub static IMA_ML: &'static str =
     "/sys/kernel/security/ima/ascii_runtime_measurements";
 pub static KEY: &'static str = "secret";
+pub static MOUNT_SECURE: bool = true;
```
frozencemetery: This shouldn't be a global constant.

src/common.rs
```diff
@@ -99,6 +108,24 @@ pub fn get_restful_parameters(urlstring: &str) -> HashMap<&str, &str> {
...
+ *
+ * If previledge requirement is met, change the owner of the path to root
+ * This function is unsafely using libc
+ */
+#[allow(dead_code)]
+pub fn chownroot(path: String) {
+    unsafe {
+        let euid = libc::geteuid();
+        if euid == 0 {
+            libc::chown(path.as_bytes().as_ptr() as *const i8, 0, 0);
```
frozencemetery: Handle return code on this.

src/crypto.rs
leonjia0112: Without this would cause function unused warning if that is not an issue.

leonjia0112: Got it. Thanks.

src/common.rs
leonjia0112: Moved to local constant.

src/common.rs
leonjia0112: A status code is returned to caller with execution result log.

leonjia0112:
Test passed, ready for a review. Thanks.

src/common.rs
```diff
@@ -99,6 +107,42 @@ pub fn get_restful_parameters(urlstring: &str) -> HashMap<&str, &str> {
...
+/*
+ * Input: path directory to be changed owner to root
+ * Return: status code, 0 for successful execution, other wise fail
+ *
+ * If privilege requirement is met, change the owner of the path to root
+ * This function is unsafely using libc. Status code is return indicating
+ * execution result.
+ */
+pub fn chownroot(path: String) -> Option<i32> {
+    unsafe {
```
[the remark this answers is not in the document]
leonjia0112: Thanks for your review. 

I changed the error handling in chownroot() function by return a result<> that could to be handled by the caller. The mount() function is also changed to return a result<> for error handling. Also, early return is implemented. Err is returned for error result.

The chownroot() function is actually used in multiple places in keylime_node, including functions in Common.rs and secure_moutn.rs.

Secure_mount is used for checking the mount status of the secure directory. It makes sure there is a tmpfs for keylime node to store keys, which is /var/lib/keylime/secure/ directory normally.

leonjia0112:
Changed based on the request. Test passed, and ready for another review. Thanks.

I changed the commit message and added some inline comments for the code. Hope these could make it clearer about what the change is used for.

src/common.rs
```diff
@@ -99,6 +108,33 @@ pub fn get_restful_parameters(urlstring: &str) -> HashMap<&str, &str> {
...
+ * If privilege requirement is met, change the owner of the path to root
+ * This function is unsafely using libc. Result is returned indicating
+ * execution result.
+ */
+pub fn chownroot(path: String) -> Result<String, i32> {
+    unsafe {
+        if libc::geteuid() != 0 {
+            error!("Privilege level unable to change ownership to root for file: {}", path);
+            return Err(-1);
+        } else {
```
frozencemetery: Close.  Use more early exit  The flow should be like:

```rust
if getuid() != 0 {
    return;
}
if chown() != 0 {
    return;
}
Ok()
```

This keeps the branching manageable, which is easier to read and reason about.

leonjia0112:
Implemented request changes. Test passed. Ready for a review. Thanks.

frozencemetery:
Change merged.  Please take a look at the rewritten commit message.

PR #16, opened 2018-11-26
Add tpmdata file io functions for keylime to read/write tpm data

leonjia0112:
Added four functions for read/white/get/set tpmdata.json file where tpm data is stored in tpm.rs. Also added a tpmdata_test.json for testing purpose.

* Changed .travis.yml to single thread testing, because parallel testing causes the file changed by other tests and fail.
* Modified tpmdata read function with error handling without panic.

Compare to Lincoln lab version:
Original python version used a global variable tpmdata to store tpmdata in global scope. This implementation eliminated global variable is to store tpm data since it is not a frequent access variable.

## Test passed. Ready for a review. Thanks.


leonjia0112:
Test passed. Ready for another review. I changed Travis back to what it was, and now all the tests should pass without fail. Thanks.

src/tpm.rs
```diff
@@ -206,9 +272,9 @@ pub fn create_deep_quote(
     mut vpcrmask: String,
 ) -> Option<String> {
     let quote_path = NamedTempFile::new().unwrap();
-    let key_handle = get_tpm_metadata("aik_handle".to_string());
-    let aik_password = get_tpm_metadata("aik_pw".to_string());
-    let owner_password = get_tpm_metadata("owner_pw".to_string());
+    let key_handle = get_tpm_metadata_content("aik_handle").unwrap();
```
frozencemetery: Do not call `unwrap` or `panic!` outside the test suite.

src/tpm.rs
```diff
@@ -26,50 +28,114 @@ pub const EXIT_SUCCESS: i32 = 0;
...
-    let t_data = read_tpm_data().unwrap();
-    let t_data_json_value = serde_json::to_value(t_data).unwrap();
-    Some(t_data_json_value[content_key].to_string())
+ * Input: content key in tpmdata
+ * Return: Result wrap value string or error code -1
+ *
+ * Getting the tpm data struct and convert it to a json value object to
+ * retrive a particular value by the given key inside the tpm data.
+ */
+fn get_tpm_metadata_content(key: &str) -> Result<Option<String>, i32> {
```
frozencemetery: This type is kind of gross.  I think you can do it with a `Result<String, i32>` (see below).  However, the i32 doesn't seem to be meaningful; all you ever put there is -1.

src/tpm.rs
```diff
@@ -26,50 +28,114 @@ pub const EXIT_SUCCESS: i32 = 0;
...
+            content
+                .as_str()
+                .unwrap()
+                .to_string()
+                .trim_matches(remove)
+                .to_string(),
+        )),
+        None => {
+            debug!("Content field [{}] doesn't exist.", key);
+            Ok(None)
```
frozencemetery: Should be an `Err`, yeah?

src/tpm.rs
```diff
@@ -525,6 +591,54 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_get_tpm_metadata_1() {
+        match set_tpmdata_test() {
+            Ok(_) => {
+                // using test tpmdata.json content must present, system won't panic
```
frozencemetery: Instead of panicing, consider printing a useful error message.

src/tpm.rs
```diff
@@ -525,6 +591,54 @@ mod tests {
...
+                    // using test tpmdata.json content must present, system won't panic
+                    let remove: &[_] = &['"', ' ', '/'];
+                    let password =
+                        get_tpm_metadata_content("owner_pw").unwrap();
+                    assert_eq!(
+                        password.unwrap().trim_matches(remove),
+                        String::from("hello")
+                    )
+                }
+                Err(_) => assert!(false),
```
frozencemetery: Instead of this, I would do something like: `assert!(set_tpmdata_test().is_ok())`

src/tpm.rs
leonjia0112: Handled error and return None if an error occurs.

src/tpm.rs
leonjia0112: Changed to Result<String, String>

src/tpm.rs
leonjia0112: Changed to Err for a little simple implementation.

src/tpm.rs
leonjia0112: Print out an error message explaining what happened.

src/tpm.rs
leonjia0112: This is great.  Changed.

leonjia0112:
Changed based on review:
1. Changed error from -1 to message that explains the detail of the error.
2. Changed Option<String> to String
3. Used is_ok() in testing
4. Remove .unwrap() that panic the system with error handling and early return.

**Test Passed. Ready for a review.**
Thanks.

leonjia0112:
Now there are only tpm.rs and tpmdata_test.json files are changed in this PR.

Note: 
tpmdata struct was removed because it is no longer used by the read_tpm_data() function.
File changed of main.rs and common.rs is because travis is complaining about the formatting, which it wasn't complaining before.

**Test passed, ready for a review**
Thanks

src/tpm.rs
```diff
@@ -26,50 +27,160 @@ pub const EXIT_SUCCESS: i32 = 0;
...
+            error!("Fail to read tpmdata.json.");
+            return Err(e);
+        }
+    };
+
+    let remove: &[_] = &['"', ' ', '/'];
+    match tpm_data.get(key) {
+        Some(content) => Ok(content
+            .as_str()
+            .unwrap()
```
frozencemetery: unwrap is not permitted here

src/tpm.rs
```diff
@@ -26,50 +27,160 @@ pub const EXIT_SUCCESS: i32 = 0;
...
+ * caller for error handling.
  */
-fn read_tpm_data() -> Result<TpmData, Box<Error>> {
-    let file = File::open("tpmdata.json")?;
-    let data: TpmData = serde_json::from_reader(file)?;
+fn read_tpm_data() -> Result<Value, Box<String>> {
+    let file = match File::open("tpmdata.json") {
+        Ok(f) => f,
+        Err(e) => {
+            return Err(Box::new(format!(
```
frozencemetery: Can you make a helper function for this `Err(Box::new(format!()))` gunk?

leonjia0112:
Add a output_error_message() function for Err(Box::new(foramt!(...))) that can take a error message and a error and return a error message string. 

.unwrap()s are removed with error handling Result<>.

** Test passed. Ready for a review.**
Thanks!

frozencemetery:
@leonjia0112 I've made some changes.  Please take a look and if you're satisfied they're okay, I'll merge.

leonjia0112:
Those changes are decent. I am ok with those changes. Thanks :)

PR #18, opened 2018-11-30
Add configuration file support

leonjia0112:
- Add configuration file contains the configuration parameters that can be changed by the user in the configuration file. 
- Removed the global parameters in temporary config location and moved to the configuration file
- Add a function to retrieve date from the config file with unit testing
- Update functions that access configuration parameters
  Note: currently, those functions that access configuration parameter doesn't have a result handling error. So unwrap is used for now. I will change those functions to take care of the error in the future PR.

**Check passed. Ready for a review**


leonjia0112:
I can close this PR for now.
I think the important thing is we need a place to let the user put the registrar IP address and port number.

Issue #19, opened 2018-12-06
Remove all calls that might panic


frozencemetery:
It's not going to be okay for the process to crash, which is what `panic!()` does.  Removal includes all calls to `unwrap()`, `assert!()`, and the like (except in test suite code).

Ideally this work would also include some CI to prevent anything that panics from getting added in new code as well.

frozencemetery:
My suggestion is to add something like [no_panic] for CI purposes.  @leonjia0112 please investigate.

leonjia0112:
Got it.

leonjia0112:
This crate is able to detect panic in rust.
https://crates.io/crates/no-panic

Will apply it in later PR.

frozencemetery:
Resolved enough for now in c0872975525f398d22a1bdee831de81ab38fa7bb

PR #28, opened 2018-12-10
Change return type in tpm.rs

leonjia0112:
Add support for tpm custom error type. This is aim on solving #40 for changing return type.
Implementation details:
1. Implement KeylimeTpmError enum has two contribution
    - error generate from TPM command execution
    - error generate from rust part of TPM interface
2. Implement From trait for Error:
    - `serde_json::error::Error`
    - `std::num::ParseIntError`    
    - `std::string::FromUtf8Error`
    - `std::io::Error`
    - `std::time::SystemTimeError`    
3. Upate return type of `run()` function in a tuple and a custom error
    `Result<(String, String), TpmExecError>`    
4. Update return type of other function in tpm.rs with `KeylimeTpmError` support, which has benefit in code efficient and early return efficient

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,6 +32,7 @@ use hyper::service::service_fn;
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
 use serde_json::Map;
 use std::collections::HashMap;
+use std::error::Error;
 use std::fs::File;
 use std::io::BufReader;
 use std::io::Read;
@@ -368,8 +369,8 @@ fn get_request_handler(
                 common::RSA_PUBLICKEY_EXPORTABLE.to_string(),
                 p.to_string(),
             ) {
-                Some(q) => q,
-                None => {
+                Ok(q) => q,
+                Err(err) => {
                     if let Err(e) = set_response_content(
                         400,
                         "Failed to crate quote from TPM.",
@@ -383,7 +384,7 @@ fn get_request_handler(
                     }
                     return emsg(
                         "TPM error. Failed to create quote from TPM.",
-                        None::<String>,
+                        Some(err.description().to_string()),
                     );
                 }
             };
@@ -396,8 +397,8 @@ fn get_request_handler(
                 v.to_string(),
                 p.to_string(),
             ) {
-                Some(q) => q,
-                None => {
+                Ok(q) => q,
+                Err(err) => {
                     if let Err(e) = set_response_content(
                         400,
                         "Failed to create deep quote from TPM.",
@@ -411,7 +412,7 @@ fn get_request_handler(
                     }
                     return emsg(
                         "TPM error. Failed to create deep quote from TPM.",
-                        None::<String>,
+                        Some(err.description().to_string()),
                     );
                 }
             };
--- a/src/secure_mount.rs
+++ b/src/secure_mount.rs
@@ -122,17 +122,24 @@ fn mount() -> Result<String, Box<String>> {
                         })?;
 
                     // mount tmpfs with secure directory
-                    tpm::run(
+                    if let Err(e) = tpm::run(
                         format!(
                             "mount -t tmpfs -o size={},mode=0700 tmpfs {}",
                             common::SECURE_SIZE,
                             s,
                         ),
                         tpm::EXIT_SUCCESS,
-                        true,
-                        false,
-                        String::new(),
-                    );
+                        None,
+                    ) {
+                        error!(
+                            "Failed to execute mount command with error {}.",
+                            e
+                        );
+                        return emsg(
+                            "Failed to mount secure directory tmpfs.",
+                            None::<String>,
+                        );
+                    }
 
                     Ok(s.to_string())
                 }
--- a/src/secure_mount.rs
+++ b/src/secure_mount.rs
@@ -1,10 +1,10 @@
 use super::*;
 
 use common::emsg;
+use std::error::Error;
 use std::fs;
 use std::os::unix::fs::PermissionsExt;
 use std::process::Command;
-
 /*
  * Input: secure mount directory
  * Return: Result wrap boolean with error message
@@ -122,24 +122,15 @@ fn mount() -> Result<String, Box<String>> {
                         })?;
 
                     // mount tmpfs with secure directory
-                    if let Err(e) = tpm::run(
+                    tpm::run(
                         format!(
                             "mount -t tmpfs -o size={},mode=0700 tmpfs {}",
                             common::SECURE_SIZE,
                             s,
                         ),
-                        tpm::EXIT_SUCCESS,
                         None,
-                    ) {
-                        error!(
-                            "Failed to execute mount command with error {}.",
-                            e
-                        );
-                        return emsg(
-                            "Failed to mount secure directory tmpfs.",
-                            None::<String>,
-                        );
-                    }
+                    )
+                    .map_err(|e| e.description().to_string())?;
 
                     Ok(s.to_string())
                 }
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -41,20 +41,26 @@ Following are function from tpm_initialize.py program
  * Getting the tpm data struct and convert it to a json value object to
  * retrive a particular value by the given key inside the tpm data.
  */
-fn get_tpm_metadata_content(key: &str) -> Result<String, Box<String>> {
-    let tpm_data = match read_tpm_data() {
-        Ok(data) => data,
-        Err(e) => return emsg("Failed to read tpmdata.json.", Some(e)),
-    };
-
+fn get_tpm_metadata_content(key: &str) -> Result<String, KeylimeTpmError> {
+    let tpm_data = read_tpm_data()?;
     let remove: &[_] = &['"', ' ', '/'];
-    match tpm_data.get(key) {
-        Some(content) => match content.as_str() {
-            Some(s) => Ok(s.to_string().trim_matches(remove).to_string()),
-            None => emsg("Can't convert to string", None::<String>),
+    tpm_data.get(key).map_or_else(
+        || {
+            Err(KeylimeTpmError::new_tpm_rust_error(
+                format!("Key: {} is missing in tpmdata.json", key).as_str(),
+            ))
         },
-        None => emsg("Key doesn't exist", None::<String>),
-    }
+        |content| {
+            content.as_str().map_or_else(
+                || {
+                    Err(KeylimeTpmError::new_tpm_rust_error(
+                        "Failed to convert Valur to stirng.",
+                    ))
+                },
+                |s| Ok(s.to_string().trim_matches(remove).to_string()),
+            )
+        },
+    )
 }
 
 /*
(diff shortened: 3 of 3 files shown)


src/main.rs
```diff
@@ -367,6 +367,12 @@ fn read_in_file(path: String) -> std::io::Result<String> {
 mod tests {
     use super::*;
 
+    fn init_logger() {
```
frozencemetery: Why's this here?

src/tpm.rs
```diff
@@ -209,12 +210,11 @@ fn get_tpm_manufacturer() -> Option<String> {
         EXIT_SUCCESS,
         true,
         false,
-        String::new(),
-    );
-    let content_result = String::from_utf8(return_output);
-    let content = content_result.unwrap();
+        None,
+    )
+    .unwrap();
```
frozencemetery: nope!

tpmdata.json
```diff
@@ -1 +1,9 @@
...
\ No newline at end of file
+{
```
frozencemetery: Why does this file keep changing?

leonjia0112:
Removed mutex lock in run().
Clear out system panic code, but those prototype functions that don't contain error handling return type need to be updated later.

Thanks!

src/secure_mount.rs
```diff
@@ -113,17 +113,26 @@ fn mount() -> Result<String, i32> {
         }
 
         // mount tmpfs with secure directory
-        tpm::run(
+        match tpm::run(
```
frozencemetery: Pull this tpm::run result into a temp variable, then use an `if let` with early exit for the error.

src/tpm.rs
```diff
@@ -162,17 +163,26 @@ pub fn is_vtpm() -> Option<bool> {
...
+        None,
+    ) {
+        Ok((o, c, f)) => {
+            info!(
+                "Successfully retrieve receive TPM capability information."
+            );
+            (o, c, f)
+        }
+        Err(e) => {
+            // need change the function return type
```
frozencemetery: Seems like it'd make sense to hold off on merging this until we know what the type should be?

src/tpm.rs
```diff
@@ -477,46 +509,45 @@ pub fn run<'a>(
...
-        // println!("{}: {}", key, value);
         env_vars.insert(key.to_string(), value.to_string());
     }
-
     env_vars.insert("TPM_SERVER_PORT".to_string(), "9998".to_string());
     env_vars.insert("TPM_SERVER_NAME".to_string(), "localhost".to_string());
     env_vars
         .get_mut("PATH")
-        .unwrap()
+        .unwrap() // unwrap Option
```
frozencemetery: Please prioritize #19.

src/tpm.rs
leonjia0112: What I mean here is to change the return type of the function to Result, but changing those function isn't part of this commit. That is why I put a comment here.

Are we able to merge this PR before change other functions' return type to Result<>?

src/tpm.rs
leonjia0112: Got it.

src/secure_mount.rs
leonjia0112: Ok.

src/tpm.rs
frozencemetery: Why don't you add another commit to this PR that accomplishes the signature changes you're suggesting?

leonjia0112: Ok, I got it.

src/main.rs
leonjia0112: deleted

src/tpm.rs
```diff
@@ -15,6 +15,7 @@ use std::io::Read;
 use std::process::Command;
 use std::process::Output;
 use std::str;
+use std::sync::Mutex;
```
leonjia0112: This is unused. Remove.

src/tpm.rs
```diff
@@ -491,27 +499,31 @@ Following are function from tpm_exec.py program
 
 /*
  * Input:
- *     cmd: command to be executed
- *     except_code: return code that needs extra handling
- *     raise_on_error: raise exception/panic while encounter error option
- *     lock: lock engage option
+ *     command: command to be executed
+ *     expect_code: code expect to have from execution result
+ *     raise_on_error: early return error while encounter error result if set
```
leonjia0112: update commment

src/tpm.rs
```diff
@@ -491,27 +499,31 @@ Following are function from tpm_exec.py program
...
+ *     lock: lock the output while executing if set
  *     output_path: file output location
  * return:
- *     tuple contains (standard output, return code, and file output)
+ *     tuple contains (return output, return code, and file output)
  *
  * Execute tpm command through shell commands and return the execution
  * result in a tuple. Implement as original python version. Haven't
- * implemented tpm stubbing and metric.
+ * implemented tpm stubbing and metric output. Lock mechanism in here is
```
leonjia0112: no stubbing is needed here.

frozencemetery:
@mbestavros I have thoughts, but maybe first you could take a look at this, and perhaps suggest how the return type of `run` could be more idiomatic?

src/tpm.rs
```diff
@@ -572,40 +584,56 @@ pub fn run<'a>(
...
+        None => {
+            return emsg("Execution status code is None.", None::<String>);
+        }
+    };
 
-    if let (Some(c), true) = (return_code, raise_on_error) {
-        if c != except_code {
-            error!(
+    // Execution return code checking
+    if return_code != expect_code {
```
mbestavros: Given that we compare return codes here, I don't see a need to include `return_code` in the returned value; if the function returns successfully, that should imply the codes match. Whoever called `run` knows the expected return code anyway; they shouldn't need it again.

If the codes don't match, we should call it out in an error specifically mentioning that as the cause and including the code that was returned (in which case it would be useful to the caller).

src/tpm.rs
```diff
@@ -572,40 +584,56 @@ pub fn run<'a>(
...
-            _ => break,
+
+            _ => break 'exec,
         }
     }
 
-    let return_output = output.stdout;
-    let return_code = output.status.code();
+    // preprocess execution result
+    let return_output = String::from_utf8(output.stdout).map_err(|e| {
```
mbestavros: I don't see `return_output` being used by anything apart from incorrect return code error handling, so I don't see a reason to return it when nothing goes wrong. (Unless I'm missing something obvious.)

src/tpm.rs
leonjia0112: You are right. I was thinking about the same thing. It will be better if the return code and the cause associated with the return code are wrapped together in one Err(). Box<String> isn't a ultimately return type for us apparently.

I can make an error type for the run() with another commit if that is ok?

src/tpm.rs
leonjia0112: `return_output` contains information that the caller function needs instead of the error output. That is because after the TPM command executed, the information goes to both the standard output and the file that was given depending on which command you were using.

src/secure_mount.rs
```diff
@@ -122,17 +122,24 @@ fn mount() -> Result<String, Box<String>> {
...
                             "mount -t tmpfs -o size={},mode=0700 tmpfs {}",
                             common::SECURE_SIZE,
                             s,
                         ),
                         tpm::EXIT_SUCCESS,
-                        true,
-                        false,
-                        String::new(),
-                    );
+                        None,
```
leonjia0112: How about use error mapping `map_err()`.

src/tpm.rs
```diff
@@ -281,17 +282,17 @@ pub fn create_quote(
         key_handle, aik_password, pcrmask, nonce, quote_path,
     );
 
-    let (_return_output, _exit_code, quote_raw) =
-        run(command, EXIT_SUCCESS, true, false, quote_path.to_string());
+    let (_, quote_raw) = match run(command, EXIT_SUCCESS, Some(quote_path)) {
```
leonjia0112: use `map_err()`

src/tpm.rs
```diff
@@ -397,16 +404,17 @@ pub fn create_deep_quote(
     );
 
     // RUN
-    let (_, _, quote_raw) =
-        run(command, EXIT_SUCCESS, true, false, quote_path.to_string());
+    let (_, quote_raw) = match run(command, EXIT_SUCCESS, Some(quote_path)) {
```
leonjia0112: same here

src/secure_mount.rs
leonjia0112: needs to change quote() function return type for using map_err().

src/tpm.rs
leonjia0112: needs to change quote() function return type for using map_err().

leonjia0112:
Add support for tpm execution custom error type os the error can wrap around both the return error code and the error message together and return to the caller.

If the return code is not from the TPM execution, 10 is used indicating that the error is coming from inside the run() function. But this I wasn't very sure if it is a good idea for the error code. Comment are welcome.

CI passed. Ready for a review. @frozencemetery @lukehinds. Thanks!!!

src/tpm.rs
```diff
@@ -491,26 +498,30 @@ Following are function from tpm_exec.py program
...
  */
 pub fn run<'a>(
     command: String,
-    except_code: i32,
-    raise_on_error: bool,
-    _lock: bool,
-    output_path: String,
-) -> (Vec<u8>, Option<i32>, Option<String>) {
-    /* stubbing  placeholder */
+    expect_code: i32,
```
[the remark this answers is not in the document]
leonjia0112: enn, I think you are right. I will remove this.
Thanks!

leonjia0112:
Ready for a review.
@frozencemetery @mbestavros @lukehinds 
Thanks.

src/tpm.rs
```diff
@@ -617,6 +533,98 @@ fn read_file_output_path(output_path: String) -> std::io::Result<String> {
     Ok(contents)
 }
 
+/*
+ * Custom Error type for tpm execution error. It contains both error from the
+ * TPM command execution result and error within the run() function.
+ *
+ * Error Code:
+ *  - 10: error cause from the function call within the run()
```
mbestavros: Unless I'm misunderstanding, this header is referring to the original custom error type (that didn't distinguish Rust and TPM errors) which has since been replaced. This should be removed and replaced with something describing the error type as it exists now.

leonjia0112: Thanks for catching this comment. I will update this.

leonjia0112: True. Early return save the world. hahaha

leonjia0112:
> Custom error type looks good, apart from a few small nitpicks I've included inline.
> 
> (It's pretty awesome to see all the space savings this change enables in the diff!)

leonjia0112:
I don't why I closed this PR. Maybe accidentally press the wrong button. Does anyone know how to reopen this PR?

lukehinds:
Might need @frozencemetery to do this, as he as admin (the button is greyed out for me).

frozencemetery:
@leonjia0112 I think only the author (you) can do it (I can't).

leonjia0112:
My button is deactivated. T^T
I don't know why.
The message of the button said: the update-run is force pushed or recreated

PR #31, opened 2018-12-11
Fix log

leonjia0112:
Add a function in the testing suite to initialize the logger. So the command
```
RUST_LOG=keylime_node=trace cargo test
```
would work.


leonjia0112:
Now this commit only fix the no logging bug in testing suite.

Thanks.

PR #33, opened 2018-12-13
Add support to create tpm ek

leonjia0112:
This function is for creating **ek** key in TPM. This is part of the functions in **init()**.

It is run everytime the node is started.
The resulting error is very sufficient for explaining the running result of creating **ek** key.

What this function does is that it simply run the `createek` command and the rest of the function is for error handling if the execution didn't succeed.

Thanks!

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -34,6 +34,63 @@ Following are function from tpm_initialize.py program
 *****************************************************************/
 
 /*
+ * Return: Result wrap the execution result
+ *
+ * This function create a new ek if this hasn't been done before. Otherwise,
+ * it will give the error saying that ek has already been created. Same
+ * implementation as the original python-keylime version. This function only
+ * run once during the starting of cloud node.
+ */
+fn create_ek() -> Result<(), Box<String>> {
+    let (output, code, _) = run(
+        String::from("createek"),
+        EXIT_SUCCESS,
+        false,
+        false,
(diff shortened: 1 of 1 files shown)


src/tpm.rs
```diff
@@ -33,6 +33,45 @@ ftpm_initialize.py
...
+fn create_ek() -> Result<(), Box<String>> {
+    let (output, code, _) = run(
+        String::from("createek"),
+        EXIT_SUCCESS,
+        false,
+        false,
+        String::new(),
+    );
+
+    if code.unwrap() != EXIT_SUCCESS {
```
frozencemetery: Please prioritize #19.

leonjia0112:
Removed unwrap().

Thanks!

src/tpm.rs
leonjia0112: removed

src/tpm.rs
```diff
@@ -33,6 +33,54 @@ ftpm_initialize.py
...
+    let (output, code, _) = run(
+        String::from("createek"),
+        EXIT_SUCCESS,
+        false,
+        false,
+        String::new(),
+    );
+
+    if let Some(EXIT_SUCCESS) = code {
+        Ok(())
```
frozencemetery: change this to a `return`, then de-indent the `else`.  Repeat below and the code should be much more flat.

PR #35, opened 2018-12-14
Update get_tpm_manufacturer function return type

leonjia0112:
1. Changed return type to Result<> instead of Option<>.
2. Removed .unwrap() calls in this function.
3. Return the manufacturer name once the token is found instead of returning at the end of the function.

Thanks!

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -158,33 +158,34 @@ pub fn is_vtpm() -> Option<bool> {
 }
 
 /*
- * getting the tpm manufacturer information
- * is_vtpm helper method
+ * Return: Result wrap manufacture info or error
+ *
+ * Getting the tpm manufacturer information fromn TPM.
+ * Helper function for is_vtpm().
  */
-fn get_tpm_manufacturer() -> Option<String> {
-    let (return_output, _return_code, _file_output) = run(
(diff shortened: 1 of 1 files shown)


src/tpm.rs
```diff
@@ -158,33 +158,34 @@ pub fn is_vtpm() -> Option<bool> {
...
         true,
         false,
         String::new(),
     );
-    let content_result = String::from_utf8(return_output);
-    let content = content_result.unwrap();
+
+    let content = match String::from_utf8(return_output) {
+        Ok(c) => c,
+        Err(e) => return emsg("Faield to convert output to String.", Some(e)),
```
lukehinds: Small typo /s/Faield/Failed

```suggestion
         Err(e) => return emsg("Failed to convert output to String.", Some(e)),
```

lukehinds: No worries, that's about the best I can offer for now on rust reviews :)

leonjia0112:
Changed to fix the typo error.

Thanks!

lukehinds:
If you can fix up the conflicts @leonjia0112 , I think this will be ready to merge as @frozencemetery gives it a +1 here: https://github.com/keylime/rust-keylime/pull/35#pullrequestreview-187501697

PR #36, opened 2018-12-17
Removed all unwraps/panics for all code except testing

leonjia0112:
1. Removed all the unwraps for all the scripts under src/ directory except the testing suite.
2. Changes functions' return type to `Result<>` after removing the unwraps, which has better error handling.
3. Add a script to CI for catch a list of banned function that could possibly be panicking the thread and terminate the process. Both the shell script and the banned list are both `.travis/` directory.

The main change in main.rs is a loop scope is used to wrap around the request, once an error is encountered early exit the loop scope with the response that has the corresponding content regarding the error.

Now there should be now unwraps/panics in the repo and prevent further use of these panic function in this repo.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -2,6 +2,7 @@ extern crate futures;
 extern crate hyper;
 extern crate serde_json;
 
+use hyper::header::HeaderValue;
 use hyper::{header, Body, Response, StatusCode};
 use serde_json::{Map, Value};
 use std::collections::HashMap;
@@ -47,11 +48,12 @@ pub static MOUNT_SECURE: bool = true;
  * 2. There is no space in the json content, but python version response
  * content contains white space in between keys.
  */
-pub fn json_response_content(
+pub fn set_response_content(
     code: i32,
-    status: String,
+    status: &str,
     results: Map<String, Value>,
-) -> Response<Body> {
+    response: &mut Response<Body>,
+) -> Result<(), Box<i32>> {
     let integerated_result = json!({
         "status": status,
         "code": code,
@@ -59,22 +61,22 @@ pub fn json_response_content(
     });
 
     match serde_json::to_string(&integerated_result) {
-        Ok(json) => {
-            // return a json response
-            Response::builder()
-                .header(header::CONTENT_TYPE, "application/json")
-                .body(Body::from(json))
-                .unwrap()
+        Ok(s) => {
+            // Dereferencing apply here because it needs to derefer the variable
+            // so it can assign the new value to it. But changing the headers
+            // doesn't require dereference is because that it uses the returned
+            // header reference and update it instead of changing it, so no
+            // dereference is needed in this case.
+            *response.body_mut() = s.into();
+            response.headers_mut().insert(
+                header::CONTENT_TYPE,
+                HeaderValue::from_static("application/json"),
+            );
+            Ok(())
         }
-
-        // This is unnecessary here，probably won't fail
         Err(e) => {
-            error!("serializing json: {}", e);
-
-            Response::builder()
-                .status(StatusCode::INTERNAL_SERVER_ERROR)
-                .body(Body::from("Internal Server Error"))
-                .unwrap()
+            error!("Failed to convert Json to string for Response body, error {}.", e);
+            Err(Box::new(-1))
         }
     }
 }
@@ -174,6 +176,14 @@ mod tests {
         assert_eq!(
             get_restful_parameters("127.0.0.1:1337/v2/verify/pubkey"),
             map
+        );
+    }
+
+    #[test]
+    fn test_set_response_content() {
+        let mut my_res: Response<Body> = Response::new("nothing".into());
+        assert!(
+            set_response_content(0, "Ok", Map::new(), &mut my_res).is_ok()
         );
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -24,6 +24,8 @@ mod crypto;
 mod secure_mount;
 mod tpm;
 
+use common::emsg;
+use common::set_response_content;
 use futures::future;
 use hyper::rt::Future;
 use hyper::service::service_fn;
@@ -44,7 +46,7 @@ fn main() {
     info!("Starting server...");
 
     /* Should be port 3000 eventually */
-    let addr = "127.0.0.1:1337".parse().unwrap();
+    let addr = ([127, 0, 0, 1], 1337).into();
 
     let server = Server::bind(&addr)
         .serve(|| service_fn(response_function))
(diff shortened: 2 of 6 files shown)


src/secure_mount.rs
```diff
@@ -30,100 +34,126 @@ fn check_mount(secure_dir: &str) -> bool {
...
+            } else {
+                return emsg("Failed to get the path string.", None::<String>);
+            }
         }
     }
 
+    // Monut the directory to file system
     let secure_dir = format!("{}{}", common::WORK_DIR, "/secure");
+    match check_mount(&secure_dir) {
+        Ok(b) => {
```
frozencemetery: Match on `Ok(false)` and `Ok(true)` rather than having an `if` inside the `match`.

src/secure_mount.rs
```diff
@@ -30,100 +34,126 @@ fn check_mount(secure_dir: &str) -> bool {
...
 
-    // if the directory is not mount to file system, mount the directory to
-    // file system
-    if !check_mount(&secure_dir) {
-        let secure_dir_clone = secure_dir.clone();
-        let secure_dir_path = Path::new(secure_dir_clone.as_str());
+                // Create directory if the directory is not exist. The
+                // directory permission is set to 448.
+                if !secure_dir_path.exists() {
+                    match fs::create_dir(secure_dir_path) {
```
frozencemetery: use an `if let` here

src/secure_mount.rs
leonjia0112: Changed.

src/secure_mount.rs
leonjia0112: Changed.

leonjia0112:
Requested change implemented.

Thanks!

leonjia0112:
This is PR that remove all the unwrap in the code base, except the testing suite.

In main, I used loop scope to wrap around the whole function, the function will exit the loop if either finish executing all the code or encounter an error.

In common, the function json_response_content, which was generating a new response this function is called, is changed to set_response_content that a response is borrowed and edit its content based on the given parameters.

In TPM, I simply use a match or if let to remove the unwrap, but many functions' return types are not Result which doesn't support error returning. Therefore, those functions need to be updated with Result return type in the future PR. An issue has been opened regarding this change.

Thanks!

leonjia0112:
Add a bash script for catching panic call in CI.

**Test Passed. Ready for a review.**
Thanks!

src/tpm.rs
```diff
@@ -151,8 +151,11 @@ pub fn is_vtpm() -> Option<bool> {
     match common::STUB_VTPM {
         true => Some(true),
         false => {
-            let tpm_manufacturer = get_tpm_manufacturer();
-            Some(tpm_manufacturer.unwrap() == "ETHZ")
+            if let Some(data) = get_tpm_manufacturer() {
```
frozencemetery: For cases where you can't eliminate the `else` clause (as here, because the `if` needs to yield a value), `match` is more clear; this should be:

```rust
match get_tpm_manufacturer() {
    Some(data) => Some(data == "ETHZ"),
    _ => Some(false).
};
```

src/tpm.rs
```diff
@@ -169,8 +172,10 @@ fn get_tpm_manufacturer() -> Option<String> {
         false,
         String::new(),
     );
-    let content_result = String::from_utf8(return_output);
-    let content = content_result.unwrap();
+    let content = match String::from_utf8(return_output) {
```
frozencemetery: (e.g., like this; this is correct and clear)

src/main.rs
```diff
@@ -63,140 +64,146 @@ fn response_function(req: Request<Body>) -> BoxFut {
...
-                            challenge.unwrap().to_string(),
-                        );
-                        match result {
-                            Ok(hmac) => {
-                                info!(
-                                    "
+    // Loop scope wrap around the request handling
+    // Exit: Encounter error early exit or exit at the end to the scope
+    loop {
+        match req.method() {
```
frozencemetery: I think it might be more clear to have a function for actual request handling that returns a result, then loop-and-log around that.

.travis/panic-test.sh
```diff
@@ -0,0 +1,28 @@
...
+# of the repo. Testing suite is excluded. If such a function presents
+# return exit 1 for fail, otherwise return exit 0 for sucess.
+
+output=""
+panic_line=0
+
+for i in src/*; do
+    tmp_line=$(sed '/Unit Testing/q' $i | grep -e "unwrap()" -e "panic!(.*)" | awk 'END {print NR}')
+    if [ "$tmp_line" -gt 0 ]; then
+        tmp_output=$(sed '/Unit Testing/q' $i | grep -n -e "unwrap()" -e "panic!(.*)")
```
frozencemetery: I'd like to see a list of banned functions pulled out as a variable rather than inlined here.

src/main.rs
leonjia0112: Create two new functions
`get_request_handler()` and `post_request_handler()'
which return Result for error handling.
In such kind of implementation, no loop scope is needed anymore.

.travis/panic-test.sh
leonjia0112: supported

src/tpm.rs
leonjia0112: changed

.travis/panic-test.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Check the unwrap and panic function presenting in src/ directory
+# of the repo. Testing suite is excluded. If such a function presents
+# return exit 1 for fail, otherwise return exit 0 for sucess.
+
+output=""
+panic_line=0
+
+for i in src/*; do
+    tmp_line=$(sed '/Unit Testing/q' $i | grep -e "unwrap()" -e "panic!(.*)" | awk 'END {print NR}')
```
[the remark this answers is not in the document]
leonjia0112: Indeed. This is better.
changed

leonjia0112:
Reimplemented the GET/POST request handler in main.rs into two separate functions that are more adequate for error handling using Result.
**Test Passed. Ready for a review.**
Thanks.

src/secure_mount.rs
```diff
@@ -30,100 +34,119 @@ fn check_mount(secure_dir: &str) -> bool {
...
+    }
 
-        // create directory if the directory is not exist
-        if !secure_dir_path.exists() {
-            match fs::create_dir(secure_dir_path) {
-                Ok(()) => {
-                    let metadata = fs::metadata(secure_dir_path).unwrap();
-                    let mut perm = metadata.permissions();
+    // Monut the directory to file system
+    let secure_dir = format!("{}{}", common::WORK_DIR, "/secure");
```
frozencemetery:
```rust
format!("{}/secure", common::WORK_DIR)
```

src/secure_mount.rs
```diff
@@ -30,100 +34,119 @@ fn check_mount(secure_dir: &str) -> bool {
...
+                let metadata =
+                    fs::metadata(secure_dir_path).map_err(|e| {
+                        Box::new(format!(
+                            "Failed to get file metadata. Error {}",
+                            e
+                        ))
+                    })?;
+
+                // This function only support unix based system
+                // perm.set_mode(448);
```
frozencemetery: This whole comment can go.

src/tpm.rs
```diff
@@ -150,10 +150,10 @@ fn write_tpm_data(data: Value) -> Result<(), Box<String>> {
 pub fn is_vtpm() -> Option<bool> {
     match common::STUB_VTPM {
         true => Some(true),
-        false => {
-            let tpm_manufacturer = get_tpm_manufacturer();
-            Some(tpm_manufacturer.unwrap() == "ETHZ")
-        }
+        false => match get_tpm_manufacturer() {
+            Some(data) => Some(data == "EtHZ"),
+            _ => Some(false),
```
frozencemetery: Why does this function return an `Option<bool>` and not a `bool`?  It never returns `None`...

src/main.rs
```diff
@@ -44,7 +46,7 @@ fn main() {
     info!("Starting server...");
 
     /* Should be port 3000 eventually */
-    let addr = "127.0.0.1:1337".parse().unwrap();
+    let addr = ([127, 0, 0, 1], 3000).into();
```
frozencemetery: Changing port number is not in scope for this PR and would need the comment above removed.

src/tpm.rs
```diff
@@ -616,6 +616,7 @@ fn read_file_output_path(output_path: String) -> std::io::Result<String> {
  * These test are for Centos and tpm4720 elmulator install environment. It
  * test tpm command before execution.
  */
+// Unit Testing
```
frozencemetery: delete

.travis/panic-test.sh
```diff
@@ -0,0 +1,38 @@
...
+            tmp_output=$(sed '/#\[cfg(test)]/q' $i | grep -n -e "$func")
+            output="${output}${tmp_output}\n"
+            panic_line=$(($panic_line+$tmp_line))
+        fi
+    done
+done
+
+# Output result
+if [ "$panic_line" -gt 0 ]
+then
```
frozencemetery: This goes on the previous line

.travis/panic-test.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Getting the banned list from banned functions
+banned_list=()
+banned_function="$1"
+while read -r line; do
+    banned_list=(${banned_list[@]} "$line")
+done < "$banned_function"
+
+output=""
+panic_line=0
```
frozencemetery: 0 isn't a good sentinel.  Consider what'll happen if a banned function appears on the first line of a file.

.travis/panic-test.sh
leonjia0112: Just curious. Would it cause any difference or this is what people normally do?

frozencemetery: You did it yourself earlier in the file.  It's also pretty normal because you get a block structure similar to C (or Rust) - "then" corresponds to "{", and "fi" to "}".

leonjia0112: I see, thanks.

leonjia0112:
**Test Passed. Ready for a review.**
Thanks.

.travis/panic-test.sh
```diff
@@ -0,0 +1,37 @@
...
+catch_panic=$false
+
+# Match panic call based on the given banned funciton list
+for i in src/*; do
+    for func in "${banned_list[@]}"; do
+	    tmp_line=$(sed '/#\[cfg(test)]/q' $i | grep -e "$func" | awk 'END {print NR}')
+        if [ "$tmp_line" -gt 0 ]; then
+            tmp_output=$(sed '/#\[cfg(test)]/q' $i | grep -n -e "$func")
+            output="${output}${tmp_output}\n"
+	    catch_panic=$true
```
frozencemetery: wrong indentation

.travis/panic-test.sh
```diff
@@ -0,0 +1,37 @@
...
+        fi
+    done
+done
+
+# Output result
+if [ $catch_panic ]; then
+    echo "exit 1. Please removed all the following panic calls."
+    printf "$output"
+    exit 1
+else
```
frozencemetery: Drop else and de-indent

src/tpm.rs
```diff
@@ -535,14 +576,16 @@ pub fn run<'a>(
...
-    if return_code.unwrap() != except_code && raise_on_error {
-        panic!(
-            "Command: {} returned {}, expected {}, output {}",
-            command,
-            return_code.unwrap(),
-            except_code.to_string(),
-            String::from_utf8_lossy(&return_output),
-        );
+    if let Some(c) = return_code {
+        if c != except_code && raise_on_error {
```
frozencemetery: combine with previous line; `if let .... && raise_on_error {`

src/tpm.rs
```diff
@@ -142,49 +142,53 @@ fn write_tpm_data(data: Value) -> Result<(), Box<String>> {
...
 }
 
 /*
- * getting the tpm manufacturer information
- * is_vtpm helper method
- */
-fn get_tpm_manufacturer() -> Option<String> {
+ *  * getting the tpm manufacturer information
+ *   * is_vtpm helper method
+ *    */
```
frozencemetery: what's up with the asterisks?

src/tpm.rs
```diff
@@ -142,49 +142,53 @@ fn write_tpm_data(data: Value) -> Result<(), Box<String>> {
...
- * output: boolean
- *
- * If tpm is a tpm elumator, return true, other wise return false
- */
-pub fn is_vtpm() -> Option<bool> {
+ *  * input: None
+ *   * output: boolean
+ *    *
+ *     * If tpm is a tpm elumator, return true, other wise return false
+ *      */
```
frozencemetery: also here

src/tpm.rs
```diff
@@ -142,49 +142,53 @@ fn write_tpm_data(data: Value) -> Result<(), Box<String>> {
...
+pub fn is_vtpm() -> bool {
     match common::STUB_VTPM {
-        true => Some(true),
-        false => {
-            let tpm_manufacturer = get_tpm_manufacturer();
-            Some(tpm_manufacturer.unwrap() == "ETHZ")
-        }
+        true => true,
```
frozencemetery: pull the outer match into early exit

.travis/panic-test.sh
leonjia0112: Changed to using boolean as a sentinel instead of using 0.

src/tpm.rs
leonjia0112: deleted

src/main.rs
leonjia0112: removed changed of port number

src/tpm.rs
leonjia0112: Changed to use boolean instead of Option<>, since bool is sufficient for this function.

src/secure_mount.rs
leonjia0112: Comment removed.

src/common.rs
```diff
@@ -47,34 +48,35 @@ pub static MOUNT_SECURE: bool = true;
  * 2. There is no space in the json content, but python version response
  * content contains white space in between keys.
  */
-pub fn json_response_content(
+pub fn set_response_content(
```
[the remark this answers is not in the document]
leonjia0112: Because I changed the function's implementation from creating a new response each time to modifying the response body. I think the new name is more appropriate for the function.

src/secure_mount.rs
```diff
@@ -30,100 +34,118 @@ fn check_mount(secure_dir: &str) -> bool {
...
-                    error!("Failed to create directory, error {}", e);
-                    return Err(-1);
+                    return emsg("Failed to create directory.", Some(e));
                 }
             }
         }
+        if let Some(s) = secure_dir_path.to_str() {
+            return Ok(s.to_string());
+        } else {
+            return emsg("Failed to get the path string.", None::<String>);
```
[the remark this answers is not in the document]
leonjia0112: changed

src/secure_mount.rs
```diff
@@ -30,100 +34,118 @@ fn check_mount(secure_dir: &str) -> bool {
...
 
         if tokens[2] == secure_dir {
-            if &tokens[0] != &"tmpfs" {
-                error!("secure storage location {} already mounted on wrong file system type: {}.  Unmount to continue.", secure_dir, tokens[0]);
-                debug!(
-                    "secure storage location {} aleady mounted on tmpfs",
-                    secure_dir
-                );
+            if tokens[0] != "tmpfs" {
+                return emsg(format!("secure storage location {} already mounted on wrong file system type: {}.  Unmount to continue.", secure_dir, tokens[0]).as_str(), None::<String>);
```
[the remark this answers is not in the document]
leonjia0112: Is the message string too long? How about using a shorter message instead?

frozencemetery:
As I am quite tired of looking at this PR; I have rewritten the checker in python and squished the results.  I'm curious if you even ran the old code; `$true` and `$false` aren't built-ins in bash.

Some effort will be spent going forward on a cohesive codebase cleanup.

leonjia0112:
My apologies. Sorry for making you have this feeling. Will be more careful next time.

PR #38, opened 2019-01-02
Add support for getting public ek key from TPM

leonjia0112:
1. Add an extra function for creating a temp file and retrieve its file path.
2. This function is used for init() funcrion in TPM initialize script when the
   cloud is node started.

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -34,6 +34,73 @@ Following are function from tpm_initialize.py program
 *****************************************************************/
 
 /*
+ * Return: Result wrap named temp file name or error
+ *
+ * Make a temp file and return its file path.
+ */
+fn create_and_get_temp_file_path() -> Result<String, Box<String>> {
+    match NamedTempFile::new() {
+        Ok(tmp_file) => match tmp_file.path().to_str() {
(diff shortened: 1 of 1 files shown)


src/tpm.rs
```diff
@@ -33,6 +33,73 @@ ftpm_initialize.py
...
+ * Execute TPM command and retrip the public ek key and save to
+ * tpmdata.json file. Same implementation as the original python
+ * version with same error hanlding procedures.
+ */
+fn get_pub_ek() -> Result<(), Box<String>> {
+    let owner_password = get_tpm_metadata_content("owner_pw")?;
+    let tmp_path = create_and_get_temp_file_path()?;
+
+    let (return_output, return_code, ek) = run(
+        format!("getpuek -pwdo {} -ok {}", owner_password, tmp_path),
```
lukehinds: As said in https://github.com/keylime/rust-keylime/pull/39#pullrequestreview-203984790 `tmp_path` - this could be risky. 

Does this have some randomness in the name generation so its not easily guessed, a bit like bash's mktemp

 ```
 mktemp 
/tmp/tmp.8kvXUx3Pxv
```

Issue #42, opened 2019-01-16
Own error type


leonjia0112:
Would it be better if there is an error type for our own crate that can map from other error? (e.g. From<io::Errror>, etc)

Because now the err handling type is a boxed string, which looks not professional. :)

frozencemetery:
(Per out-of-band conversation, this is a good idea.)

lukehinds:
Likely another one that can be closed (based on the merged PR above)

frozencemetery:
@mbestavros how close are you on this?

mbestavros:
I've got a basic custom error for `crypto.rs` working on my end. I'm currently hashing out the best way to wrap `ErrorStack` errors thrown by `rust-openssl`; once that's done, I'll integrate them into the existing bindings and push for review.

mbestavros:
@frozencemetery and @lukehinds: The above two PRs (#49 and #52) implement a custom error type for `crypto.rs` and adjust the return types of the functions introduced in #43, as specified. Unless we'd like to keep this issue open to encapsulate other potential custom errors elsewhere in the codebase, I think this is fair game to be resolved. 

(Personally, I think it'd make more sense to open separate issues for other custom error types, should we want to pursue them.)

frozencemetery:
Agreed, closing.

Issue #20, opened 2018-12-06
Crypto interface


frozencemetery:
All cryptographic operations should be done either on the TPM or through a rust openssl shim.  Interface should hide all "tricky bits" from the rest of keylime.

lukehinds:
I think we can close this now, according to the above merged PR?

frozencemetery:
@mbestavros can you speak to this?

mbestavros:
There are still a few stubs we need that aren't done yet; some are WIP on my end. I'm currently working on building our custom `CryptoError` type for everything in `crypto.rs`  (#42) before doing further work on bindings.

PR #43, opened 2019-01-23
Add OpenSSL bindings for RSA public key import and random RSA key generation

mbestavros:
This is a **work-in-progress** pull request that implements bindings for two crypto functions necessary for the implementation of the Keylime Rust client (as per @leonjia0112). These also constitute a part of the crypto interface as specified in #20. Documentation for both functions is included inline.

Some notes to keep in mind:
- Per offline discussion with @leonjia0112 and @frozencemetery, we will be building our own error type to reconcile the different varieties of error the entire client will be dealing with, as documented in #42. For this reason, error handling should be considered incomplete for these bindings -- with the expectation that the functions should be revised to work with the new error type once it's been implemented.
- Tests do not yet exist for either function -- they should be added in after we implement our own error variety as specified above.

Tagging @frozencemetery and @leonjia0112 for review.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -5,8 +5,11 @@ extern crate openssl;
 use openssl::error::ErrorStack;
 use openssl::hash::MessageDigest;
 use openssl::pkcs5;
-use openssl::pkey::PKey;
+use openssl::pkey::{PKey, Public};
+use openssl::rsa::Rsa;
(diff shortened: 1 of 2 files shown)


mbestavros:
@frozencemetery: Fixed up formatting issues with my additions, and added another commit for a missing semicolon in `secure_mount.rs`.

PR #52, opened 2019-02-26
Update existing crypto bindings to return custom KeylimeCryptoError

mbestavros:
This PR is a follow-up to #49, implementing the new custom error as a return type for all existing crypto bindings. For most functions, the only change is a new return type; however, the `rsa_import_pubkey` function is refactored and overall cleaner as a result.

These changes are part of issue #42.

Tagging @frozencemetery, @leonjia0112, and @lukehinds for review.


leonjia0112:
Looks good to me.

PR #61, opened 2019-04-13
Update README.md

leonjia0112:
Add dependency requirements to the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,19 @@ TPM emulator is a program that runs in the deamon to mimic TPM commands.
 
 ## Prerequisites
 
+### Required Packages
+The rust-keylime agent requires the following packages for both compile and run time.
+
+For Fedora, use the following command
+```
+$ dnf install openssl-devel gcc
+```
+
+For Ubuntu OS, use the following command
+```
+$ apt-get install openssl-dev gcc
+```
+
 ### Rust
 
 Make sure Rust is installed before running Keylime. Installation


README.md
```diff
@@ -25,6 +25,13 @@ TPM emulator is a program that runs in the deamon to mimic TPM commands.
 
 ## Prerequisites
 
+### Required Package
```
lukehinds: Required Package**s**

README.md
```diff
@@ -25,6 +25,13 @@ TPM emulator is a program that runs in the deamon to mimic TPM commands.
...
+Rust-keylime agent rust code implementation require the following packages to both compile time and run time.
```
lukehinds: This rolls off the tongue a little easier..

> The rust-keylime agent requires the following packages for both compile and run time.

frozencemetery: Please note that these are *Fedora* packages.  (If you want to include the Debian names, that would also be fine.)  Formatting this as a `dnf install ...` would also be helpful.

frozencemetery:
The url subject of your commit message should capture what the change is, not just what files have changed.  For a change this small, there shouldn't be a body.

leonjia0112:
Request change resolved. Ready for a review.
@frozencemetery @lukehinds @mbestavros

leonjia0112:
> I would change the commit message as `Update README.md with package dependencies`. This way when we do releases, we can use `git log` and grab the brief details of the change for the release notes.

This one sounds better! Thanks!

PR #63, opened 2019-04-15
Update crypto.rs

leonjia0112:
This is an example implementation of a hardcoded long string,
which rust-fmt won't be able to take the case of. Use format!()
to concentrate long two short string to create a long string,
instead of putting a single long string together.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -194,7 +194,14 @@ mod tests {
         let key = String::from("mysecret");
         let message = String::from("hellothere");
         let mac = do_hmac(key, message);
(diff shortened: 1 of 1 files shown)


leonjia0112:
> Please refer to [how to write a git commit message](https://chris.beams.io/posts/git-commit/). In particular, the subject of this commit isn't useful (I know that it updates crypto.rs from the files changed already...) and more importantly I have no idea _why_ this change is being made.

I feel like the string is too long in here, which can be shortened a little bit. For such kind of thing, rustfmt won't take care of it. But I will update the commit message to make it a little more meaningful.

leonjia0112:
Change addressed and test passed.
Ready for a review. @frozencemetery @mbestavros

PR #70, opened 2019-04-19
Implement stale bot in keylime/rust-keylime

lukehinds:
This (free) bot marks issues / PRs as stale when there is no activity
for 60 days. It will initially make a comment giving 7 days notice.

Should someone comment or apply a change to a PR, the issue will
no longer be considered stale and the automated close tag will be
aborted / revoked.

This really helps keep the PR / issue queue clean of issues
considered unimportant or PR's that have become abandoned.

More details can be found here [0]

[0] https://probot.github.io/apps/stale/

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed after 7 days if no further activity occurs.
+  Thank you for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false


frozencemetery:
Is there perhaps a way to do this only on PRs?  I find it rather frustrating when I file an issue somewhere and it eventually just gets closed as stale without ever getting looked at, and would prefer to avoid that here.

lukehinds:
> Is there perhaps a way to do this only on PRs? I find it rather frustrating when I file an issue somewhere and it eventually just gets closed as stale without ever getting looked at, and would prefer to avoid that here.

It only works for issues unfortunately. So we could tweak the `exemptLabels` so that labels such as `bug` or `enhancement` never expire. A lot of folks rightly have the same concern of stale bot, but it actually plays out nice in real world use. Every issue reaching xx days (we can set that) results in the author, reviewer, participants getting a notification. All they then need to is make a comment within 7 days (also configurable) and it comes out of the expire period. This serves two good purposes, on most projects a lot of installation type issues get opened and then the author vanishes - this way they get notification, so if its important to them , they can say so and keep it alive or at least update with how they fixed the issue. The other one is to remind people to revisit and old issue, quite often folks will raise an issue as a means to lodge an idea on something nice to do in the project - the notfication serves them to revisit where they might decide 'this is no longer relevant as we did X' or 'now that X work is complete we can now do Y'.

It's up to you guys though, I am aware it might be overkill right now as we have a small audience and the rust client is still a WIP, so there is no burning need, but when we get big it can save a project hitting the point where you see 500+ issue lists with a lot of old stale uncared for issues in there. 

Honestly though, happy to park this for now and if it becomes a problem we can replicate this in - in the mean time you can observe how it plays out in /keylime.

frozencemetery:
That makes sense.  I think let's park this for now and see how it works for the rest of keylime.

frozencemetery:
The bot won't let me unassign myself, it seems.  I unfortunately have not had, nor do I anticipate having, cycles to work on keylime.  So don't expect anything from me here :)

mbestavros:
@lukehinds If @frozencemetery is manually removed as a reviewer, he should be unassigned. I don't have permission, however :/

lukehinds:
done!

PR #75, opened 2019-07-03
Add tpm2 version 2 support

leonjia0112:
Implemented a new version of TPM2 in Rust.
This has a TPM2 object that has fields for each instance:
- tpmdata
- algorithms
- legacy_tools

By using this object, we can certainly maintain an TPM object instance in the agent server instead of run tpm functions maintain the file data file. This implementation has merged with the main function. Looking for comment/feedback/suggestion. Thanks.

@jetwhiz @lukehinds @mbestavros @frozencemetery


src/tpm2.rs
```diff
@@ -0,0 +1,955 @@
...
+        let tf_path = self.temp_file_get_path(&tf)?;
+
+        let cmd = match self.legacy_tools {
+            true => format!(
+                "tpm2_readpublic -H {} -o {} -f pem",
+                hex::encode(self.get_tpmdata("ek_handle")),
+                tf_path
+            ),
+            false => format!(
+                "tpm2-readpublic -c {} -o {} -f pem",
```
lukehinds: needs an underscore

~~tpm2-readpublic~~

tpm2_readpublic

leonjia0112: I think I am going to design a test suite in an emulator environment.

leonjia0112:
> Looks ok to me @leonjia0112 , I will see if Robbie / Nabil or Charlie have any reviews before merging.
> 
> How are you testing this code (outside of unit tests), did you check these against an emulator?

I think I am going to design a test suite in an emulator environment.

leonjia0112:
CI passed.

src/crypto.rs
```diff
@@ -15,6 +15,14 @@ use std::io::Error as StdIOError;
 use std::io::Read;
 use std::string::String;
 
+pub fn generate_random_byte(
```
frozencemetery: bytes, plural.

src/main.rs
```diff
@@ -23,6 +23,7 @@ mod common;
 mod crypto;
 mod secure_mount;
 mod tpm;
```
frozencemetery: You should remove the tpm mod as part of this.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+use std::time::SystemTime;
+use tempfile::NamedTempFile;
+use tpm::KeylimeTpmError;
+
+const MAX_TRY: usize = 10;
+const RETRY_SLEEP: Duration = Duration::from_millis(50);
+const TPM_IO_ERROR: i32 = 5;
+const RETRY: usize = 4;
+
+static EMPTYMASK: &'static str = "1";
```
frozencemetery: This needs a comment.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+            None => String::new(),
+            Some(s) => s.to_string(),
+        }
+    }
+
+    pub fn init(
+        &mut self,
+        self_activate: bool,
+        config_pw: String,
+    ) -> Result<(String, String, String, String, String), KeylimeTpmError>
```
frozencemetery: Document what these five strings are.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+        let aik_name = self.get_tpmdata("aik_name");
+        Ok((ek, ekcert.into(), aik, ek_tpm, aik_name))
+    }
+
+    fn emulator_warning(&self) {
+        if self.is_software_tpm() {
+            warn!("{}{}{}{}",
+               "INSECURE: Keylime is using a software TPM emulator rather than",
+               "a real hardware TPM.\nINSECURE: The security of Keylime is NOT",
+               "linked to a hardware root of trust.\nINSECURE: Only use Keylim",
```
frozencemetery: Don't break words when wrapping text.  Check that spacing is preserved on other lines.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+                    KeylimeTpmError::new_tpm_rust_error(
+                        "Error: get public aik key failed.",
+                    )
+                })?,
+            )
+        })?;
+        Ok(())
+    }
+
+    pub fn get_tpm_version(&self) -> i32 {
```
frozencemetery: What is this and why is it a function instead of a constant?

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+        }
+    }
+
+    fn random_password(
+        &self,
+        length: usize,
+    ) -> Result<String, KeylimeTpmError> {
+        let rand_byte = crypto::generate_random_byte(&length)?;
+        let alphabet: Vec<char> =
+            "abcdefghijklmnopqrstuvwxyz0123456789ABCDEFGIJKLMNOPQRSTUVWXYZ"
```
frozencemetery: Add 3 more characters to this?  (Or document why not.)

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+           );
+        }
+    }
+
+    fn set_password(
+        &mut self,
+        config_pw: String,
+    ) -> Result<(), KeylimeTpmError> {
+        let owner_pw = match config_pw.as_str() {
+            "generate" => self.random_password(20)?,
```
frozencemetery: Is this length fixed by something?  If so, please document it.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+                        "failed to read nvram and retrieve ekcert.",
+                    )
+                })?;
+                Ok(base64::encode(ekcert))
+            }
+        }
+    }
+
+    fn read_key_nvram(&mut self) -> Result<String, KeylimeTpmError> {
+        // all error need to be handled by run function
```
frozencemetery: I don't understand this comment.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+    ) -> Result<String, KeylimeTpmError> {
+        let quote_tf = NamedTempFile::new()?;
+        let sign_tf = NamedTempFile::new()?;
+        let pcr_tf = NamedTempFile::new()?;
+
+        if pcrmask.is_empty() {
+            pcrmask = EMPTYMASK.to_string();
+        }
+
+        if !(data.is_empty()) {
```
frozencemetery: No parens.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+
+    pub fn create_deep_quote(
+        &self,
+        nonce: String,
+        data: String,
+        pcrmask: String,
+        vpcrmask: String,
+    ) -> Result<String, KeylimeTpmError> {
+        Err(KeylimeTpmError::new_tpm_rust_error(
+            "Deep quote in progress.",
```
frozencemetery: This... seems wrong?  Why even have this function?  Or are you saying it's not done, at which point: please finish it.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+                                "application. Keylime is incompatible with ",
+                                "with other TPM TSS application like trousers/",
+                                "tpm-tools. Please uninstall or disable.",
+                            )
+                            .as_str(),
+                        ));
+                    }
+
+                    info!(
+                        "Failed to call TPM {}/{} times, try again in {}s.",
```
frozencemetery: "trying again in", or "retrying in".

src/tpm2.rs
lukehinds: i don't mind removing this as its going to be re-implemented anyway (Its part of the old vTPM code that was prototyped on Xen). We plan to port this to QEMU, but the design is still in flight.

src/tpm2.rs
```diff
@@ -0,0 +1,957 @@
...
+                        number_tries, MAX_TRY, RETRY,
```
lukehinds: I can't work out what `RETRY` is for, its set as a const and mutated anywhere else..it will always be `4` from what I can see?

lukehinds: Is this meant to be `retry_interval`?

https://github.com/keylime/keylime/blob/adef1c831f858429560b292f9eebd91c37cc776f/keylime.conf#L101-L103

lukehinds: Further reading, I guess this is meant to be `RETRY_SLEEP`

Is it possible we could use the same naming conventions set in the keylime.conf file.

I have a PR ready to implement INI config loading, and so it makes it easier for me to map values.

src/tpm2.rs
leonjia0112: This is based on the implementation of the python version of Keylime. Looks like a constant is a good practice here. What do you think? @jetwhiz .

src/main.rs
leonjia0112: Ultimately, this should be removed. But if this is removed, for now, the tpm functions in the main function will not work.

src/tpm2.rs
leonjia0112: What other characters are you suggesting to add to the string?

src/tpm2.rs
leonjia0112: Don't worry about this one, it was my note. Removed.

src/tpm2.rs
leonjia0112: That is a good catch!! Sorry for the confusion.

src/tpm2.rs
leonjia0112: I think @jetwhiz has the reason because the length wasn't specified in the man page of the command. 20 was a default value for the function.

src/main.rs
frozencemetery: You have modules `tpm` and `tpm2`.  I don't understand why we wouldn't only have one or the other.

src/tpm2.rs
frozencemetery: Right now you have upper and lowercase alphabet, plus numbers - that's 62, or would be if you hadn't left one out.  (Uppercase 'H'.)  Ideally, it would be 64.  base64 typically uses + and /.

This also allows an alternate, cleaner approach: generate random bytes of the appropriate length, then base64 encode them.

src/tpm2.rs
frozencemetery: Why would it matter what the python version has as constants versus functions?  This is a separate codebase.  If we're not using it as a function, it shouldn't be a function.

src/tpm2.rs
frozencemetery: You did not check that spacing is preserved when wrapping lines.

frozencemetery:
@leonjia0112 I think you may have neglected to upload a new revision of this PR?  The things you have marked as removed/resolved are not removed/resolved in the latest version on here.

lukehinds:
I think with tpm 2.0 being the only version we support, you could rename tpm2 to tpm. 

will review other code shortly.

lukehinds:
We should wait for this https://github.com/keylime/keylime/pull/188 to finish its course first, so we know the exact syntax to use for `tpm2_*` commands:

src/main.rs
lukehinds: @leonjia0112 what are the dependencies from tpm.rs -> main? 

I am not sure why you did not resolve them as part o this patch.

lukehinds:
closing as we will now integrate directly with the ESAPI

PR #76, opened 2019-07-05
Build out infrastructure for POST request handling

mbestavros:
This draft PR contains work done toward handling POST requests for the Rust Keylime agent. It is not ready to be merged into the mainline project as of yet.

The bulk of the work focuses on refactoring the components that handle incoming requests; before, the body of the request was unused, and the request type was used to dictate what happened after. The body was then disposed of and unusable by the handling function. This has been changed, and the POST request's body is now passed onto the POST function to handle it appropriately. That handling logic still needs to be built out, and is a logical next step towards taking this to completion.


lukehinds:
Going to close this, there is so little here / incomplete, that it would be easier and more pragmatic for the engineer who develops this, to do so from afresh.

PR #80, opened 2019-10-13
Update README.md to remove TPM1 ref  and update URL

lukehinds:
Updated old URLs pointing to the mit-keylime repository and
remove references on TPM1 emulator / stack which are no longer
planned to be supported (only TPM 2.0)

Recommend that when we are in the position (working rust agent) we
can update with the details needed to fully install and set up

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,20 +5,19 @@
 ## Overview
 
 This is a Rust implementation of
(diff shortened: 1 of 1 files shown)


lukehinds:
I will sink this one as commit message is improved.

Issue #83, opened 2019-10-29
Secure Mount fails to find existing mount to do incorrect token from string split


lukehinds:
The current mount check code is also not working:

https://github.com/keylime/rust-keylime/blob/ff1543e389eca833cb881add777101d994bdc3c1/src/secure_mount.rs#L30-L44

`tokens[2]` is also incorrect, it should match `/tmp/secure`, but the current match is:

```
"secure type tmpfs (rw,relatime,size=1024k,mode=700)"
```

This is mismatched as we use a back slash to split strings:

```
let tokens: Vec<&str> = line.split('\').collect();
```
This can be fixed by using a whitespace instead...

```
let tokens: Vec<&str> = line.split(' ').collect();
```

We should also emit when an existing mount is found, currently we do not report this.

PR #84, opened 2019-10-30
Fix line split in secure mount and notify

lukehinds:
Currently the code splits the mount command line returns
incorrectly. This results in a non match which means a new
mount is created everytime, instead of using an existing mount.

This results in many mounts being present on the system, every
time the agent is restarted.

This is due to the split element being a back slash `\` whereas
it should be a whitespace `' '`

There is also no feedback provided to the user when an existing
mount is used. This information should be provided to the user.

Resolves: #83

### DIFF
--- a/src/secure_mount.rs
+++ b/src/secure_mount.rs
@@ -28,7 +28,7 @@ fn check_mount(secure_dir: &str) -> Result<bool, Box<String>> {
 
     // Check mount list for secure directory
     for line in lines {
-        let tokens: Vec<&str> = line.split('/').collect();
+        let tokens: Vec<&str> = line.split(' ').collect();
 
         if tokens.len() < 3 {
             continue;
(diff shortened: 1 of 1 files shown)


lukehinds:
Apologies, there is an additional change I omitted to mention which is making the function public. I can reverse this change or put into a new patch, but this is required so that we can actually call the function, so I am happy to have it in this patch if this does not bother others.

frozencemetery:
How does this interact with #82, given that's moving the code around?  Seems like we should wait until that's merged to do this, right?

lukehinds:
> How does this interact with #82, given that's moving the code around? Seems like we should wait until that's merged to do this, right?

Sorry, yes I should have marked this as depending on the other patch landing first.

lukehinds:
@frozencemetery we are good to merge this one now.

Next I will work on landing the tpm2 patch and then we start to call the various functions and connect the code together. I think first we will get the TPM operations working, and then look at the request handlers and post interface in the agent.

Issue #66, opened 2019-04-18
Create tpm struct that can store the tpm info


leonjia0112:
Build a rust version of tpm_interface struct, like the tpm_abstract class in python that can store the information of tpm, for example, hash algorithm, local metadata, and config, which help it manipulate the tpm functions easier. 

I think it is probably a good idea to let the tpm_interface implement the tpm function. Because now the hash algorithm and other manifest are hardcode, by storing those in the tpm_interface struct should make the code now modulized.

PR #86, opened 2020-09-04
hash: Port HashAlgorithms

ashcrow:
Not sure if this is truly needed or not, but just in case ....

`is_accepted` left out as it's more of a generic function.

May help with #66

See: https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_abstract.py#L47


ashcrow:
Updating for linting.

lukehinds (approved):
looks good, thanks steve

Issue #87, opened 2020-09-15
Migrate travis to github actions


lukehinds:
Placeholder issue for CI migration.

@mbestavros another one you might be interested to collaborate / review on.

PR #98, opened 2020-09-24
Migrate to github actions

lukehinds:
Travis is very flaky, the webhooks fail often and its just not as
much of a seamless experience as github actions.

This will implement three checks at gate (same as before):

* cargo fmt
* cargo test
* ./ci/nopanic.py

Resolves: #87


ashcrow (approved):
The change looks correct to me based on what I read in the GH actions doc.

mbestavros (approved):
LGTM!

PR #101, opened 2020-09-29
Update keylime.conf to latest version

lukehinds:
We may need to think of better ways of syncing the conf between
both projects. For now I have moved the file over in its entirety
as we will require integration tests and so forth (so keys around
the verifier and register will be used there).

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -1,80 +1,67 @@
 [general]
 #=============================================================================
-# these options are self explanatory ip/port combinations to find the various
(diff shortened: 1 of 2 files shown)


lukehinds:
some stuff  wrong in CI, will make an issue now.

ashcrow (approved):
I can't speak to all the config options but the move from `general` :arrow_right: `cloud_agent` looks accurate.

mpeters (approved):
lgtm

PR #107, opened 2020-10-01
Implement esapi crate and conf edition 2018

lukehinds:
There are a number of changes in here:

* Implement the ESAPI wrapper crate
* Set up a simple connection context to the TPM
* Remove extern crate calls for edition 2018
* Comment out unused tpm calls in response handlers

Use of this change will require setting up the tpm2-tss. I will
document this in the readme and / or wiki


archive/main.rs
```diff
@@ -0,0 +1,552 @@
+#[macro_use]
```
puiterwijk, line 1: I would suggest not putting in the `archive/` files here.
If we need them, they'll be in the Git history forever.

lukehinds: I agree. I had the same thought and considerated both. I will make another commit and remove archive/

ashcrow:
CI failure is [rustfmt results](https://github.com/keylime/rust-keylime/pull/107/checks?check_run_id=1193570816#step:3:13).

lukehinds:
would be nice to merge this, so we can start to lay down the main tpm code next week

is there anything else anyone would like changed?

ashcrow (approved):

Issue #108, opened 2020-10-01
BUG tpm-tss required in CI


lukehinds:
Now that the esapi-crate is imported, CI requires tpm2-tss 

```
 error: failed to run custom build command for `tss-esapi v4.0.9-alpha.1`

Caused by:
  process didn't exit successfully: `/home/runner/work/rust-keylime/rust-keylime/target/debug/build/tss-esapi-fec97757dc7cff86/build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-env-changed=TSS2_ESYS_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=TSS2_ESYS_STATIC
  cargo:rerun-if-env-changed=TSS2_ESYS_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_SYSROOT_DIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR

  --- stderr
[...]

PR #112, opened 2020-10-01
ci: Install tpm-tss2 dependencies

ashcrow:
Resolves: #108


ashcrow:
The failure CI hit is unrelated to this bug and will be worked out in another PR (possibly #107)

PR #113, opened 2020-10-06
Implements TPM vendor check function

lukehinds:
This change implements a means to check the vendor of the TPM.

The main use for this function is to establish if we are using
a software based TPM and in turn warn that no hardware root of
trust is available

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -1,10 +1,29 @@
+use std::collections::HashMap;
 use std::str::FromStr;
-
 use tss_esapi::constants::algorithm::HashingAlgorithm;
-use tss_esapi::constants::tss as tss_constants;
+use tss_esapi::constants::tss::*;
 use tss_esapi::tss2_esys::{ESYS_TR, ESYS_TR_NONE, ESYS_TR_RH_OWNER};
 use tss_esapi::Context;
 use tss_esapi::Tcti;
+
+fn int_to_str(num: u32) -> String {
+    let mut num = num;
+    let mut result = String::new();
+
+    loop {
+        let chr: u8 = (num & 0xFF) as u8;
+        num >>= 8;
+        if chr == 0 {
+            continue;
+        }
+        let chr = char::from(chr);
+        result.insert(0, chr);
+        if num == 0 {
+            break;
+        }
+    }
+    result
+}
 
 /*
  * Input: None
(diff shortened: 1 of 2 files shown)


lukehinds:
Still learning the idioms of rust, if you feel anything can be improved / optimised let me know of follow up with a commit

image

src/tpm.rs
```diff
@@ -23,3 +42,53 @@ pub(crate) fn get_tpm2_ctx() -> Result<tss_esapi::Context, tss_esapi::Error> {
...
+
+    let mut allprops = HashMap::new();
+    let (capabs, more) = ctx.get_capabilities(
+        TPM2_CAP_TPM_PROPERTIES,
+        TPM2_PT_MANUFACTURER,
+        80,
+    )?;
+
+    if capabs.capability != TPM2_CAP_TPM_PROPERTIES {
+        panic!("Invalid property returned");
```
ashcrow, line 67: nit: May be worth noting what was expected and/or what was received.

src/tpm.rs
```diff
@@ -23,3 +42,53 @@ pub(crate) fn get_tpm2_ctx() -> Result<tss_esapi::Context, tss_esapi::Error> {
...
+    // SAFETY: This is a C union, and Rust wants us to make sure we're using the correct one.
+    // We checked the returned capability type just above, so this should be fine.
+    unsafe {
+        for i in 0..capabs.data.tpmProperties.count {
+            let capab = capabs.data.tpmProperties.tpmProperty[i as usize];
+            allprops.insert(capab.property, capab.value);
+        }
+    }
+
+    let mut vend_strs: String = [
```
ashcrow, line 79: nit: My brain turns to mush when I see many map/filter/etc.. statements on something. Consider adding a comment explaining what's going on here for folks like me :laughing:.

src/tpm.rs
```diff
@@ -1,11 +1,30 @@
...
 use std::str::FromStr;
-
 use tss_esapi::constants::algorithm::HashingAlgorithm;
-use tss_esapi::constants::tss as tss_constants;
+use tss_esapi::constants::tss::*;
 use tss_esapi::tss2_esys::{ESYS_TR, ESYS_TR_NONE, ESYS_TR_RH_OWNER};
 use tss_esapi::Context;
 use tss_esapi::Tcti;
 
+fn int_to_str(num: u32) -> String {
```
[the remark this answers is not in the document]
puiterwijk: Well, it's not really a standard integer-to-string operation. It's basically parsing every byte in a `u32` (i.e. up to 4 bytes) as a different hex character. Because that's how TPM does strings.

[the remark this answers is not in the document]
puiterwijk: (Renaming to `int_to_string` is reasonable though)

src/tpm.rs
puiterwijk: Well, this `TPM2_CAP_TPM_PROPERTIES` expectation is because a few lines up, it's asking for `get_capabilities(TPM2_CAP_TPM_PROPERTIES, ...)`, so this is purely a "Huh, why did the TPM return something incorrect".

src/tpm.rs
puiterwijk: Yeah, sorry. Let me explain what's going on, so Luke can add comments.

This basically first uses a set of constants (`TPM2_PT_VENDOR_STRING_{1,2,3,4}`) as lookup indexes into `allprops`, to get all the values from there where they're set. After that, it removes the values that don't exist (`is_some`), removes the `Option<_>` wrappers (`unwrap()`), removes the items that are actually a value (i.e. `!= && 0`), then puts the individual parts to a different through `int_to_str` to get a `Vec<String>` with the parts, and then makes a single string from them (`.join("")`).

src/tpm.rs
```diff
@@ -23,3 +42,53 @@ pub(crate) fn get_tpm2_ctx() -> Result<tss_esapi::Context, tss_esapi::Error> {
...
+ * Return:
+ *   TPM Vendor Type
+ *   tss_esapi::Error
+ *
+ * get_tpm_vendor will retrieve the type of TPM. This allows us
+ * to perform opertions, such as understand the host is using a
+ * software TPM as opposed to hardware TPM.
+ */
+pub fn get_tpm_vendor() -> Result<String, tss_esapi::Error> {
+    let mut ctx = get_tpm2_ctx()?;
```
puiterwijk, line 57: I think you want to get the context in `main()` and pass that one through to here as `&mut Context`.

src/tpm.rs
lukehinds: I added comments. in time we can move this into a helper function (if they all need the some manipulation)

ashcrow (commented):
LGTM other than the rust format warning in CI

lukehinds:
After speaking with @puiterwijk , I will close this and he will follow up with a patch implementing the same, but using his new error code contained within PR #121

PR #121, opened 2020-10-07
Build Error struct and use it

puiterwijk:
Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>


src/main.rs
```diff
@@ -23,56 +23,57 @@ use tempfile;
...
     // run server forever
-    hyper::rt::run(server);
+    //hyper::rt::run(server);
+    server.await?;
+
+    Ok(())
 }
 
-fn response_function(req: Request<Body>) -> BoxFut {
+async fn response_function(req: Request<Body>) -> Result<Response<Body>> {
```
lukehinds, line 76: Is there any reason for changing to asynchronous , we historically only used async on the verifier (which is one to many) where as the agent has never been a performance factor (as it only deals serving with one verifier typically or talking to the TPM)

src/common.rs
```diff
@@ -34,32 +36,36 @@ pub static MOUNT_SECURE: bool = true;
...
-            error!("Error: unable to read config file: {} ", conf_name);
-            process::exit(1);
-        }
-    };
+pub(crate) fn config_get(
+    conf_name: &str,
+    section: &str,
+    key: &str,
+) -> Result<String> {
+    let conf = Ini::load_from_file(conf_name)?;
```
lukehinds, line 44: That's a lot more tidy!

src/main.rs
puiterwijk: Yes, because that's technically what you always did with Hyper, it was just hidden behind a `Future`. I just rebased to Hyper 0.13 as part of this, so the types worked out.

lukehinds (approved):

PR #126, opened 2020-10-13
Incorrect format from old tool chain

lukehinds:
Apologies folks, my battling with rustfmt led to me breaking the
build. This change now passed cargo fmt and also builds correctly.


lukehinds:
fixed in #129

PR #131, opened 2020-10-13
Implements try_join for running multiple async

lukehinds:
This change enables us to run multiple async operations by using
`futures::try_join`

With this we will be able to run both the web service and the
revocation service, whereas <function>.await will block futher
code execution


ashcrow (approved):
LGTM

puiterwijk (approved):

Issue #92, opened 2020-09-24
Implement Revocation Service


lukehinds:
A revocation service is required for the agent to receive revocation events from the verifier.

PR #134, opened 2020-10-15
Implement revocation service from Python code

ashcrow:
Adds:
- Cargo: Add zmq and bump versions
- crypto: `rsa_verify` function
- main: `run_revocation_service` implementation
- tests/Dockerfile: Updated dep

Note: [Callback](https://github.com/keylime/keylime/blob/561a5b065a206f9873c9892754638fba0c0d7acc/keylime/keylime_agent.py#L596) is not implemented here.

See: https://github.com/keylime/rust-keylime/issues/92

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -3,7 +3,7 @@ use openssl::hash::MessageDigest;
 use openssl::pkcs5;
 use openssl::pkey::{PKey, Private, Public};
 use openssl::rsa::{Padding, Rsa};
-use openssl::sign::Signer;
+use openssl::sign::{Signer, Verifier};
 use std::fs::File;
 use std::io::Read;
 use std::string::String;
@@ -123,12 +123,28 @@ fn to_hex_string(bytes: Vec<u8>) -> String {
     strs.join("")
 }
 
+/*
+ * Input: Local rsa cert, and remote message and signature
+ * Output: true if they are verified, otherwise false
+ *
+ * Verify a remote message and signature against a local rsa cert
+ */
+pub(crate) fn rsa_verify(
+    rsa_key: Rsa<Public>,
+    message: &str,
+    signature: &str,
+) -> Result<bool> {
+    let keypair = PKey::from_rsa(rsa_key)?;
+    let mut verifier = Verifier::new(MessageDigest::sha256(), &keypair)?;
+    verifier.update(message.as_bytes())?;
+    Ok(verifier.verify(&signature.as_bytes())?)
+}
+
 // Unit Testing
 #[cfg(test)]
 mod tests {
     use super::*;
     use openssl::rsa::Rsa;
-    use openssl::sign::Verifier;
 
     // compare with the result from python output
     #[test]
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,6 +42,7 @@ mod error;
 mod hash;
 mod keys_handler;
 mod quotes_handler;
+mod revocation;
 mod secure_mount;
 mod tpm;
 
@@ -95,11 +96,7 @@ async fn main() -> Result<()> {
     .run()
     .map_err(|x| x.into());
     info!("Listening on http://{}:{}", cloudagent_ip, cloudagent_port);
-    try_join!(actix_server, run_revocation_service())?;
-    Ok(())
-}
-
-async fn run_revocation_service() -> Result<()> {
+    try_join!(actix_server, revocation::run_revocation_service())?;
     Ok(())
 }
 
--- a/src/revocation.rs
+++ b/src/revocation.rs
@@ -1,0 +1,130 @@
+#[macro_use]
+use log::*;
+
+use crate::common::config_get;
+use crate::crypto;
+use crate::error;
+use crate::error::{Error, Result};
+use crate::secure_mount;
+
+use std::path::Path;
+
+use serde_json::Value;
+
+/// Handles revocation messages via 0mq
+/// See:
+/// - URL: https://github.com/keylime/keylime/blob/master/keylime/revocation_notifier.py
+///   Function: await_notifications
+pub(crate) async fn run_revocation_service() -> Result<()> {
+    let mount = secure_mount::mount()?;
+    let revocation_cert_path =
+        format!("{}/unzipped/RevocationNotifier-cert.crt", mount);
+
+    // Connect to the service via 0mq
+    let context = zmq::Context::new();
+    let mysock = context.socket(zmq::SUB)?;
+
+    mysock.set_subscribe(b"")?;
+
+    let revocation_ip =
+        config_get("/etc/keylime.conf", "general", "receive_revocation_ip")?;
+    let revocation_port = config_get(
+        "/etc/keylime.conf",
+        "general",
+        "receive_revocation_port",
+    )?;
+    let endpoint = format!("{}:{}", revocation_ip, revocation_port);
+
+    mysock.connect(endpoint.as_str())?;
+
+    // Unlike the python agent we do not attempt lazy loading. We either
+    // have the certificate, or we don't. If we don't have a key or can't load
+    // the key we return a Configuration error as the service will not work.
+    let cert_key = if Path::new(&revocation_cert_path).exists() {
+        info!(
+            "Loading the revocation certificate from {}",
+            revocation_cert_path
+        );
+        match crypto::rsa_import_pubkey(revocation_cert_path) {
+            Ok(v) => v,
+            Err(e) => {
+                return Err(error::Error::Configuration(String::from(
+                    "Can not load pubkey",
+                )))
+            }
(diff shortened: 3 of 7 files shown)


src/main.rs
```diff
@@ -76,7 +77,69 @@ async fn runWebServer() -> Result<()> {
...
+
+    loop {
+        let rawbody = mysock
+            .recv_string(0)
+            .expect("failed to receive 0mq message")
+            .unwrap();
+        let body: Value = serde_json::from_str(rawbody.as_str())?;
+        if cert_key == "" {
+            // load up the CV signing public key
+            if revocation_cert_path != "" && Path::new(&revocation_cert_path).exists() {
```
ashcrow, line 116: `revocation_cert_path` can't be empty -- will drop the former and keep the later check

ashcrow:
Will fix format issues

lkatalin (approved):
I think this makes sense, with the caveat that it should get some more sets of eyes as I'm new to Keylime.

lukehinds:
The code looks good to me. Would it take much to refactor into a `zmq.rs`? I am aware I left that placeholder function for you to use, but I am now it bit mindful that we should try to keep `main.rs` slimmed down when we can. 

The other thing would be some unit tests, but I am understanding this might not be easy with ZMQ and might be more suited to a functional / integration test.

ashcrow:
@lukehinds That's a good idea. I'll move it out into it's on file and have it utilized in `main.rs`

ashcrow:
Updated

lukehinds (approved):

src/revocation.rs
```diff
@@ -0,0 +1,134 @@
...
+    let endpoint = format!("{}:{}", revocation_ip, revocation_port);
+
+    mysock
+        .connect(endpoint.as_str())
+        .expect("could not connect to revocation service");
+
+    // Unlike the python agent we do not attempt lazy loading. We either
+    // have the certificate, or we don't. If we don't have a key or can't load
+    // the key we return a Configuration error as the service will not work.
+    if Path::new(&revocation_cert_path).exists() {
```
puiterwijk, line 52:
```suggestion
    let cert_key = if Path::new(&revocation_cert_path).exists() {
```

src/revocation.rs
```diff
@@ -0,0 +1,134 @@
...
+        info!(
+            "Loading the revocation certificate from {}",
+            revocation_cert_path
+        );
+        cert_key =
```
puiterwijk, line 57:
```suggestion
```

src/revocation.rs
```diff
@@ -0,0 +1,134 @@
...
+            match crypto::rsa_import_pubkey(revocation_cert_path.clone()) {
+                Ok(v) => v,
+                Err(e) => {
+                    return Err(error::Error::Configuration(String::from(
+                        "Can not load pubkey",
+                    )))
+                }
+            };
```
puiterwijk, line 65:
```suggestion
            }
```

src/revocation.rs
puiterwijk: If you take this, together with the following two patches, you don't need the `let mut cert_key`, as you can set it in the single block of code! Rust is awesome, because an `if ... { V }` statement returns the `V` value!

src/revocation.rs
```diff
@@ -0,0 +1,134 @@
...
+/// Handles revocation messages via 0mq
+/// See:
+/// - URL: https://github.com/keylime/keylime/blob/master/keylime/revocation_notifier.py
+///   Function: await_notifications
+pub(crate) async fn run_revocation_service() -> Result<()> {
+    let mount = secure_mount::mount()?;
+    let revocation_cert_path =
+        format!("{}/unzipped/RevocationNotifier-cert.crt", mount);
+
+    let mut cert_key;
```
puiterwijk, line 24:
```suggestion
```

src/revocation.rs
```diff
@@ -0,0 +1,134 @@
...
+        )));
+    }
+
+    info!("Waiting for revocation messages on 0mq {}", endpoint);
+
+    // Main revocation service loop. If a message is malformed or
+    // can not be verified the loop continues.
+    loop {
+        let mut rawbody = match mysock.recv_string(0) {
+            Ok(v) => v.expect("unable to read message from 0mq"),
```
puiterwijk, line 79: So, in this `loop {` you don't want to use the `?` operator, since that will break out of the entire function.
Instead, maybe use something like:
```
let rawbody = match mysock.recv_string(0) {
    Ok(v) => match v {
        Some(v) => v,
        _ => {
            warn!("Unable to read message from 0mq");
            continue;
         },
    Err(e) => ....
```

puiterwijk (changes requested):
Please fix the comments I pointed out. There's a few more `.expect()` uses.

ashcrow:
Thank you @puiterwijk. I'll review and update!

ashcrow:
Updated based on @puiterwijk's review.

src/crypto.rs
```diff
@@ -128,6 +129,23 @@ fn to_hex_string(bytes: Vec<u8>) -> String {
...
+/*
+ * Input: Local rsa cert, and remote message and signature
+ * Output: true if they are verified, otherwise false
+ *
+ * Verify a remote message and signature against a local rsa cert
+ */
+pub fn rsa_verify(
+    rsa_key: Rsa<Public>,
+    message: &str,
+    signature: String,
```
puiterwijk, line 141: Note that this is still a `String`.

src/revocation.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    // Unlike the python agent we do not attempt lazy loading. We either
+    // have the certificate, or we don't. If we don't have a key or can't load
+    // the key we return a Configuration error as the service will not work.
+    let cert_key = if Path::new(&revocation_cert_path).exists() {
+        info!(
+            "Loading the revocation certificate from {}",
+            revocation_cert_path
+        );
+        match crypto::rsa_import_pubkey(revocation_cert_path.clone()) {
```
puiterwijk, line 49: If `rsa_import_pubkey` had accepted an `&str`, you wouldn't have needed to `.clone()` the path here. That's exactly why that is more idiomatic 😄

src/revocation.rs
```diff
@@ -0,0 +1,131 @@
...
+            Some(v) => v,
+            _ => {
+                warn!("No msg on revocation message from server");
+                continue;
+            }
+        };
+
+        // Verify the message and signature with our key
+        let mut verified =
+            crypto::rsa_verify(cert_key.clone(), message, signature);
```
puiterwijk, line 105: Is there any chance `cert_key` could be given as a reference too (`&...`)? That would save us from having to `clone()` every time we get a message.

puiterwijk (commented):
A few more notes remaining, but starting to look decent!

src/revocation.rs
ashcrow: We can make `rsa_verify` take a reference but once inside the function it seems like we need to clone for [`PKey::from_rsa`](https://docs.rs/openssl/0.10.31/openssl/pkey/struct.PKey.html#method.from_rsa) though I may be wrong. Thoughts?

src/revocation.rs
ashcrow: Makes sense, thanks! I've updated the others but I'll leave this one be for now since it's in `rsa_import_pubkey`.

ashcrow:
Rebased

src/revocation.rs
ashcrow: Bump :-)

lukehinds: I am good with this now, @puiterwijk if @ashcrow answer satisfies you, we can merge this

ashcrow:
Will rebase shortly

src/revocation.rs
puiterwijk: Sure. Let's merge it like this and improve later.

ashcrow:
Rebased!

ashcrow:
Bump

lukehinds (approved):

lukehinds:
hey @puiterwijk this needs approval from you as well (changes requested)

image

lukehinds:
@puiterwijk ,please ack if the review request is resolved, so we can merge this. I am sure we will revisit the code again once then parts are rigged together.

puiterwijk (approved):
Sorry for forgetting to hit the Approve button. Thanks for the PR!

ashcrow:
Bah! Needs another rebase!

lukehinds:
hey @ashcrow , keen to approve this for you. It seems all we have now are some redundant clones (see CI 'Fedora tests'). If you could make another quick commit I would lie to merge this and start rigging up the revocation actions.

lkatalin:
Looks like this resolves #92 (hopefully this comment will link it)

PR #135, opened 2020-10-21
actix_web Implementation

lukehinds:
Implement actix web and handlers required for the keylime agent. 

The routes are mostly placeholders for now.


src/main.rs
```diff
@@ -29,55 +28,39 @@ use error::{Error, Result};
...
-
-    let service = make_service_fn(|_| async {
-        Ok::<_, Error>(service_fn(handlers::response_function))
-    });
-    let server = Server::bind(&addr).serve(service);
-
-    info!("Listening on http://{}", addr);
-
-    server.await?;
+    let actix_server = HttpServer::new(move || {
```
ashcrow, line 40: The mounting of routes and the creation of the server directly seems like it would make sense in another file and used here. However I think this is good as is and can be changed later _if_ we decide to do so.

ashcrow (approved):
Looks good -- though I'd recommend squashing the commits down before merge.

ashcrow:
```
  thread 'main' panicked at 'Error with pkg-config finding tss2-esys.: Failure { command: "\"pkg-config\" \"--libs\" \"--cflags\" \"tss2-esys\" \"tss2-esys >= 2.3.3\"", output: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "Requested \'tss2-esys >= 2.3.3\' but version of tss2-esys is 2.3.2\nYou may find new versions of tss2-esys at https://github.com/tpm2-software/tpm2-tss\n" } }', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tss-esapi-4.0.9-alpha.1/build.rs:14:14
```

Same error I was seeing. Looks like a version mismatch.

src/main.rs
lukehinds: I will have a play with that , something like `routes.rs` , it may be tricky though because of actix being heavily future based. I had a good look on github searches , but not many folks were separating the routes out of main. I do agree though, its a cleaner way of doing it.

lukehinds:
> ```
>   thread 'main' panicked at 'Error with pkg-config finding tss2-esys.: Failure { command: "\"pkg-config\" \"--libs\" \"--cflags\" \"tss2-esys\" \"tss2-esys >= 2.3.3\"", output: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "Requested \'tss2-esys >= 2.3.3\' but version of tss2-esys is 2.3.2\nYou may find new versions of tss2-esys at https://github.com/tpm2-software/tpm2-tss\n" } }', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tss-esapi-4.0.9-alpha.1/build.rs:14:14
> ```
> 
> Same error I was seeing. Looks like a version mismatch.

@puiterwijk I am not sure if I am recalling correctly, but you thought you had a fix for this?

PR #139, opened 2020-11-04
Use git tss-esapi and add tpm vendor check

puiterwijk:
This PR makes us use the Git copy of tss-esapi for now, while we are still actively adding features there.
It also adds the TPM vendor check from #113.


lukehinds (approved):
LGTM

PR #141, opened 2020-11-12
Add create_ek() to tpm.rs

lkatalin:
If this needs more functionality, different return types, etc., please let me know.

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -72,7 +72,7 @@ pub(crate) fn create_ek(
             &mut offset,
         );
         if res != 0 {
-            panic!("out of memory or invalid data received from TPM");
+            panic!("out of memory or invalid data received from TPM"); //#[allow_ci]
         }
         tpm_pub_vec.set_len(offset as usize);
     }
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -1,10 +1,11 @@
 use std::str::FromStr;
 
-use crate::Result;
+use crate::{common::config_get, Error as KeylimeError, Result};
 
 use tss_esapi::{
-    constants::{algorithm::HashingAlgorithm, tss::*},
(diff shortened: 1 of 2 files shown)


src/tpm.rs
```diff
@@ -24,3 +25,36 @@ pub(crate) fn get_tpm2_ctx() -> Result<tss_esapi::Context> {
...
+ * let (key, cert) = tpm::create_ek(context, Some(AsymmetricAlgorithm::Rsa))
+ */
+pub(crate) fn create_ek(
+    context: &mut Context,
+    alg: Option<AsymmetricAlgorithm>,
+) -> Result<(KeyHandle, Vec<u8>)> {
+    let alg = match alg {
+        Some(a) => a,
+        None => {
+            // TODO: What other values could be in keylime.conf?
```
puiterwijk, line 42: I think the only other value we should consider correct is `ecc`.
I'd say that the `_ =>` should return an error instead.

puiterwijk (commented):
This looks decent with just one small comment!

lkatalin:
Here is an updated PR with an error for encryption algorithms other than rsa and ecc, and retrieving the TPM pub object and turning it into a `Vec<u8>`.

I don't think that turning the TPM pub object into a `Vec<u8>` should return `KeylimeError::InvalidRequest`; it seems better to return a `TPM(tss_esapi::Error::WrapperError(...WrapperErrorKind::WrongValueFromTPM))`, but I wasn't able to create one because it would not let me import `WrapperErrorKind`. Any thoughts on the best Error type here? If we end up just using an error from Keylime, I will drop the import of the `tss_esapi::Error`.

lkatalin:
Depends on #145 #144

puiterwijk (approved):

PR #143, opened 2020-11-17
Allow nominator for CI checks on  banned calls

lukehinds:
Calls to nopanic / unwrap can be allowed by using #[allow_ci]

### DIFF
--- a/tests/nopanic.ci
+++ b/tests/nopanic.ci
@@ -1,5 +1,9 @@
 #!/usr/bin/env python3
 
+'''
+To prevent CI failing for approved instance of banned words, add a comment: #[allow_ci]
+'''
+
 import os
 
 banned = ["unwrap(", "panic!("]
@@ -9,14 +13,12 @@ print("Files to check: %s" % srcs)
 
 failed = False
 for f in srcs:
-    print("Checking file %s" % f)
-    contents = open("src/%s" % f, "r").read().split("#[cfg(test)]")[0]
-    for b in banned:
-        if b not in contents:
-            continue
-
-        failed = True
-        print("File %s calls banned function: %s)" % (f, b))
-    pass
-
+    with open("src/" + f) as src_file:
+        for line_no, line in enumerate(src_file):
+            for b in banned:
+                if b not in line or "#[allow_ci]" in line:
+                    continue
+                failed = True
+                print("File %s on line number  %s calls banned function: %s)" % (f, line_no, b))
+            pass
 exit(failed)
--- a/tests/nopanic.ci
+++ b/tests/nopanic.ci
@@ -19,6 +19,6 @@ for f in srcs:
                 if b not in line or "#[allow_ci]" in line:
                     continue
                 failed = True
-                print("File %s on line number  %s calls banned function: %s)" % (f, line_no, b))
+                print("File %s on line number  %s calls banned function: %s)" % (f, line_no + 1, b))
             pass
 exit(failed)


tests/nopanic.ci
```diff
@@ -9,14 +13,12 @@ print("Files to check: %s" % srcs)
...
-            continue
-
-        failed = True
-        print("File %s calls banned function: %s)" % (f, b))
-    pass
-
+    with open("src/" + f) as src_file:
+        for line_no, line in enumerate(src_file):
+            for b in banned:
+                if b not in line or "#[allow_ci]" in line:
```
lkatalin, line 19: I think this would need to check for `#[allow_ci]` in the previous line, unless I am misreading this.

lukehinds: Hmm, this seems to work well:

https://github.com/keylime/rust-keylime/runs/1413393155?check_suite_focus=true

The previous line is just to go over the banned dict , I think this could likely be optimized a bit better using regexp, but we don't need to worry about performance too much

lukehinds: it looks like it was not even working before (it was allowing a load of unwraps).

lkatalin: Ah, yeah, that's a lot of unwraps - it's good to have this as an improvement. 

I am seeing the line numbers reported as one off from where the `unwrap()` is called, though, ex. `File crypto.rs on line number  200 calls banned function: unwrap()` while I see the `unwrap()` on 201: https://github.com/keylime/rust-keylime/blob/master/src/crypto.rs#L201. Not a big deal but easy to change.

tests/nopanic.ci
```diff
@@ -9,14 +13,12 @@ print("Files to check: %s" % srcs)
...
+                    continue
+                failed = True
+                print("File %s on line number  %s calls banned function: %s)" % (f, line_no, b))
```
lkatalin, line 22:
```suggestion
                print("File %s on line number  %s calls banned function: %s)" % (f, line_no + 1, b))
```

tests/nopanic.ci
lkatalin: My concern is more about the `#[allow_ci]` being placed on the previous line from `panic!` or `unwrap()` - would that be caught here?

lkatalin:
@lukehinds I created a branch to test this by using `#[allow_ci]` on one of the `unwraps()` [here](https://github.com/keylime/rust-keylime/commit/3fa8d33d19f25a048769c213473827d7883f5905), but the line was still flagged by the `nopanic` test as `File cmd_exec.rs on line number  136 calls banned function: unwrap()`. Is this a different way than how you were thinking of using it? I was thinking of something like [this](https://github.com/keylime/rust-keylime/compare/master...lkatalin:allowpanic).

There are still improvements to be made, ex. it's a bit awkward to put this on the line immediately before an `unwrap()` instead of immediately before a multi-line statement or function using `unwrap()` like a real Rust attribute, but it works for now without a lot of parsing.

tests/nopanic.ci
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env python3
 
+'''
+To prevent CI failing for approved instance of banned words, add a comment: #[allow_ci]
```
lkatalin, line 4: Also, the more I think about it, it may be more readable to call this something like `#[allow_panic]`.

lukehinds: I see, I was thinking of tagging it on the same line , we do the same in bandit:

https://github.com/PyCQA/bandit/blob/9b4cf91e51b2cea6754bbb80f85bf968c92d4201/doc/source/config.rst#suppressing-individual-lines

I am not of any strong view though, it could go underneath - but then there could be an instance of two (or more) sequential lines  having unwraps.

tests/nopanic.ci
lukehinds: I see now, so I planned to have it on the same line as the instance of panic! / unwrap

`key.unwrap() #[allow_ci]`

Here is an example of it working:

image

(Note, this is before you're change to increment the line by +1 so it records the correct line number).

lkatalin: Ah, this makes much more sense now. Thanks for the clarification!

lkatalin (approved):
Thanks @lukehinds

tests/nopanic.ci
lukehinds: no worries. what you would you like to do around the naming convention? I went with allow_ci rather than a mention of panics (as it also covers unwraps as well)?  I am open to suggestions here.

lukehinds:
> Thanks @lukehinds
> 
> Edit: Do we need CI to pass for this to merge? And / or should I make a separate patch to either get rid of the `unwraps()` or mark them as acceptable?

We can merge this and then follow up with a patch. I am fine to mark all those unwraps as allow, I imagine a lot of those will be refactored out when we work more on the crypto side. 

I will go ahead and merge this.

lukehinds:
actually, looks like I need to fix these first (the repo is stopping admin overriding for mergeing)

tests/nopanic.ci
lkatalin: I think `allow_ci` is fine for now; if we end up allowing more types of banned calls, we may want to differentiate later. Alternatively `allow_banned` is possibly a little more descriptive.

PR #149, opened 2020-12-02
Registrar call

lukehinds:
This call takes tpm artifacts (ek, ekcert, ek_tpm, aik etc) and sends the to the registrar using a http client (reqwest).

The registrar then returns a keyblob in the response, which will then be passed to tpm activation. 

Note crypto strings are temporary and will be removed by final commit (once the AIK code is in).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,12 +7,14 @@ pub(crate) enum Error {
     InvalidRequest,
     Ini(ini::ini::Error),
     Configuration(String),
+    Reqwest(reqwest::Error),
     Serde(serde_json::Error),
     Permission,
     IO(std::io::Error),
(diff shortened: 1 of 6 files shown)


src/error.rs
```diff
@@ -24,6 +26,9 @@ impl fmt::Display for Error {
             }
             Error::TPM(err) => write!(f, "TPM Error encountered: {}", err),
             Error::ActixWeb(err) => write!(f, "HttpServer({})", err),
+            Error::UUIDError(err) => {
```
lukehinds, line 29: note to self: copy and paste residue

puiterwijk (approved):
I think this code looks pretty decent, but I am biased, so I think we should get another review of this.

PR #151, opened 2020-12-10
Implement get_ek() function

### LINKED ISSUE #115: Implement use EK Function
The following use EK function is required

https://github.com/keylime/keylime/blob/72c04b05bfb8990d1ac68186657ac27f30902503/keylime/tpm/tpm2.py#L444

This issue is dependent on implementation of #114 first.

lkatalin:
Resolves #115 

I changed the name ~~to `get_ek()` because that seems like what it does (rather than "use" the ek)~~.
Please let me know if I should make changes or if this isn't the right way to generate the key handle.

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -83,3 +83,37 @@ pub(crate) fn create_ek(
 
     Ok((handle, cert, tpm_pub_vec))
 }
+
+/* Converts a hex value in the form of a string (ex. from keylime.conf's
+ * ek_handle) to a key handle.
+ *
+ * Input: &str
+ * Return: Key handle
+ *
+ * Example call:
+ * let ek_handle = tpm::ek_from_hex_str("0x81000000");
+ */
+pub(crate) fn ek_from_hex_str(val: &str) -> Result<KeyHandle> {
+    let val = val.trim_start_matches("0x");
+    Ok(KeyHandle::from(u32::from_str_radix(val, 16)?))
+}
+
+#[test]
+fn ek_from_hex() {
+    assert_eq!(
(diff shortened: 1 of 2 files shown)


src/tpm.rs
```diff
@@ -83,3 +83,27 @@ pub(crate) fn create_ek(
...
+ * Input: None
+ * Return: Key handle
+ *
+ * Example call:
+ * let ek_handle = tpm::get_ek();
+ */
+
+pub(crate) fn get_ek() -> Result<KeyHandle> {
+    let cfg_handle =
+        config_get("/etc/keylime.conf", "cloud_agent", "ek_handle")?;
```
puiterwijk, line 98: Perhaps have the function take an `&str` of the config value, and use that? Basically, I think it would be nice if this just did the parsing into a `KeyHandle`, and let the config handling over to the main functions.

puiterwijk (approved):
Other than one small nitpick, looks good!

src/tpm.rs
lkatalin: @puiterwijk Let me know if this shorter version is more like what you're talking about here. I can revise if not.

puiterwijk (approved):
:shipit:

puiterwijk:
Actually, it might be useful to get a test added. But the code looks good to me.

lkatalin:
@puiterwijk I added some tests. I didn't see a way to get a value back out of a `KeyHandle`, so I didn't test converting it back.

puiterwijk (approved):

puiterwijk:
Looks good to me! Thanks!

PR #157, opened 2021-02-03
Basic quote handlers

lukehinds:
The following code replicates python agent code:

https://github.com/keylime/keylime/blob/d987c71dd878492ac2b59eca781009fe270d219a/keylime/keylime_agent.py#L95-L114

This is more or a less at the point that we can call the tpm for quote.

Signed-off-by: Luke Hinds <lhinds@redhat.com>


ashcrow (approved):
Code looks good for wiring up the handlers with placeholders. If possible it would be nice to squash the commits down.

ashcrow (approved):

PR #158, opened 2021-02-04
Implement changes for registrar call

### LINKED ISSUE #95: Implement Agent registration
A HTTP Client is required to interact with the registrars rest interface, this allows the transfer of keys required to set up a hardware route of trust.

### LINKED ISSUE #119: Create AIK
Implement the create AIK function

https://github.com/keylime/keylime/blob/72c04b05bfb8990d1ac68186657ac27f30902503/keylime/tpm/tpm2.py#L559

This first requires object creation implemented in issues #115 , #115 , #116 , #117 and #118

lkatalin:
This PR adds relevant prerequisites and then uses `create_ek()` to retrieve the EK cert and pubkey and adds functions to retrieve the AK pubkey as well. It sends all of this data to the registrar. Much of it was written or co-written by @lukehinds so I have included him on those commits. Supersedes #149 . 

Fixes #119 
Fixes #95 

~~@lukehinds @puiterwijk I'm especially interested in making sure the `create_ak()` function is correct.~~

- [x] Currently needs to be rebased on #162 to pass CI.

- [x] TODO: I can also add unit testing with [`reqwest_mock`](https://docs.rs/reqwest_mock/0.7.0/reqwest_mock/).


src/main.rs
```diff
@@ -51,11 +52,39 @@ use std::fs::File;
...
 use std::path::Path;
+use tss_esapi::constants::algorithm::AsymmetricAlgorithm;
+use uuid::Uuid;
 
 use error::{Error, Result};
 
 static NOTFOUND: &[u8] = b"Not Found";
 
+fn get_uuid(agent_uuid_config: &str) -> String {
+    match agent_uuid_config {
```
ashcrow, line 63: Will this cover all possible entries or should there be a default match as well?

lkatalin: @ashcrow Thanks for looking at this - line 77 below is supposed to take care of any cases that don't match these pre-defined strings. Were you thinking of something else?

ashcrow: Ah I see. The submatch attempts to parse and either passes and returns or fails and returns a new uuid. That should be fine. My brain was looking for a `_` match :-)

lkatalin: Ah yeah, gotcha. : ) Thanks for double checking.

ashcrow:
This looks good from my POV!

src/registrar_agent.rs
```diff
@@ -0,0 +1,95 @@
...
+where
+    D: serde::Deserializer<'de>,
+{
+    Option::<WrappedBase64Encoded>::deserialize(deserializer)
+        .map(|wrapped| wrapped.map(|wrapped| wrapped.0))
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Register<'a> {
+    ek: &'a [u8],
```
puiterwijk, line 37: This is an `str`, since it's PEM encoded

src/registrar_agent.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[serde(serialize_with = "serialize_as_base64")]
+    ekcert: &'a [u8],
+    #[serde(serialize_with = "serialize_as_base64")]
+    ek_tpm: &'a [u8],
+    aik: &'a [u8],
```
puiterwijk, line 42: This is `&str` as well.

src/registrar_agent.rs
```diff
@@ -0,0 +1,95 @@
...
+    };
+    reqwest::Client::new()
+        .post(&format!(
+            "http://{}:{}/agents/{}",
+            registrar_ip, registrar_port, agent_uuid
+        ))
+        .json(&data)
+        .send()
+        .await?
+        .json()
```
puiterwijk, line 92: This will fail if the server returned an error, because then `code` is non-200 and `status` is not `success`, and it won't have a `results` block.

puiterwijk (changes requested):
Unfortunately, the types are somewhat vague, so we'll need some fixes :(.
I'll send you my fixes so you can use that.

lkatalin:
@puiterwijk Let me know when you have a good idea of the changes you have / want on this one in addition to your comments above.

src/registrar_agent.rs
lkatalin: I moved the status code check to be before this `.json()` is called and just used the built-in `reqwest::StatusCode` method.

src/registrar_agent.rs
```diff
@@ -0,0 +1,144 @@
...
+    let data = Activate { auth_tag };
+
+    let addr = format!(
+        "http://{}:{}/agents/{}/activate",
+        registrar_ip, registrar_port, agent_uuid
+    );
+    let resp = reqwest::Client::new().put(&addr).json(&data).send().await?;
+
+    if !resp.status().is_success() {
+        return Err(anyhow!(
```
lkatalin, line 97: Maybe it makes sense to use one of our custom error types here instead of `anyhow`? I could see a case for either.

src/main.rs
```diff
@@ -68,11 +103,57 @@ async fn main() -> Result<()> {
...
-    info!("Starting server...");
+    let registrar_ip =
+        config_get("/etc/keylime.conf", "registrar", "registrar_ip")?;
+    let registrar_port =
+        config_get("/etc/keylime.conf", "registrar", "registrar_port")?;
+    let agent_uuid_config =
+        config_get("/etc/keylime.conf", "cloud_agent", "agent_uuid")?;
+    let agent_uuid = get_uuid(&agent_uuid_config);
+
+    {
```
lkatalin, line 127: @puiterwijk What is the purpose of this bracket? I believe this is from your changes.

src/registrar_agent.rs
lkatalin: I ended up using a custom type so that we can check the status codes.

lkatalin:
I've added some unit testing for `do_register_agent` and `do_activate_agent`. I ended up using `[wiremock](https://docs.rs/wiremock/0.5.1/wiremock/)` for this, as it's better documented than `[reqwest_mock](https://docs.rs/reqwest_mock/0.7.0/reqwest_mock/)` and seems to have more features (not just a client, etc.).

src/main.rs
lukehinds: that does not look like a strange scope, is the compiler not complaining about this?

lukehinds (approved):

src/main.rs
puiterwijk: This was basically because this activation step is in the `main()` function, which keeps running until the entire agent terminates.
And I didn't want to have secrets lingering in memory during the full runtime.

src/main.rs
```diff
@@ -71,9 +106,54 @@ async fn main() -> Result<()> {
...
+        let auth_tag = hex::encode(&auth_tag);
+
+        registrar_agent::do_activate_agent(
+            &registrar_ip,
+            &registrar_port,
+            &agent_uuid,
+            &auth_tag,
+        )
+        .await?;
+    }
```
puiterwijk, line 155: This is the closing bracket: after we activate, we no longer need `keyblob`.

lkatalin: Yeah, I did notice it created its own scope that lasted until this line, I just wasn't sure why.

PR #159, opened 2021-02-10
Call revocation actions

lkatalin:
This code runs those revocation scripts which were sent by the tenant and can be accessed from the secure mount point post-attestation.

I've decided not to check whether the revocation scripts begin with `local_action`, since they are all listed in `action_list` anyhow. But if someone has a reason that they should be marked that way, I can add this check back again.

I had a few other thoughts and questions inline marked with `TODO` and would love some suggestions on those points. The most important being: If these scripts don't run for whatever reason (ex. the Python version was too old), is that detected in some other part of the code? It seems like we would want a guarantee that a node can't continue to be in the cluster if it doesn't run the scripts, since it could still be in communication with the compromised node, for example.

### DIFF
--- a/src/revocation.rs
+++ b/src/revocation.rs
@@ -7,9 +7,150 @@ use crate::error;
 use crate::error::{Error, Result};
 use crate::secure_mount;
 
+use std::io::Write;
 use std::path::Path;
+use std::process::{Child, Command, Output, Stdio};
 
 use serde_json::Value;
+
+/// Runs a script with a json value as argument (used for revocation actions)
+pub(crate) fn run_action(
+    dir: &Path,
+    script: &str,
+    json: Value,
+) -> Result<Output> {
+    let raw_json = serde_json::value::to_raw_value(&json)?;
+
+    let mut child = match Command::new(format!("{}{}", "./", script))
+        .current_dir(dir)
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+    {
+        Ok(child) => child,
+        Err(e) => {
+            let msg = format!(
+                "ERROR: failed to run revocation action: {}, received: {}",
+                script, e
+            );
+            error!("{}", msg);
+            return Err(Error::Other(msg));
+        }
+    };
+
+    let msg = format!(
+        "ERROR: failed writing to stdin on revocation action: {:?}",
+        script
+    );
+    if let Err(e) = child
+        .stdin
+        .as_mut()
+        .expect(&msg)
+        .write_all(raw_json.get().as_bytes())
+    {
+        error!("{}", msg);
+        return Err(Error::Other(msg));
+    }
+
+    let output = match child.wait_with_output() {
+        Ok(output) => output,
+        Err(e) => {
+            let msg = format!("ERROR: failed to wait on child process while running revocation action: {:?}", script);
+            error!("{}", msg);
+            return Err(Error::Other(msg));
+        }
+    };
+
+    if !output.status.success() {
+        let code = match output.status.code() {
+            Some(code) => code.to_string(),
+            None => {
+                "no code; process likely terminated by signal".to_string()
+            }
+        };
+        let stdout = String::from_utf8(output.stdout)?;
+        let stderr = String::from_utf8(output.stderr)?;
+
+        let mut msg = format!(
+            "ERROR: revocation action {} returned with {}\n",
+            script, code
+        );
+
+        if !stdout.is_empty() {
+            msg = format!(
+                "{}{}",
+                msg,
+                format!(
+                    "ERROR: revocation action {} stdout: {:?}",
+                    script, stdout
+                )
+            );
+        }
+
+        if !stderr.is_empty() {
+            msg = format!(
+                "{}{}",
+                msg,
+                format!(
+                    "ERROR: revocation action {} stderr: {:?}",
+                    script, stderr
+                )
+            );
+        }
+
+        error!("{}", msg);
+        return Err(Error::Other(msg));
+    }
+
+    info!(
+        "{}",
+        format!("INFO: revocation action {:?} successful", script)
+    );
+    Ok(output)
+}
+
+/// Runs revocation actions received from tenant post-attestation
+pub(crate) fn run_revocation_actions(
+    json: Value,
+) -> Result<Vec<Result<Output>>> {
+    #[cfg(not(test))]
+    pretty_env_logger::init();
+
+    #[cfg(test)]
+    let mount = concat!(env!("CARGO_MANIFEST_DIR"), "/tests");
+
+    #[cfg(not(test))]
+    let mount = secure_mount::mount()?;
+    let unzipped = format!("{}/unzipped", mount);
+    let action_file = format!("{}/unzipped/action_list", mount);
+
+    let mut outputs = Vec::new();
+
+    if Path::new(&action_file).exists() {
+        let action_data = std::fs::read_to_string(action_file)
+            .expect("unable to read action_list");
+
+        let action_list = action_data
+            .split('\n')
+            .filter(|&script| !script.is_empty())
+            .map(|script| script.trim())
+            .collect::<Vec<&str>>();
+
+        if !action_list.is_empty() {
+            for action in action_list {
+                let output =
+                    run_action(&Path::new(&unzipped), action, json.clone());
+                outputs.push(output);
+            }
+        } else {
+            warn!("WARNING: no actions found in revocation action list");
+        }
+    } else {
+        warn!("WARNING: no action_list found in secure directory");
+    }
+    Ok(outputs)
+}
 
 /// Handles revocation messages via 0mq
 /// See:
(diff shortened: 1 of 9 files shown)


src/revocation.rs
```diff
@@ -8,9 +8,99 @@ use crate::error::{Error, Result};
...
+    // otherwise this could be an attack vector
+    let python_version = String::from_utf8(
+        Command::new("python")
+            .arg("-V")
+            .output()
+            .expect("could not find Python in $PATH")
+            .stdout,
+    )?;
+
+    if python_version.split_whitespace().collect::<Vec<&str>>()[1]
```
ashcrow, line 64: nit: It would be ideal to have this occur once when the agent initializes.

src/revocation.rs
```diff
@@ -8,9 +8,99 @@ use crate::error::{Error, Result};
...
+                return Err(Error::Other(msg));
+            }
+        };
+
+    if !status.success() {
+        let msg = format!(
+            "ERROR: revocation action {:?} returned with status {:?}",
+            script, status
+        );
+        error!("{}", msg);
```
ashcrow, line 36: Is there any output to `stdout` or `stderr` we may need to capture?

src/revocation.rs
```diff
@@ -8,9 +8,99 @@ use crate::error::{Error, Result};
...
 
 use std::path::Path;
+use std::process::Command;
 
 use serde_json::Value;
 
+/// Runs a python script (used for revocation actions)
+pub(crate) fn run_python(dir: &Path, script: &str) -> Result<()> {
+    let status =
+        // TODO: Should we assume that these script names are all correctly
```
ashcrow, line 18: My $0.02: We should assume the end in `.py` as in the future it's _possible_ there could be actions that are in other languages. AFAIK there is no plans for that but I think assuming `.py` for python scripts that are run directly via interpreter is reasonable.

ashcrow (commented):
Looks like great progress!! :smile:

lukehinds (commented):
This is looking good to me. It would be nice to have some tests, just trying to think how we can do that. Not sure if we need to mock or can rely on the CI containers having a py env.

src/revocation.rs
lkatalin: Great idea - this takes care of my worry about the scripts not executing, also. I've moved it to its own function `check_system_prerequisites()` in `main.rs`.

src/revocation.rs
lkatalin: Also a great idea, I added this. Thanks!

src/revocation.rs
lkatalin: Sounds good to me. I will assume if there are any non-Python scripts, they will be sorted appropriately in `run_revocation_actions()` in the future.

lkatalin:
> This is looking good to me. It would be nice to have some tests, just trying to think how we can do that. Not sure if we need to mock or can rely on the CI containers having a py env.

@lukehinds This is true. I went ahead and added a unit test (may add a few more error cases as well before this is merged). I think we should add Python to all relevant containers if it isn't there already, as it will be necessary to run Keylime (see comment above), including future integration tests. Right now it looks like they are passing with the Python dependency.

> This is looking good to me. It would be nice to have some tests, just trying to think how we can do that. Not sure if we need to mock or can rely on the CI containers having a py env.

@lukehinds This is a good point and I'm adding some unit tests. They pass locally, but as you pointed out, there is a problem with not having Python in the CI container. I'm going to try to fix this in a separate PR because it is proving to be tricky.

src/revocation.rs
```diff
@@ -118,8 +237,7 @@ pub(crate) async fn run_revocation_service() -> Result<()> {
                     "Revocation signature validated for revocation: {}",
                     msg_payload
                 );
-                // TODO: Implement callback
-                //callback(msg_payload)
+                let _ = run_revocation_actions(Some(msg_payload))?;
```
lkatalin, line 240: @ashcrow What was the idea behind passing `msg_payload` to this function in your previous commit? I've done this, but I don't end up using it. Is there something we should be parsing from the payload?

src/main.rs
```diff
@@ -52,13 +52,42 @@ use std::fs::File;
...
 use error::{Error, Result};
 
 static NOTFOUND: &[u8] = b"Not Found";
 
+// Runs checks required for Keylime to work on the system.
+fn check_system_prerequisites() -> Result<()> {
+    // Check Python version > 3.x.
+    // This is necessary for running revocation scripts.
+    let python_version = String::from_utf8(
+        Command::new("python")
```
ashcrow, line 66: Note: One more thing here I need to double check -- but on CoreOS system python is the only provided python. I'll double check to see if it's in PATH or if it has to be found directly. Either way I don't see it as a blocker as if we need a second or expanded check we can add it later.

src/revocation.rs
ashcrow: It comes from https://github.com/keylime/keylime/blob/master/keylime/revocation_notifier.py#L122 which leads to [this](https://github.com/keylime/keylime/blob/1d0f695c0848b001acb0ab01de46094494884506/keylime/keylime_agent.py#L630) and more importantly [this execution](https://github.com/keylime/keylime/blob/1d0f695c0848b001acb0ab01de46094494884506/keylime/keylime_agent.py#L592) where it finally runs and passes the content into the action. Looking at [this revocation action](https://github.com/keylime/keylime/blob/1d0f695c0848b001acb0ab01de46094494884506/keylime/revocation_actions/update_crl.py#L22) it does seem like it's expected to pass the data through.

ashcrow: In the python code it actually imports the code and calls the method directly. I think we will need to update the actions to be able to take the payload through `stdin` or something similar.

@mpeters @lukehinds @puiterwijk thoughts?

lkatalin: @ashcrow Thanks for looking into that. Yeah, it looks like in `execute()` it just checks that it's a revocation and then does nothing else with it?

One thing this PR hasn't dealt with, though, is the CRL, if we are expecting to do that.

src/main.rs
lkatalin: Hmm, this seems important - I'd like to add it.

ashcrow: Based on the discussion in our meeting earlier I have a feeling this won't be needed.

lkatalin:
@ashcrow @lukehinds I will have to figure out this CI failure tomorrow (not able to replicated it locally) and might add some more unit tests, but was wondering if you have other comments on how it works currently. Thanks.

lkatalin:
Tests passing again!

src/revocation.rs
```diff
@@ -8,9 +8,177 @@ use crate::error::{Error, Result};
 use crate::secure_mount;
 
 use std::path::Path;
+use std::process::{Child, Command, Output, Stdio};
 
 use serde_json::Value;
 
+// TODO: Add more types
+#[derive(Debug, Copy, Clone)]
+pub(crate) enum CmdType {
```
ashcrow, line 17: If we expect the command to be self contained per the last sync (IE: it can be directly executed) why differentiate?

src/revocation.rs
```diff
@@ -8,9 +8,135 @@ use crate::error::{Error, Result};
...
+pub(crate) fn run_action(
+    dir: &Path,
+    script: &str,
+    json: Value,
+) -> Result<Output> {
+    let raw_json = serde_json::value::to_raw_value(&json)?;
+
+    let child = match Command::new(format!("{}{}", "./", script))
+        .current_dir(dir)
+        .arg(raw_json.get())
```
ashcrow, line 25: I believe this would pass the data as if it was an argument on a commandline instead of it it going in through standard input. If it was in a shell it would look like:

```
./path/to/script {"hello": ["I", "am", "the", "payload"]}
```

What I think we need is to have the `stdin`. Which would be closer to the following shell example:

```
echo -n '{"hello": ["I", "am", "the", "payload"]}' | ./path/to/script 
```

I've not done this myself in rust before so I searched a bit and [this came up](https://stackoverflow.com/questions/49218599/write-to-child-process-stdin-in-rust) and looks like it's a good example:

src/revocation.rs
lkatalin: You're right, this was over-engineered. Thanks!

src/revocation.rs
lkatalin: Ah, I see what you're saying. Thanks for the example, I'll try it that way.

ashcrow:
Let us know once you're ready for a re-review!!

src/revocation.rs
```diff
@@ -7,10 +7,151 @@ use crate::error;
...
+    let mount = concat!(env!("CARGO_MANIFEST_DIR"), "/tests");
+
+    #[cfg(not(test))]
+    let mount = secure_mount::mount()?;
+    let unzipped = format!("{}/unzipped", mount);
+    let action_file = format!("{}/unzipped/action_list", mount);
+
+    let mut outputs = Vec::new();
+
+    if Path::new(&action_file).exists() {
```
ashcrow, line 130: future: It may make sense to load this once rather than when we run revocations. However, I believe this is fine and may be what we want long term depending on the expectations of the users.

ashcrow (approved):
:+1:  :smile:

lukehinds (approved):

lukehinds:
looks good to me , @lkatalin you can self merge if you like (just in case you have further tweaks).

PR #161, opened 2021-02-24
Override config location with environment variable

ashcrow:
`KEYLIME_CONFIG` can be set to override the default config file location.

*Note*:
I think we should eventually rework the configuration system so that we load the configuration once and then reload when we receive the proper signal. My reasoning is that the current set up means that if the file changes while the agent is running then the result could be old and new variables in use based on when they have been read. I've started to make structs for the keylime config. If folks agree I'll slowly but surely chip away at that no matter if this merges or not :smile:

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -4,6 +4,7 @@
 use crate::error::{Error, Result};
 use ini::Ini;
 use log::*;
+use std::env;
 
 /*
  * Constants and static variables
@@ -12,6 +13,7 @@ pub const STUB_VTPM: bool = false;
 pub const STUB_IMA: bool = true;
 pub const TPM_DATA_PCR: usize = 16;
 pub const IMA_PCR: usize = 10;
+pub static DEFAULT_CONFIG: &str = "/etc/keylime.conf";
 pub static RSA_PUBLICKEY_EXPORTABLE: &str = "rsa placeholder";
 pub static TPM_TOOLS_PATH: &str = "/usr/local/bin/";
 pub static IMA_ML_STUB: &str = "../scripts/ima/ascii_runtime_measurements";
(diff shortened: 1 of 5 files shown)


src/common.rs
```diff
@@ -20,14 +22,46 @@ pub static WORK_DIR: &str = "/tmp";
...
 /*
  * Input: config file location (e.g. /etc/keylime.conf), [section] and key
  * Return: Returns the matched key
  *
  * Example call:
- * let port = common::config_get("/etc/keylime.conf""general","cloudagent_port");
+ * let port = common::config_get("/etc/keylime.conf", "general","cloudagent_port");
  */
-pub(crate) fn config_get(
+pub(crate) fn explicit_config_get(
```
lkatalin, line 64: What if this function dropped its `conf_name` parameter and called `config_file_get()`? Then maybe we wouldn't need `config_get()` to wrap this one (and we could change the name back to `config_get()`)?

lkatalin (commented):
This is going to be a great improvement; thank you for working on this!

src/common.rs
ashcrow: I think that would be totally valid. My thought was originally to keep the older function around in case there was a need but if there isn't any expectation for future need I'm 100% on board with this!

ashcrow: Will update with this pattern.

ashcrow:
Hrm the failure looks like code I didn't touch ...

```
114 | / pub(crate) fn run_revocation_actions(
115 | |     json: Value,
116 | | ) -> Result<Vec<Result<Output>>> {
117 | |     #[cfg(not(test))]
...   |
152 | |     Ok(outputs)
153 | | }
    | |_^
```

I'm willing to look at it deeper if required though.

lkatalin:
> Hrm the failure looks like code I didn't touch ...
> 
> ```
> 114 | / pub(crate) fn run_revocation_actions(
> 115 | |     json: Value,
> 116 | | ) -> Result<Vec<Result<Output>>> {
> 117 | |     #[cfg(not(test))]
> ...   |
> 152 | |     Ok(outputs)
> 153 | | }
>     | |_^
> ```
> 
> I'm willing to look at it deeper if required though.

Looks like the failure is related to the revocation scripts (not sure why clippy didn't catch this on the earlier PR). I'm making a fix for this and then you can rebase on that. The changes LGTM!

lkatalin:
I went down a bit of a rabbit hole, but #162 should help once it is merged.

ashcrow:
Rebased on #162

ashcrow:
Woo! All green!

lkatalin (approved):
Looks great : )

ashcrow:
Rebased

ashcrow:
Merging per discussion.

PR #162, opened 2021-03-02
Update error handling for revocation scripts

lkatalin:
A clippy error on #161 indicated better ways of handling errors on revocation scripts. I've made updates here that should make the code more readable and make clippy happy. Also, added a unit test for an error case in the revocation scripts.

Tl;dr the error handling now largely takes place in `error.rs` and in the `run_revocation_actions()` function, which just adds some context (of which script was running) to an `Error::Execution` to create an `Error::Script`. This makes the `run_action()` function much shorter and gets rid of some redundant boilerplate code.


tests/unzipped/test_err.json
```diff
@@ -0,0 +1,3 @@
+{
+    "goodbye" : "there"
```
ashcrow, line 2: 😆👍

ashcrow (approved):

PR #165, opened 2021-03-05
Small fixes relating to Keylime components working together

lkatalin:
These all relate to running the `keylime_agent`, `keylime-verifier`, and `keylime-registrar` with `docker-compose`, but some of the fixes are also generally applicable. Backstory: I am trying to make sure the components can work together before writing the 3-party key protocol. These are changes I made while fixing some bugs or trying to understand them.

~~I am still debugging some errors, so there are probably a few more commits to be added. EDIT: But these could go in an immediate future PR as well.~~

~~Currently:
`ERROR keylime_agent::revocation   > Path for the 0mq socket: /tmp/secure/unzipped/RevocationNotifier-cert.crt doesn't exist`~~

### DIFF
--- a/docker/fedora/keylime_rust.Dockerfile
+++ b/docker/fedora/keylime_rust.Dockerfile
@@ -10,6 +10,7 @@ LABEL version="2.0.1" description="Keylime - Bootstrapping and Maintaining Trust
 
 # environment variables
 ARG BRANCH=master
+ENV KEYLIME_HOME ${HOME}/keylime
 ENV container docker
 COPY dbus-policy.conf /etc/dbus-1/system.d/
 
@@ -26,6 +27,7 @@ dbus \
 dbus-daemon \
 dbus-devel \
 dnf-plugins-core \
+efivar-devel \
 gcc \
 git \
 glib2-devel \
@@ -47,10 +49,16 @@ tpm2-abrmd \
 tpm2-tools \
 tpm2-tss \
 tpm2-tss-devel \
-uthash-devel"
+uthash-devel \
+czmq-devel"
 
 RUN dnf makecache && \
   dnf -y install $PKGS_DEPS && \
   dnf clean all && \
   rm -rf /var/cache/dnf/*
 
+# Move keylime.conf to expected location in /etc/
+WORKDIR ${KEYLIME_HOME}
+RUN git clone https://github.com/keylime/keylime.git && \
+cd keylime && \
+cp keylime.conf /etc/keylime.conf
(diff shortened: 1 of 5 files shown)


docker/fedora/keylime_rust.Dockerfile
```diff
@@ -47,10 +48,16 @@ tpm2-abrmd \
...
+czmq-devel"
 
 RUN dnf makecache && \
   dnf -y install $PKGS_DEPS && \
   dnf clean all && \
   rm -rf /var/cache/dnf/*
 
+# Move keylime.conf to expected location in /etc/
+WORKDIR ${KEYLIME_HOME}
+RUN git clone https://github.com/keylime/keylime.git $KEYLIME_HOME
```
ashcrow, line 61: nit: consider combining the RUN with `&& \` so it's one layer.

lkatalin: Good point, thanks.

ashcrow (commented):
Looks good so far. Let us know if you wish to add those extra commits referenced or if you'd prefer to put them through in a different PR!

lkatalin:
@ashcrow I think I'd actually like to merge it and add more in a bit. It would help with keeping track of branches while I test some things.

src/secure_mount.rs
```diff
@@ -111,7 +120,14 @@ pub(crate) fn mount() -> Result<String> {
                         .map(|path| {
                             info!("Changed path {} owner to root.", path);
                         })
-                        .map_err(|_| Error::SecureMount)?;
+                        .map_err(|e| {
+                            Error::SecureMount(
+                                format!(
+                                    "unable to change secure path dir owner to root: received exit code {}",
+                                    e.code().unwrap().unwrap() //#[allow_ci]
```
ashcrow, line 127: _Generally_ we want to avoid `unwrap` but there may be cases where it's fine. It may be worth noting in a comment why it's in use in these cases.

lkatalin: Completely agree. I don't think it can be avoided here, as they are `Option` types (but if you know a way, let me know). I added comments as suggested.

lkatalin:
@ashcrow Sorry to push a new version after you approved! I took your point about the `unwrap()` comments, and made a few small changes in the `dependencies` and `secure mount` commits. Also I had forgotten that I was still looking for input on the meaning of the comment on the `MOUNT_SECURE` flag [here](https://github.com/keylime/rust-keylime/pull/165/commits/2970c6cd80f63fcfb4803c0cce3eacabe880aa68#diff-b26887fc4e146bce39c62a2a1738f20dfd42357cbb76fe40708d259d8d6bb166L65). The original comment didn't seem to match the flag, so I made an update, but I want to know I am not misunderstanding something before it is merged.

lukehinds (approved):

lkatalin:
Merging this, as it had gotten two approvals, passed CI, and the update I pushed was just a comment.

PR #172, opened 2021-03-18
WIP to add keylime_tenant

lkatalin:
This is a draft. The goal is to get a tenant container to kick off provisioning the agent.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,15 @@
 version: '3.6'
 services:
+  keylime-tenant:
+    build:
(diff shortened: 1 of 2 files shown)


THS-on:
@lkatalin is there still a plan to implement this?  Most of the agent e2e is covered by https://github.com/RedHat-SP-Security/keylime-tests the question is if we want to implement tests separate from that.

lkatalin:
@THS-on Thanks for asking, I think we can close for now unless a future need comes up.

Issue #170, opened 2021-03-16
Activate credential error 2337


lkatalin:
```
2f1-4a97-9ef7-75bd81c00000
keylime_agent_1      | WARNING:esys:src/tss2-esys/api/Esys_ActivateCredential.c:321:Esys_ActivateCredential_Finish() Received TPM Error 
keylime_agent_1      | ERROR:esys:src/tss2-esys/api/Esys_ActivateCredential.c:105:Esys_ActivateCredential() Esys Finish ErrorCode (0x00000921) 
keylime_agent_1      | Error: TPM(Tss2Error(FormatZero(FormatZeroResponseCode { .0: 2337, error_number: 33, format_selector: false, version: true, tcg_vendor_indicator: false, severity: true })))

```

lkatalin:
This is the full error with message and error kind:

```
Error: TPM { fullerror: Tss2Error(FormatZero(FormatZeroResponseCode { .0: 2337, error_number: 33, format_selector: false, version: true, tcg_vendor_indicator: false, severity: true })), kind: Some(Lockout), rc_message: "authorizations for objects subject to DA protection are not allowed at this time because the TPM is in DA lockout mode" }

```

lkatalin:
@puiterwijk Any ideas on this one? Searching around seems to indicate it may be related to not cleanly shutting down the TPM and could be fixed with a "clear" setting in the BIOS, but I am not sure.

lkatalin:
@puiterwijk Pointed out the following solutions (thanks!):

1. Make sure containers are removed after exiting, otherwise they may still be using the TPM and cause a lockout. I'm able to do this for the agent by running `docker-compose run --rm keylime_agent` (another way would be to exit with `docker-compose down`)
2. Clear the lockout (inside the container) with `tpm2_dictionarylockout --clear-lockout`
3. Use `tpm2_dictionarylockout --setup-parameters --recovery-time 0`(ONLY for testing, NOT to be used on systems that require real TPM security!)

Removing the containers seems to be working atm.

lkatalin:
If any of the above commands don't work, then try:
`sudo tpm2_changeauth -c owner`
`sudo tpm2_changeauth -c endorsement` or `sudo tpm2_changeauth -c endorsement -p keylime`

After each command, check the status with `echo $?` to make sure it succeeded (code `0`). It will print error messages even when it succeeds, so you must use `echo $?` to check.

lkatalin:
[...]

PR #179, opened 2021-04-05
Add Identity and Integrity Quote functionality

lkatalin:
This is a draft to return the identity Quote to the tenant. Outstanding todos:
- ~~Is it okay to generate the NK?~~
- ~~Is it okay to regenerate the AIK?~~
- How to get the [`TPM2B_ATTEST`](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi-sys/src/bindings/x86_64-unknown-linux-gnu.rs#L3116) and [`Signature`](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/utils/mod.rs#L722) data types + the `NKpub` (which is an `openssl::rsa::Rsa<Public>`) into a format acceptable to [`actix_web::dev::Body`](https://docs.rs/actix-web/3.3.2/actix_web/dev/enum.Body.html)? This looks like it can be a String or bytes or maybe a `serde_json::Value`. But are these TSS ESAPI types compatible with any of those?

### DIFF
--- a/src/quotes_handler.rs
+++ b/src/quotes_handler.rs
@@ -1,8 +1,15 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2021 Keylime Authors
 
+use crate::{
+    error::Error, registrar_agent::serialize_as_base64, tpm, QuoteData,
+};
 use actix_web::{web, HttpResponse, Responder};
-use serde::Deserialize;
+use log::*;
+use openssl::pkey::{PKey, Private, Public};
+use serde::{Deserialize, Serialize};
+use std::{convert::TryInto, sync::Mutex};
+use tss_esapi::Context;
 
 #[derive(Deserialize)]
 pub struct Ident {
@@ -17,7 +24,28 @@ pub struct Integ {
     partial: String,
 }
 
-pub async fn identity(param: web::Query<Ident>) -> impl Responder {
+// TODO: Is this how things will be parsed on the tenant side?
+// Reference: https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_main.py#L1014
+// TODO: What is the PCR blob supposed to be?
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Quote<'a> {
+    #[serde(serialize_with = "serialize_as_base64")]
+    pub quote: &'a [u8],
+    #[serde(serialize_with = "serialize_as_base64")]
+    pub signature: &'a [u8],
+    #[serde(serialize_with = "serialize_as_base64")]
+    pub pcrblob: &'a [u8],
+    #[serde(serialize_with = "serialize_as_base64")]
+    pub nk_pub: &'a [u8],
+}
+
+// This is a Quote request from the tenant, which does not check
+// integrity measurement. It should return this data:
+// { QuoteAIK(nonce, 16:H(NK_pub)), NK_pub }
+pub async fn identity(
+    param: web::Query<Ident>,
+    data: web::Data<QuoteData>,
+) -> impl Responder {
     // nonce can only be in alphanumerical format
     if !param.nonce.chars().all(char::is_alphanumeric) {
         HttpResponse::BadRequest()
@@ -27,17 +55,23 @@ pub async fn identity(param: web::Query<Ident>) -> impl Responder {
             ))
             .await
     } else {
-        // place holder for identity quote code
-        HttpResponse::Ok()
-            .body(format!(
-                "Calling Identity Quote with nonce: {}",
-                param.nonce
-            ))
-            .await
+        info!("Calling Identity Quote with nonce: {}", param.nonce);
+
+        let idquote = tpm::quote(param.nonce.as_bytes(), None, data)?;
+
+        HttpResponse::Ok().json(idquote).await
     }
 }
 
-pub async fn integrity(param: web::Query<Integ>) -> impl Responder {
+// This is a Quote request from the cloud verifier, which will check
+// integrity measurement. The PCRs inclued in the Quote will be specified
+// by the mask, vmask. It should return this data:
+// { QuoteAIK(nonce, 16:H(NK_pub), xi:yi), NK_pub}
+// where xi:yi are additional PCRs to be included in the quote.
+pub async fn integrity(
+    param: web::Query<Integ>,
+    data: web::Data<QuoteData>,
+) -> impl Responder {
     // nonce, mask, vmask can only be in alphanumerical format
     if !param.nonce.chars().all(char::is_alphanumeric) {
         HttpResponse::BadRequest()
@@ -61,12 +95,12 @@ pub async fn integrity(param: web::Query<Integ>) -> impl Responder {
             ))
             .await
     } else {
-        // place holder for integrity quote code
-        HttpResponse::Ok()
-            .body(format!(
-                "Calling Integrity Quote with nonce: {}",
-                param.nonce
-            ))
-            .await
+        info!("Calling Itegrity Quote with nonce: {}", param.nonce);
+
+        // todo: mask or vmask?
+        let integrityquote =
+            tpm::quote(param.nonce.as_bytes(), Some(&param.mask), data)?;
+
+        HttpResponse::Ok().json(integrityquote).await
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -111,4 +111,6 @@ impl From<tss_esapi::Error> for Error {
     }
 }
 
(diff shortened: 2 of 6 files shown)


lkatalin:
I'm un-drafting this not because it's done, but because I could use reviews on this version of it.

lkatalin:
I updated this to get a few steps closer to what I think will be needed to send the data back to the tenant with serializing an `IdentQuote` struct (modeled on how we are returning data in `registrar_agent.rs` for activation and registration). Many questions remain about returning this data. Currently, sending any test JSON back to the tenant (ex. `{}`) results in a JSON parsing error on the tenant side. Still debugging this and could use any wisdom.

I also ended up changing some of the functions in `crypto.rs` to return and handle `PKey`s instead of `Rsa`, based on #181. And also have added some functionality in `tpm.rs`. These could be split off into their own PRs for the sake of organization.

src/quotes_handler.rs
```diff
@@ -27,17 +68,62 @@ pub async fn identity(param: web::Query<Ident>) -> impl Responder {
...
+                err,
+                kind: None,
+                message: format!("TPM identity quote error: {:?}", err),
+            })?;
+
+        let idquote = IdentQuote {
+            quote: "placeholderquote".as_bytes(),
+            signature: "placeholdersig".as_bytes(),
+            pcrblob: "placeholderpcrblob".as_bytes(),
+            nk_pub: &nk_pub.public_key_to_pem().unwrap(), //#[allow_ci] can't convert openssl ErrorStack to ResponseError
```
ashcrow, line 117: I'm OK with this as is since it's an iteration but could the `unwrap` be replaced with a `match` before this block with a rewrap of the error?

lkatalin:
I broke off parts of this into other PRs to keep things more tidy and make it easier to review. This should pass CI when it's rebased on some of those others.

lkatalin:
Needs to be rebased on other PRs. WIP.

lkatalin:
There remains some trouble with passing in a struct with a Mutex inside of it to actixweb. I'm going to try to figure this out tomorrow. Also need to add testing for the `read_mask()` function, and the biggest blocker is still how to pass back the data to the tenant and verifier.

lkatalin:
New version is up for review. The only incomplete part is the return of the data in the `Quote` structure, which requires being able to turn a `TPM2B_ATTEST` struct and a `Signature` struct into bytes, figuring out what the [PCR blob](https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_main.py#L1011) is supposed to be, and figuring out how the CV and tenant parse out the `NK_pub`.

lkatalin:
> New version is up for review. The only incomplete part is the return of the data in the `Quote` structure, which requires being able to turn a `TPM2B_ATTEST` struct and a `Signature` struct into bytes, figuring out what the [PCR blob](https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_main.py#L1016) is supposed to be, and figuring out how the CV and tenant parse out the `NK_pub`.

Update: the "PCR blob" appears to be a "PCR input file to save the list of PCR values that were included in the quote", as documented in [`tpm2_checkquote`](https://www.mankier.com/1/tpm2_checkquote), based on this [line](https://github.com/keylime/keylime/blob/149dffe7aa7ba8296689710854317545778d2414/keylime/tpm/tpm_main.py#L985) in Python Keylime's tpm file.

lkatalin:
This is the error from the tenant when trying to send back any JSON from the identity quote handler (even `"[]"`):

```
2021-04-21 17:53:47.156 - keylime.tenant - INFO - Agent D432FBB3-D2F1-4A97-9EF7-75BD81C00000 deleted from the CV
[root@localhost ~]# keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -f /root/excludes.txt --uuid D432FBB3-D2F1-4A97-9EF7-75BD81C00000 --allowlist /root/allowlist.txt --exclude /root/excludes.txt -c add
Using config file /etc/keylime.conf
2021-04-21 17:54:09.805 - keylime.tpm - INFO - TPM2-TOOLS Version: 4.3.0
2021-04-21 17:54:09.807 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2021-04-21 17:54:09.807 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x408000
2021-04-21 17:54:09.807 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x808000
/usr/lib/python3.9/site-packages/urllib3/connectionpool.py:997: InsecureRequestWarning: Unverified HTTPS request is being made to host '127.0.0.1'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  warnings.warn(
  File "/usr/local/bin/keylime_tenant", line 33, in <module>
    sys.exit(load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_tenant')())
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/cmd/tenant.py", line 15, in main
    tenant.main()
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/tenant.py", line 1048, in main
    mytenant.do_quote()
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/tenant.py", line 765, in do_quote
    raise e
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/tenant.py", line 750, in do_quote
    response_body = response.json()
  File "/usr/lib/python3.9/site-packages/requests/models.py", line 898, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib64/python3.9/site-packages/simplejson/__init__.py", line 525, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.9/site-packages/simplejson/decoder.py", line 373, in decode
    raise JSONDecodeError("Extra data", s, end, len(s))
simplejson.errors.JSONDecodeError: Extra data: line 1 column 10 - line 1 column 25 (char 9 - 24)
```

mpeters:
@lkatalin can you capture the exact HTTP response that's sent to the server side? Either put some debug code on the server side before it does the parsing of the JSON or maybe use wireshark to intercept in between.

lkatalin:
> @lkatalin can you capture the exact HTTP response that's sent to the server side? Either put some debug code on the server side before it does the parsing of the JSON or maybe use wireshark to intercept in between.

Wireshark was a great idea. It looks like the agent itself may be returning a `500: Internal Server Error` to the quote request, though I could be wrong as I'm just learning to use Wireshark today. Compare the good response from the Python agent with `HTTP/JSON` ...

pythonagentresponse

...vs. the bad response from the Rust agent:

rustagentresponse

This is strange as the actix_web handler returns an `HttpResponse::Ok` in the code. I actually do have an idea of one thing that could be causing it, but I can't try it at the moment because using the Python agent locked me out of the TPM again and none of the commands that worked to unlock it last time (documented [here](https://github.com/keylime/rust-keylime/issues/170)) are working.

lukehinds:
I suspect this might be from a type disparity, I run into these all the time when trying to get a python based app to talk with a rust/go/c etc app due to pythons uni / byte code handling

We could try printing out `response_body = response.json()` (line 750 in tenant.py) and then scrutinise the different structure / encoding of the reply from both the python agent and the rust agent. Ping me on hangouts if you need someone to grab the python side.

lkatalin:
After unlocking the TPM with `tpm2_clear` (all of the other commands from [here](https://github.com/keylime/rust-keylime/issues/170) somehow did not work), I reverse engineered the Python packets further and managed to get past the JSON error by sending these data structures with hardcoded data that I got from the Python packets:

```
#[derive(Serialize)]
struct JsonWrapper {
    code: u32, // must be hand-coded for some reason instead of coming from actual web data?
    status: String, // this too
    results: MyObj
}

#[derive(Serialize)]
struct MyObj {
    quote: String,
    hash_alg: String,
    enc_alg: String,
    sign_alg: String,
    pubkey: String,
}
```

I have no idea yet how this relates to the Quote + Sig + PCRblob parsed [here](https://github.com/keylime/keylime/blob/149dffe7aa7ba8296689710854317545778d2414/keylime/tpm/tpm_main.py#L1014). My guess is that all goes in the `quote` field.

Next I'll insert the actual values from a current quote instead of hardcoded values and see if I can get past the "invalid quote" error.

lkatalin:
Closing as this has been superseded by #195 and #196

PR #181, opened 2021-04-06
Pass a PKeyRef to asym_verify

puiterwijk:
Ensure we no longer need to clone the RSA key every time we want to
validate anything.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>


puiterwijk:
A fix for the clippy lint is in #182

lkatalin (approved):
:+1:

PR #182, opened 2021-04-06
Fix uppercase acronyms and use thiserror From derives

puiterwijk:


lkatalin (approved):
Ah, this is great! I also just fixed these acronyms in my draft PR. : ) I like shortening the `from` here as well. I can rebase on top of this.

PR #183, opened 2021-04-19
Add scoping to attestation server threads

lkatalin:
The identity quote and ukey requests need to share state (the ephemeral key pair, NK; the client will use the public key embedded in the quote response to encrypt the U or V key it is returning). Likewise for the integrity quote and vkey requests. 
However, they should not share this state with the main thread or with each other. This splits the actix_server into smaller pieces to scope them accordingly.

Alternatives:
Creating actix [middleware](https://actix.rs/docs/middleware/) to pass data from one service to another.

To do: If this is a solution we want to use, make a `create_server()` function to reduce redundant code.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -155,12 +155,10 @@ async fn main() -> Result<()> {
         info!("SUCCESS: agent activated");
     }
 
-    let actix_server = HttpServer::new(move || {
+    // Service identity quote requests from the tenant
+    let identity_server = HttpServer::new(move || {
         App::new()
-            .service(
-                web::resource("/keys/verify")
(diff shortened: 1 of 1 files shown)


lukehinds (commented):
Would you not get Address / Port already in use errors binding to the same IP / port for the multiple binds you have?

lkatalin:
I'm closing this because, as @puiterwijk pointed out, we will be keeping the U key in memory anyway, so a threat model where someone gains access to NK (used to encrypt U and V in transit) is not really useful and adds complication. I will just pass NK in from the main thread.

PR #184, opened 2021-04-19
Use PKeys consistently

lkatalin:
Following on #181, use the openssl crate's PKey type instead of the inner
RSA type. Also, add additional key-related helper functions. (These make the future identity quote code less messy.)

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -4,14 +4,14 @@
 // use super::*;
 use openssl::hash::MessageDigest;
 use openssl::pkcs5;
-use openssl::pkey::{PKey, PKeyRef, Private, Public};
+use openssl::pkey::{Id, PKey, PKeyRef, Private, Public};
 use openssl::rsa::{Padding, Rsa};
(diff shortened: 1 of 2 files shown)


src/crypto.rs
```diff
@@ -41,21 +41,42 @@ pub(crate) fn do_hmac(
...
+        Id::RSA => {
+            let rsa = Rsa::from_public_components(
+                privkey.rsa()?.n().to_owned()?,
+                privkey.rsa()?.e().to_owned()?,
+            )
+            .map_err(crate::Error::Crypto)?;
+            PKey::from_rsa(rsa).map_err(crate::Error::Crypto)
+        }
+        _ => {
+            unimplemented!()
```
lukehinds, line 77: Might be worth marking a TODO to handle this with an Err instead when we are near release.

lukehinds (approved):

src/crypto.rs
lkatalin: Done.

PR #193, opened 2021-05-13
Add marshaling functions for TPM structs

### LINKED ISSUE #192: Rewrite TPM marshaling functions as macros
There is a lot of redundant code around marshaling the various TPM structs. These cases could probably be covered by a macro or perhaps some more generic function or trait.

Relevant functions in `tpm.rs` include:
`tpm_digest_to_vec`
`tpm_sig_to_vec`
`tpm_att_to_vec`

There is also marshaling in `create_ek()` that could use the future function.

### LINKED ISSUE #194: PCR output file is incorrect
Even basing it on [this function](https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_quote.c#L47-L91), two errors are returned:

`ERROR: Failed to read PCR selection from file` from this [line](https://github.com/tpm2-software/tpm2-tools/blob/f152cd0777ff6874389eef7ef94f2e95933e1f54/tools/misc/tpm2_checkquote.c#L260)
(from the identity quote handler)

`ERROR: Malformed PCR file, pcr count cannot be greater than 32, got: 4294967295` from this [line](https://github.com/tpm2-software/tpm2-tools/blob/f152cd0777ff6874389eef7ef94f2e95933e1f54/tools/misc/tpm2_checkquote.c#L271)
(from the integrity quote handler)

It's strange that these errors are different, as they are handled by the same code.

lkatalin:
Fixes #192 
Also fixes #194 with many thanks to @puiterwijk for debugging

**### Please start with reviewing this PR before the others!**

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -21,12 +21,16 @@ use tss_esapi::{
         session_handles::AuthSession,
     },
     structures::{
-        Digest, DigestValues, EncryptedSecret, IDObject, Name, PcrSlot,
+        Digest, DigestValues, EncryptedSecret, IDObject, Name,
+        PcrSelectionList, PcrSlot,
     },
     tss2_esys::{
-        Tss2_MU_TPM2B_PUBLIC_Marshal, TPM2B_PUBLIC, TPMS_SCHEME_HASH,
+        Tss2_MU_TPM2B_DIGEST_Marshal, Tss2_MU_TPM2B_PUBLIC_Marshal,
+        Tss2_MU_TPMT_SIGNATURE_Marshal, TPM2B_PUBLIC, TPML_DIGEST,
+        TPML_PCR_SELECTION, TPMS_SCHEME_HASH, TPMT_SIGNATURE,
         TPMT_SIG_SCHEME, TPMU_SIG_SCHEME,
     },
+    utils::{PcrData, Signature},
     Context, Tcti,
 };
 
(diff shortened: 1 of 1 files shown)


lkatalin:
@lukehinds @puiterwijk @ashcrow Not sure if this will ping you since it's been open for a while, but switched from draft because it's ready for review whenever someone has a few moments. Thanks!

src/tpm.rs
```diff
@@ -83,26 +87,85 @@ pub(crate) fn create_ek(
...
+// These marshaling functions use the macro above and are based on this format:
+// https://github.com/fedora-iot/clevis-pin-tpm2/blob/main/src/tpm_objects.rs#L64
+// ... and on these marshaling functions:
+// https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi-sys/src/ \
+// bindings/x86_64-unknown-linux-gnu.rs#L16010
+//
+// Functions can be created using the following form:
+// create_marshal_fn!(name_of_function_to_create, struct_to_be_marshaled, marshaling_function);
+//
+create_marshal_fn!(pub_to_vec, TPM2B_PUBLIC, Tss2_MU_TPM2B_PUBLIC_Marshal);
```
ashcrow, line 135: I confuse myself with macros. Is `create_marshal_fn` creating `pub_to_vec` which is used later?

lkatalin: @ashcrow Yeah, exactly - there are slightly different functions that all have the same basic structure. In all of the functions, some struct `TPM2B_something` needs to be turned into a vector and has a corresponding marshaling function. The macro can create all of these functions without replicating a bunch of code with only slight variations.

`pub_to_vec` and the other functions created here are used in the next PR on top of this one.

ashcrow (approved):
Seems reasonable to me!

PR #195, opened 2021-06-07
Add Quote functionality

lkatalin:
- [x] This should pass CI when rebased onto #193

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -2,33 +2,40 @@
 // Copyright 2021 Keylime Authors
 
 use std::convert::{TryFrom, TryInto};
+use std::io::prelude::*;
 use std::str::FromStr;
 
-use crate::{common::config_get, Error as KeylimeError, Result};
+use crate::{
+    common::config_get, quotes_handler::KeylimeIdQuote,
+    Error as KeylimeError, QuoteData, Result,
+};
+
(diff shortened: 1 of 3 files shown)


lkatalin:
@lukehinds @ashcrow Ready for review!

src/tpm.rs
```diff
@@ -436,6 +443,157 @@ pub(crate) fn read_mask(mask: &str) -> Result<Vec<PcrSlot>> {
...
+    pcr_data: PcrData,
+) -> Result<String> {
+    // marshal structs to vec in expected formats. these formats are
+    // dictated by tpm2_tools.
+    let att_vec = &att.attestationData[0..att.size as usize];
+    let sig_vec = sig_to_vec(sig.try_into()?);
+    let pcr_vec = pcrdata_to_vec(pcrs_read, pcr_data);
+
+    // zlib compression
+    let mut att_comp = ZlibEncoder::new(Vec::new(), Compression::default());
```
ashcrow, line 467: nit: This may be something to turn into a function since the same steps happen multiple times. Not a required change though.

ashcrow (approved):

src/tpm.rs
lkatalin: Could do in a future commit!

PR #196, opened 2021-06-07
Add handling for Identity and Integrity quotes

lkatalin:
This should pass CI when rebased on #195 

Thoughts: the JsonWrapper and KeylimeQuote variations are completely based on the Python interface, so they could possibly go in their own interface module. On the other hand, there isn't much code here so it may not be necessary to separate it that way. Opinions welcome.

Also: will we ever need to use the vmask? From what I understand, it is only necessary when tying a vtpm value to a hardware tpm value, which isn't used anymore in the Python codebase? But I could be wrong here.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -57,6 +57,8 @@ pub(crate) enum Error {
     Other(String),
 }
 
+impl actix_web::ResponseError for Error {}
+
 impl Error {
     pub(crate) fn http_code(&self) -> Result<u16> {
         match self {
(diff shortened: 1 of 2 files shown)


lkatalin:
@lukehinds @ashcrow @puiterwijk This is the next one when you get a chance : )

ashcrow (approved):
The code looks familiar. Was this pulled out of an earlier PR?

lkatalin:
> The code looks familiar. Was this pulled out of an earlier PR?

Yes, it was. I had a longer PR earlier that I broke down into smaller self-contained pieces to make them easier to review and also make them easier to iterate on as I debug. Thanks for looking at it @ashcrow !

PR #197, opened 2021-06-07
Add U and V key handlers, and decrypt + run payload

### LINKED ISSUE #97: Implement Payload Handling Three Part Key Deviation
If an agent passes its attestation, the keylime tenant will provide an encrypted payload. This payload is securely transferred using a key generated by the tenant, and cryptographically split into two, half a key then goes to the verifier and half to the tenant. When an agent passes its attestation, the tenant and the verifier will be release key parts which the agent then combines into a single key that allows them to un-encrypt the payload.

### LINKED ISSUE #99: Implement Auto Run Script
Keylime auto executes a shell script transferred in the encrypted payload, once each machine passes its initial attestation checks.

### LINKED ISSUE #202: need to lazy load or reconsider how we load revocation
In the python agent they lazy load `RevocationNotifier-cert.crt`. The reason for this is that the agent cannot access `RevocationNotifier-cert.crt` (from secure/unzipped) until after its received the unzipped payload from the `keylime_tenant` first (which contains the `RevocationNotifier-cert.crt` file and others). The reason for this is you can't provision the agent to receive the payload unless its running first.

## debug example

```
RUST_LOG=keylime_agent=trace cargo run
   Compiling keylime_agent v0.1.0 (/root/rust-keylime)
    Finished dev [unoptimized + debuginfo] target(s) in 6.83s
     Running `target/debug/keylime_agent`
 INFO  keylime_agent > Starting server...
 INFO  keylime_agent > Misformatted UUID: neptune
 INFO  keylime_agent::registrar_agent > Sending data to http://127.0.0.1:8890/agents/ffd57add-d217-4a81-ade3-e140b648df3a
 INFO  keylime_agent                  > SUCCESS: agent registered
 INFO  keylime_agent                  > SUCCESS: agent activated
 INFO  keylime_agent                  > Listening on http://127.0.0.1:9002
 INFO  keylime_agent::secure_mount    > Using existing secure storage tmpsfs mount /tmp/secure
 INFO  keylime_agent::revocation      > Connecting to revocation endpoint at tcp://127.0.0.1:8992...
 ERROR keylime_agent::revocation      > Path /tmp/secure/unzipped/RevocationNotifier-cert.crt for the 0mq socket doesn't exist
Error: Configuration("Path /tmp/secure/unzipped/RevocationNotifier-cert.crt for the 0mq socket socket doesn't exist")
```

### LINKED ISSUE #203: Esys handle does not exist
When using a HW Infineon TPM 2.0 , I get the following exception from tss2-esys

```
RUST_LOG=keylime_agent=trace cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/keylime_agent`
 INFO  keylime_agent > Starting server...
 INFO  keylime_agent::registrar_agent > Sending data to http://127.0.0.1:8890/agents/d432fbb3-d2f1-4a97-9ef7-75bd81c00000
 INFO  keylime_agent                  > SUCCESS: agent registered
 INFO  keylime_agent                  > SUCCESS: agent activated
 INFO  keylime_agent                  > Listening on http://127.0.0.1:9002
 INFO  keylime_agent::secure_mount    > Using existing secure storage tmpsfs mount /tmp/secure
 INFO  keylime_agent::revocation      > Connecting to revocation endpoint at tcp://127.0.0.1:8992...
 INFO  keylime_agent::revocation      > Waiting for revocation messages on 0mq tcp://127.0.0.1:8992
 INFO  keylime_agent::quotes_handler  > Calling Identity Quote with nonce: vJZ2IuUl1y3knfZkT2FT
 INFO  keylime_agent::quotes_handler  > Calling Integrity Quote with nonce: VLIbA1dHNEFhjSTF76XD
ERROR:esys:src/tss2-esys/esys_iutil.c:1096:esys_GetResourceObject() Error: Esys handle does not exist (70018). 
ERROR:esys:src/tss2-esys/api/Esys_Quote.c:176:Esys_Quote_Async() signHandle unknown. ErrorCode (0x00070018) 
ERROR:esys:src/tss2-esys/api/Esys_Quote.c:82:Esys_Quote() Error in async function ErrorCode (0x00070018) 
ERROR:esys:src/tss2-esys/esys_iutil.c:1096:esys_GetResourceObject() Error: Esys handle does not exist (70018). 
ERROR:esys:src/tss2-esys/api/Esys_Quote.c:176:Esys_Quote_Async() signHandle unknown. ErrorCode (0x00070018) 
ERROR:esys:src/tss2-esys/api/Esys_Quote.c:82:Esys_Quote() Error in async function ErrorCode (0x00070018) 
```

### LINKED ISSUE #221: Allow payloads of varying length
Otherwise we get this error:
```
thread 'actix-rt:worker:0' panicked at 'source slice length (83) 
does not match destination slice length (392)', 
/rustc/5a7834050f3a0ebcd117b4ddf0bc1e8459594309/library/core/src/slice/mod.rs:3059:13
```

### LINKED ISSUE #226: Check keylime.conf before listening for revocations
If `[cloud-agent][listen_notfications = False]` then we shouldn't listen for revocations at all.

lkatalin:
- [x] I'll need to figure out how to avoid using
```
#![feature(string_remove_matches)]
#![feature(mutex_unlock)]
```
in order for CI to pass as these are unstable features. The former should be easy, the latter maybe not so much.

- [x] There are a few details still to fill in re: running the decrypted payload, checking if it's a zip file, etc.
~~The decrypted payload appears to be the excludes list; there is no `autorun.sh` sent.~~

Resolves #97 
Resolves #203 
Resolves #221 
Resolves #226 
Resolves #202 
Resolves #99


src/keys_handler.rs
```diff
@@ -2,26 +2,207 @@
...
+) -> Result<Vec<u8>, openssl::error::ErrorStack> {
+    // parse out payload iv, tag, ciphertext
+    let (iv, rest) = payload.split_at(AES_BLOCK_SIZE);
+    let (ciphertext, tag) = rest.split_at(rest.len() - AES_BLOCK_SIZE);
+
+    // note: it's unclear in the python codebase whether this is AES 128, but 128 seems
+    // to be the default
+    // https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#symmetric-encryption-modes
+    decrypt_aead(
+        Cipher::aes_128_gcm(),
```
puiterwijk, line 64:
```suggestion
        match symm_key.len() {
          16 => Cipher::aes_128_gcm(),
          32 => Cipher::aes_256_gcm(),
          _ => return Err(....),
        }       
```

src/keys_handler.rs
```diff
@@ -2,26 +2,207 @@
...
+// keylime/crypto.py#L189
+pub fn decrypt_payload(
+    payload: &[u8],
+    symm_key: &[u8],
+) -> Result<Vec<u8>, openssl::error::ErrorStack> {
+    // parse out payload iv, tag, ciphertext
+    let (iv, rest) = payload.split_at(AES_BLOCK_SIZE);
+    let (ciphertext, tag) = rest.split_at(rest.len() - AES_BLOCK_SIZE);
+
+    // note: it's unclear in the python codebase whether this is AES 128, but 128 seems
```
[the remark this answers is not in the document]
puiterwijk: Since I'm pretty sure that for `cryptography`, the default depends on the length of the key.

src/keys_handler.rs
lkatalin: Terrific idea.

src/keys_handler.rs
```diff
@@ -2,26 +2,207 @@
...
+    aes::AesKey,
+    encrypt::{Decrypter, Encrypter},
+    pkey::{PKey, Private, Public},
+    rsa::{Padding, Rsa},
+    symm::{decrypt_aead, Cipher},
+};
+
+use crate::{error::Error, QuoteData};
+
+// This value comes from and must match the Python codebase
```
puiterwijk, line 19: I think a better reference for this constant could be [FIPS 197](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.197.pdf), Introduction: `This standard specifies the Rijndael algorithm ([3] and [4]), a symmetric block cipher that can process data blocks of 128 bits, ...`. (128 bits / 8 bits-per-byte gives you 16)

src/keys_handler.rs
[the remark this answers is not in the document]
puiterwijk: (Also, given the fact that `ukey`, `vkey` and `sym_key` are 32 bytes, that means that this is 256 bit AES.

src/keys_handler.rs
```diff
@@ -2,26 +2,207 @@
...
+    let (ciphertext, tag) = rest.split_at(rest.len() - AES_BLOCK_SIZE);
+
+    // note: it's unclear in the python codebase whether this is AES 128, but 128 seems
+    // to be the default
+    // https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#symmetric-encryption-modes
+    decrypt_aead(
+        Cipher::aes_128_gcm(),
+        symm_key,
+        Some(iv),
+        &[0u8], // todo: check that we don't need additional auth data here?
```
puiterwijk, line 67: There is no additional data: that would be a call like `encryptor.authenticate_additional_data(associated_data)` in the `encrypt` function, or `decryptor.authenticate_additional_data(authenticated_data)` in `decrypt`.

src/keys_handler.rs
```diff
@@ -2,26 +2,207 @@
...
+        iv,
+        tag,
+    )
+}
+
+pub(crate) unsafe fn try_combine_keys() -> bool {
```
puiterwijk, line 73: All the `unsafe` stuff makes me feel a bit worried. Would it be a possibility to put this data as a fully locked `struct` field in the overall HTTP state?

src/keys_handler.rs
puiterwijk: Actually, a better approach: get rid of this constant. Use [Cipher.block_size](https://docs.rs/openssl/0.10.34/openssl/symm/struct.Cipher.html#method.block_size).

src/keys_handler.rs
lkatalin: Good point, I've put this in the app_data now.

lkatalin:
@puiterwijk @lukehinds @ashcrow This is updated, I'm able to run a small script with it. Still having the problem referenced in #202 but that is the fail point right now after the payload script runs.

lkatalin:
I can also split these commits out into their own PRs if that makes it easier to review; just let me know.

lkatalin:
> @puiterwijk @lukehinds @ashcrow This is updated, I'm able to run a small script with it. Still having the problem referenced in #202 but that is the fail point right now after the payload script runs.

Happy to say #202 is fixed now in this PR. To use the revocation feature and ensure that the revocation certificate is sent correctly, you must run the tenant with the following command (example directory and script shown here):

```
[root@localhost ~]# ls /root/senddir/
autorun.sh

[root@localhost ~]# keylime_tenant -v 127.0.0.1 -t 127.0.0.1 --uuid d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --allowlist /root/allowlist.txt --include /root/senddir --cert /root/ca --exclude /root/excludes.txt -c add
```

... where `autorun.sh` is any script you want for testing. You must also ensure that the following values are set in `/etc/keylime.conf`:

```
[cloud_agent]
...
extract_payload_zip = True

...
listen_notfications = True
...
payload_script=autorun.sh
```

I'll include this in some documentation.

src/main.rs
```diff
@@ -109,6 +120,158 @@ fn get_uuid(agent_uuid_config: &str) -> String {
...
+
+// checks if keylime.conf indicates the payload should be unzipped, and does so if needed.
+// the input string is the directory where the unzipped file(s) should be stored.
+pub(crate) fn optional_unzip_payload(unzipped: &str) -> Result<()> {
+    let is_zip = config_get("cloud_agent", "extract_payload_zip")?;
+    if bool::from_str(&is_zip.to_lowercase())? {
+        let zipped_payload = config_get("cloud_agent", "dec_payload_file")?;
+        info!("Unzipping payload {} to {}", zipped_payload, unzipped);
+
+        let mut status = Command::new("unzip")
```
ashcrow, line 254: There is nothing wrong with this, though I do want to ask: Why use the system command instead of a library?

ashcrow (commented):
One question but overall :+1:

lukehinds:
This LGTM, would you like me to test again first thing tomorrow @lkatalin ?

src/main.rs
lkatalin: @ashcrow I tried a library, but it was a bit more fiddly than I expected and this just worked. I suppose we could always update it on a second pass if we want to make some enhancements.

lukehinds: The zip_rs is not nice I agree, check out compress tools, I find it a lot simpler https://crates.io/crates/compress-tools

lkatalin:
Thanks @ashcrow and @lukehinds !

@lukehinds Thanks for the offer - if you test again with this branch, you will get the same intermittent quote error we saw at the end of last week. I bypassed that to test the U/V key handling behavior by setting the Python `tpm2_checkquote` function to return `true` so that I could at least build out the rest of the code following the quote check. So if you want to test the code in this branch, I could send you a branch of the Python code that I use for testing the interaction. If you want to test the whole system, we will need to fix the intermittent quote error first. I am about to give that a go, but could use help, as it seems to be esoteric TPM behavior.

src/main.rs
lkatalin: @lukehinds Sure will do, I can push a new version of that commit if I get that working.

lkatalin: The library seems straightforward, but I keep getting 

```
 INFO  keylime_agent                  > Unzipping payload decrypted_payload.zip to /tmp/secure/unzipped
Error: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })
```
Despite
```
[keylime-ima@localhost rust-keylime]$ ls /tmp/secure/unzipped/
decrypted_payload.zip  derived_tci_key
[keylime-ima@localhost rust-keylime]$ file /tmp/secure/unzipped/decrypted_payload.zip 
/tmp/secure/unzipped/decrypted_payload.zip: Zip archive data, at least v2.0 to extract
[keylime-ima@localhost rust-keylime]$ ls -la /tmp/secure/unzipped/decrypted_payload.zip 
-rw-r--r--. 1 root root 8738 Jul 12 14:02 /tmp/secure/unzipped/decrypted_payload.zip
```

The code:
```
let zipped_payload = config_get("cloud_agent", "dec_payload_file")?;
        info!("Unzipping payload {} to {}", &zipped_payload, &unzipped);

        let mut source = fs::File::open(&zipped_payload)?;    
        assert!(Path::new(&zipped_payload).exists());

        let dest = Path::new(&unzipped);
        assert!(Path::new(&dest).exists());

        let file_list = list_archive_files(&mut source).expect("failed to list files");
        info!("File list: {:?}", file_list);

        uncompress_archive(&mut source, &dest, Ownership::Ignore).expect("failed to unzip");
```

Still debugging...

lkatalin: Okay, it worked with a test example so it should be doable here shortly.

lkatalin: It says it succeeds, but then the unzipped files aren't there. All I can think of is maybe it isn't opening with the correct permissions, as it's apparently `r` only. This should be easy to update but for some reason is not.

lkatalin: I figured out the error... apparently `compress_tools::list_archive_files` mutates the `src` archive you give it so that there are no files left afterwards. Bizarre!

PR #214, opened 2021-07-05
Do not fail if EK cert is not present in TPM NV

ueno:
According to the spec (4.5.2), it is not mandatory that an EK
certificate is pre-provisioned in NVRAM; actually some TPM chips do
not have it, e.g., AMD fTPM, and the Python registrar tolerates
"ekcert: null".

### DIFF
--- a/src/registrar_agent.rs
+++ b/src/registrar_agent.rs
@@ -24,6 +24,19 @@ where
     })
 }
 
+fn serialize_maybe_base64<S>(
+    value: &Option<Vec<u8>>,
+    serializer: S,
+) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    match *value {
+        Some(ref value) => serializer.serialize_str(&base64::encode(value)),
+        None => serializer.serialize_none(),
+    }
+}
+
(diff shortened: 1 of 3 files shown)


lukehinds:
tagging @puiterwijk

src/registrar_agent.rs
```diff
@@ -40,11 +53,8 @@ fn is_empty(buf: &[u8]) -> bool {
 
 #[derive(Debug, Serialize, Deserialize)]
 struct Register<'a> {
-    #[serde(
-        serialize_with = "serialize_as_base64",
-        skip_serializing_if = "is_empty"
```
puiterwijk, line 45: Hah, right. I thought I had made the code check whether `ekcert` is in the json, but instead I made it assume that that's there and just check if that's json `null`.
Thanks for finding and fixing that!

puiterwijk (approved):
Thank you very much for this patch!

puiterwijk:
@ueno could you rebase it? Then me or anyone else with write access can merge this!

ueno:
Thank you for the review; rebased.

PR #230, opened 2021-07-13
Do not resend pubkey to CV after attestation

### LINKED ISSUE #211: Do not cause CV to resend V key
Currently after the integrity quote succeeds, the CV begins polling the agent at regular intervals and validating the quotes returned, which is expected. However, when the CV sends a request for an integrity quote, it always also re-sends a V key. My understanding is this key should only be sent once. So there is probably some piece of information the agent must send to the CV to get the CV to move into a polling state instead of a "send V key" state. Requires more investigation.

lkatalin:
This keeps the CV from resending the V key each time it polls for
an integrity quote.

Fixes #211

### DIFF
--- a/src/quotes_handler.rs
+++ b/src/quotes_handler.rs
@@ -3,7 +3,7 @@
 
 use crate::{tpm, Error as KeylimeError, QuoteData};
 
-use crate::common::IMA_ML;
+use crate::common::{IMA_ML, KEY_LEN};
 use actix_web::{web, HttpRequest, HttpResponse, Responder};
 use log::*;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


src/quotes_handler.rs
```diff
@@ -183,20 +228,34 @@ pub async fn integrity(
...
+                    data.pub_key
+                        .public_key_to_pem()
+                        .map_err(KeylimeError::from)?,
+                )
                 .map_err(KeylimeError::from)?,
-        )
-        .map_err(KeylimeError::from)?;
+            );
+            let response = JsonIntegWrapperPreAttestation::new(quote);
+            info!("GET integrity quote returning 200 response");
```
ashcrow, line 248: If I'm reading this correctly we will _always_ get a response of 200 since it's hard coded in the `JsonIntegWrapperPreAttestation`, correct? If that's true, why log on it? Same question on the `JsonIntegWrapperPostAttestation` log line later.

lkatalin: @ashcrow From my understanding and testing of the handler, if it encounters an error, it will return an HTTP code 500 Internal Server Error with the error message as part of the returned object. If it reaches this line, all the processing has succeeded, so we return 200.

ashcrow (approved):

PR #231, opened 2021-07-13
Run payload script from a shell

ueno:
Otherwise, it throws "Exec format error (os error 8)" unless it is a
native binary.  Also simplified the logic to set executable bit of the
script.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -214,14 +214,8 @@ pub(crate) fn run(dir: &str, script: &str, agent_uuid: &str) -> Result<()> {
         return Err(Error::Other(format!("{} not found", script_location)));
     }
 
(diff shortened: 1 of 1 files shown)


ashcrow (commented):
Changes look good to me. One question: will `test-output` file be left on the filesystem as a test side effect? If so, should we remove it or maybe add it to .gitignore?

ueno:
The file should be automatically removed, because `run` changes the directory to the tempdir before running the script (so `test-output` should be created inside tempdir).

lkatalin:
I like these changes also.

lkatalin (approved):

lkatalin:
@ashcrow I'll merge this since your question was addressed.

PR #232, opened 2021-07-14
Fix for race condition bug

### LINKED ISSUE #210: Assess severity of invalid quote bug
Intermittently, quote validation and/or PCR check (including IMA PCR check) fails in both the tenant and verifier. As there is sometimes success, it seems the crypto should be working correctly. This is what @puiterwijk and I tried to address with adding both SHA1 and SHA256 banks, see [reference](https://github.com/keylime/rust-keylime/pull/197/commits/53cedf9a11a15a36c77cfc7318588330f1138207).

Do others also get this error? Is this something wrong in the agent or a race condition with TPMs more generally?

lkatalin:
This change adds a check to make sure digests returned in the Quote match the PCR values retrieved from `pcr_read`. There can be a mismatch here due to a race condition inherent to TPM2.0. See section `17.6.2` in TCG's [TPM spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part1_Architecture_pub.pdf) for more on this race condition.

There were also some ESAPI bugs related to the ordering of data in `PcrSelectionList` and `PcrData`, affecting our results. Therefore the code is temporarily pinned to a branch of the ESAPI until @puiterwijk 's ESAPI [bug fixes](https://github.com/parallaxsecond/rust-tss-esapi/pull/236) are merged.

Fixes #210

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,8 @@ serde_json = { version = "1.0", features = ["raw_value"] }
 tempfile = "3.0.4"
 tokio = {version = "0.2", features = ["full"]}
 tokio-io = "0.1"
-tss-esapi = "5.0"
+#tss-esapi = "5.0"
+tss-esapi = { git = "https://github.com/puiterwijk/rust-tss-esapi.git", branch = "keylime" }
 thiserror = "1.0"
 zmq = "0.9.2"
 uuid = {version = "0.8", features = ["v4"]}
--- a/src/error.rs
+++ b/src/error.rs
@@ -22,6 +22,8 @@ pub(crate) enum Error {
     InvalidRequest,
     #[error("Configuration loading error: {0}")]
     Ini(#[from] ini::ini::Error),
+    #[error("Infallible: {0}")]
+    Infallible(#[from] std::convert::Infallible),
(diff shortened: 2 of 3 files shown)


puiterwijk:
I'm unassigning myself, since I've co-authored this patch.

ashcrow (commented):
It does feel odd to check multiple times in hopes of getting one pass but it seems that may be the best workaround. One request though is to add to the commit message the following so it's in the git history:

- Note that the use of tss-esapi fork is temporary and that it's to get around an upstream bug
- Give a high level explanation of the race condition bug

lkatalin:
> It does feel odd to check multiple times in hopes of getting one pass but it seems that may be the best workaround. One request though is to add to the commit message the following so it's in the git history:
> 
>     * Note that the use of tss-esapi fork is temporary and that it's to get around an upstream bug
> 
>     * Give a high level explanation of the race condition bug

Great idea, @ashcrow , I updated the commit message.

It is a bit of an awkward workaround but apparently is the state of the art currently when working with TPMs (it's mentioned in the TPM spec as one of two ways to avoid the race condition).

ashcrow (approved):

PR #238, opened 2021-07-15
Accept sets of U and V keys; use new Key types

### LINKED ISSUE #209: Add functionality to receive multiple U and V keys
In Python, there are sets of U and V keys that can be stored and combining each pair is attempted to find the symmetric payload decryption key. Add this functionality (Rust model currently assumes only one U and one V key).

[Reference](https://github.com/keylime/keylime/blob/master/keylime/keylime_agent.py#L400):
```
def attempt_decryption(self):
        """On reception of a U or V value, this method is called to attempt the decryption of the Cloud Init 
         script
        At least one U and V value must be received in order to attempt encryption. Multiple U and V 
        values are stored
        to prevent an attacker from sending U/V values to deny service.
        """
        with uvLock:
            both_u_and_v_present = False
            return_value = False
            for u in self.u_set:
                for v in self.v_set:
                    both_u_and_v_present = True
                    return_value = self.decrypt_check(u, v)
                    if return_value:
                        # reset u and v sets
                        self.u_set = set([])
                        self.v_set = set([])
                        return return_value
            # TODO check on whether this happens or not.  NVRAM causes trouble
            if both_u_and_v_present:
                pass
                # logger.critical("Possible attack from: " + str(handler.client_address) + ".  
                Both U (potentially stale from TPM NVRAM) and V present but unsuccessful in attempt 
                to decrypt check value.")
            return return_value
```

### LINKED ISSUE #237: Consider a Key type
Keys are currently `[u8; 32]` with the key length as a constant. It would be more readable to just have this as a type.

lkatalin:
The key types seem like a good idea, especially if we start using vectors of U and V keys, which become cumbersome to read everywhere. I'm not completely convinced about making the U and V keys into sets as it adds a bit of complexity and seems like not the best way to mitigate attacks, but this would at least get the Rust agent to do what the Python agent is doing. Fixes #209 fixes #237

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -38,6 +38,54 @@ pub const AUTH_TAG_LEN: usize = 96;
 pub const KEY_LEN: usize = 32;
 pub const AES_BLOCK_SIZE: usize = 16;
 
+// symmetric keys as bytes
+pub type KeyBytes = [u8; KEY_LEN];
+
+// a vector holding keys
+#[derive(Debug, Clone)]
+pub struct KeySet {
+    pub set: Vec<SymmKey>,
+}
+
+impl Default for KeySet {
+    fn default() -> Self {
+        let set = Vec::new();
+        KeySet { set }
+    }
+}
+
+impl KeySet {
+    pub fn all_empty(&self) -> bool {
+        self.set.iter().all(|&key| key.is_empty())
+    }
+}
+
+// a key of len KEY_LEN
+#[derive(Debug, Clone, Copy)]
+pub struct SymmKey {
+    pub bytes: KeyBytes,
+}
+
+impl Default for SymmKey {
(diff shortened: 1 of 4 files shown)


puiterwijk (approved):
It's really annoying that this is necessary, but this seems fine.

ashcrow (approved):
I agree with @puiterwijk. Though a secondary effect is the use of the key(s) is a bit easier to read even though the implementation is more complex.

lkatalin:
Yes, and there are still loopholes, for example if an attacker sends a flawed key before Keylime sends two correct keys, there will still be a failure. But it will provide a little protection until some larger design issues are sorted out across the Keylime components.

@ashcrow Glad it's easier to read! I should have made a Key type to begin with.

PR #242, opened 2021-07-19
Add support for contact ip and port for the verifier and tenant to use

THS-on:
This adds support for enhancement proposal https://github.com/keylime/enhancements/issues/50.
I wanted to try the rust but I needed that feature to make it work with my other systems.

Note: This is my first time implementing something in rust so somethings might be implemented incorrectly.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -137,6 +137,22 @@ pub(crate) fn registrar_port_get() -> Result<String> {
     config_get_env("registrar", "registrar_port", "REGISTRAR_PORT")
 }
 
+/// Returns the contact ip for the agent if set
+pub(crate) fn cloudagent_contact_ip_get() -> Option<String> {
+    match config_get_env("cloud_agent", "agent_contact_ip", "KEYLIME_AGENT_CONTACT_IP") {
+        Ok(ip) => Some(ip),
+        Err(_) => None // Ignore errors because this option might not be set
+    }
+}
+
+/// Returns the contact ip for the agent if set
+pub(crate) fn cloudagent_contact_port_get() -> Option<u32> {
+    match config_get_env("cloud_agent", "agent_contact_port", "KEYLIME_AGENT_CONTACT_PORT") {
+        Ok(port) => Some(port.parse::<u32>().unwrap()),
+        _ => None // Ignore errors because this option might not be set
+    }
+}
+
 /*
  * Input: [section] and key
  * Return: Returns the matched key
--- a/src/common.rs
+++ b/src/common.rs
@@ -109,92 +109,32 @@ pub(crate) fn config_file_get() -> String {
 
 /// Returns revocation ip from keylime.conf if env var not present
 pub(crate) fn revocation_ip_get() -> Result<String> {
-    match env::var("REVOCATION_IP") {
-        Ok(ip) => {
-            // The variable length must be larger than 0 to accept
-            if !ip.is_empty() {
-                Ok(ip)
-            } else {
-                config_get("general", "receive_revocation_ip")
-            }
-        }
-        _ => config_get("general", "receive_revocation_ip"),
-    }
+    config_get_env("general", "receive_revocation_ip", "REVOCATION_IP")
 }
 
 /// Returns revocation port from keylime.conf if env var not present
 pub(crate) fn revocation_port_get() -> Result<String> {
-    match env::var("REVOCATION_PORT") {
-        Ok(ip) => {
-            // The variable length must be larger than 0 to accept
-            if !ip.is_empty() {
-                Ok(ip)
-            } else {
-                config_get("general", "receive_revocation_port")
-            }
-        }
-        _ => config_get("general", "receive_revocation_port"),
-    }
+    config_get_env("general", "receive_revocation_port", "REVOCATION_PORT")
 }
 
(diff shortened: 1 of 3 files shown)


src/common.rs
```diff
@@ -109,91 +109,47 @@ pub(crate) fn config_file_get() -> String {
...
-            // The variable length must be larger than 0 to accept
-            if !ip.is_empty() {
-                Ok(ip)
-            } else {
-                config_get("registrar", "registrar_ip")
-            }
-        }
-        _ => config_get("registrar", "registrar_ip"),
-    }
+    config_get_env("registrar", "registrar_ip", "REGISTRAR_IP")
```
lukehinds, line 132: this needs to be cloud_agent:

`config_get_env("cloud_agent", "registrar_ip", "REGISTRAR_IP")`

src/common.rs
```diff
@@ -109,91 +109,47 @@ pub(crate) fn config_file_get() -> String {
...
-        }
-        _ => config_get("registrar", "registrar_port"),
+    config_get_env("registrar", "registrar_port", "REGISTRAR_PORT")
+}
+
+/// Returns the contact ip for the agent if set
+pub(crate) fn cloudagent_contact_ip_get() -> Option<String> {
+    match config_get_env("cloud_agent", "agent_contact_ip", "KEYLIME_AGENT_CONTACT_IP") {
+        Ok(ip) => Some(ip),
+        Err(_) => None // Ignore errors because this option might not be set
```
lukehinds, line 144: Are you sure there could be a legitimate error here?

lukehinds (changes requested):
I have not been able to test the agent_contact stuff yet, it looks like something in python keylime has broken the quote operation here (nothing to do with this change), I will try to figure out what has happened.

src/common.rs
THS-on: The key might not be in `keylime.conf` and not specified as environment variable which is ok because this setting is optional. We currently have no distinction between error and key not found, so I didn't found a better way to handle that.

src/common.rs
THS-on: Yes, this should be `cloud_agent` but I've just kept the old behaviour. 
If I change that the `keylime.conf` which currently doesn't contain that key and `registrar_port_get(...)` should be also updated.

THS-on:
>  Any tip offs of how I should set up using the agent_contact feature?

@lukehinds 
* Run your agent on another IP and port than `127.0.0.1` and `9002` for example `192.168.0.1` and `9003`.
* Then set the either the keys `agent_contact_ip` and `agent_contact_port` or the respective environment variables to those new values.
* Add agent to the verifier using `keylime_tenant`, but only specify the uuid not the IP or port.
* The agent should be successfully added to the verifier with IP and port specified with `agent_contact_ip`/`agent_contact_port`

src/common.rs
THS-on: @lukehinds should I include the update of the `keylime.conf` in this PR or should I move those changes into a new PR?

lukehinds: good question, at some point we need to think about how we insure parity between the config here and the config over n keylime/keylime.

I am fine with it being in this PR.

src/common.rs
lukehinds: ack, thanks!

src/common.rs
```diff
@@ -109,91 +109,55 @@ pub(crate) fn config_file_get() -> String {
...
+}
+
+/// Returns the contact ip for the agent if set
+pub(crate) fn cloudagent_contact_port_get() -> Option<u32> {
+    match config_get_env(
+        "cloud_agent",
+        "agent_contact_port",
+        "KEYLIME_AGENT_CONTACT_PORT",
+    ) {
+        Ok(port) => Some(port.parse::<u32>().unwrap()),
```
ueno, line 159: Afaik `.unwrap()` is not allowed in this project unless it's explicitly marked with `//#[allow_ci]`, but in this case `str::parse<u32>` actually could fail if `port` is not parsable. Maybe good to change the return type of this function to `Result<Option<u32>>`?

lkatalin: I agree with @ueno .

lkatalin (commented):
@THS-on This is a great improvement and makes the code cleaner as well - thanks! 
When `keylime.conf` is updated and the return type of `cloudagent_contact_port_get()` is a `Result` I think it will be good to go.

src/common.rs
THS-on: Fixed

src/common.rs
THS-on: Done. I've now updated the keylime.conf and moved the keys to the correct section.

lkatalin:
@THS-on The `keylime.conf` in the Python repo is much longer; why is this? Maybe I misread but I thought we were going to replicate it exactly.

THS-on:
@lkatalin we only the the keys that are for the agent. We don't need to configuration for the registrar, tenant, verifier, webapp, ca and the Python specific logging setup.

lkatalin:
> @lkatalin we only need the keys that are for the agent. We don't need to configuration for the registrar, tenant, verifier, webapp, ca and the Python specific logging setup.

Gotcha, well that's reasonable.

lkatalin (approved):

ueno (approved):
Looks good to me as well.

PR #243, opened 2021-07-21
Add "v1.0" prefix to agent APIs

mpeters:
This is in preparation for the versioned-apis feature to land in keylime main.


mpeters:
First time with rust... I know this is as simple as it gets but wanted to make sure it's the right approach.

lukehinds (approved):
GTM

PR #244, opened 2021-07-28
Generate Cargo.lock file

aplanas:
The Cargo FAQ explain that to determine deterministic builds of binary
files, is recommended to commit the Cargo.lock files in binary projects.

This becomes relevant when building a RPM package for distribution, when
all the depenencies are generated via "cargo vendor".  The build will
fail if "Cargo.lock" is not present.

This commit removes the Cargo.lock file from .gitignore and generate a
Cargo.lock file using the Rust 1.53.0 release.


lukehinds:
Sorry for the delay on this @aplanas , could you link to the relevant Cargo FAQ, that would be enough for me to approve this and get it merged.

aplanas:
Oh sure!

https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

lukehinds (approved):

Issue #239, opened 2021-07-15
Remove temporary dependency on bug fix branch for ESAPI


lkatalin:
Once [this](https://github.com/parallaxsecond/rust-tss-esapi/pull/238) merges (there are a few other ESAPI fixes coming in too).

PR #245, opened 2021-07-28
build: Adjust package dependencies to compile in Fedora

ueno:
This makes it possible to build this crate with the Rust toolchain in
Fedora, so the keylime_agent binary can be packaged as an RPM.

The commit comprises the following:
- Use tss-esapi 6.1.1 from crates.io
- Bump rust-ini dependency to 0.17
- Bump pretty_env_logger dependency to 0.4
- Bump hex dependency to 0.4
- Move wiremock to dev-dependencies

Fixes: #239

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -231,7 +231,7 @@ pub(crate) fn config_get(section: &str, key: &str) -> Result<String> {
         }
     };
 
-    Ok(value.clone())
+    Ok(value.to_string())
 }
 
 /*
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,7 +81,7 @@ dependencies = [
  "serde_urlencoded",
  "sha-1",
  "slab",
- "time 0.2.27",
+ "time",
 ]
 
 [[package]]
@@ -247,7 +247,7 @@ dependencies = [
  "serde_json",
  "serde_urlencoded",
  "socket2 0.3.19",
- "time 0.2.27",
+ "time",
  "tinyvec",
  "url",
 ]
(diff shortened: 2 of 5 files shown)


ueno:
Preliminary packages are available at https://copr.fedorainfracloud.org/coprs/ueno/rust-keylime/

src/common.rs
```diff
@@ -230,7 +230,7 @@ pub(crate) fn config_get(section: &str, key: &str) -> Result<String> {
         }
     };
 
-    Ok(value.clone())
+    Ok(value.to_string())
```
lukehinds, line 233: looks like this might be a separate matter? (or is this related to the dependency bump?)

ueno: This is needed for rust-ini bump from 1.12.1 to 1.17.0. In 1.12.1, `value` is supposed to be a `String`, because `ini::ini::Properties` is an alias to [`HashMap<String,String>`](https://docs.rs/rust-ini/0.12.1/ini/ini/type.Properties.html), while in 1.17.0, it is redefined as [a dedicated struct](https://docs.rs/rust-ini/0.17.0/ini/struct.Properties.html#method.get), whose `get` method returns `&str`; so a manual conversion to `String` is necessary.

lukehinds: makes sense, thanks!

ueno:
I've also updated the `tss-esapi` dependency  to 6.0 as it includes @puiterwijk's patch.

lukehinds:
LGTM @ueno , if you could rebase this I am happy to approve.

ueno:
Thanks; rebased and removed the `tss-esapi` update (as it's already part of https://github.com/keylime/rust-keylime/commit/31ac452fb42dbb0654903eb1a72c17c8c6173200).

lukehinds (approved):

ueno:
@lkatalin @lukehinds now that tss-esapi 6.1.1 has been published on crates.io, I've incorporated the changes needed for using that version (plus removal of `unsafe` block which is no longer necessary, and `Tcti` -> `TctiNameConf` [transition](https://github.com/parallaxsecond/rust-tss-esapi/blob/6.x.y/tss-esapi/src/lib.rs#L102)).

lkatalin:
@ueno This looks good to me, awesome work, thank you!
I guess the changes I pinned to earlier from the [tss-esapi github](https://github.com/parallaxsecond/rust-tss-esapi) (ex. `SignatureSchemeAlgorithm` instead of `SignatureScheme`) are not published yet in `6.1.1` so we have to revert them for now.

lkatalin (approved):

ueno:
Thanks for the review!

PR #246, opened 2021-09-13
Add keylime_ima_emulator executable for testing

ueno:
To run with a TPM emulator, IMA needs to be simulated as PCRs are not
automatically updated by the kernel.  For this purpose the
keylime_ima_emulator command is provided in the Python implementation.
This patch ports the command to enable standalone testing without
installing the Python Keylime stack.


src/ima_emulator.rs
```diff
@@ -0,0 +1,163 @@
...
+        return Err(ImaEmulatorError::Other(
+            "Unable to find current measurement list position, Resetting the TPM emulator may be neccesary".to_string()));
+    }
+
+    Ok(position)
+}
+
+fn main() -> std::result::Result<(), ImaEmulatorError> {
+    let tcti = Tcti::from_environment_variable()?;
+    let mut context = unsafe { Context::new(tcti) }?;
```
lukehinds, line 109: we are generally trying to avoid unsafe (the compiler is currently complaining about this), is there not another approach we could leverage here?

lukehinds:
tried this out @ueno , but cannot get further than:

```
target/debug/keylime_ima_emulator
Error: TssEsapiError(WrapperError(ParamsMissing))
```

src/ima_emulator.rs
ueno: This will be unnecessary, once `tss-esapi` is upgraded to 6.x.

ueno:
@lukehinds the program requires a TCTI envvar (either `TPM2TOOLS_TCTI`, `TCTI`, or `TEST_TCTI`) set, because it is supposed to be used with a software TPM implementation (it should print a better error message, yes).

lukehinds (approved):
OK, looks good to me.

PR #248, opened 2021-09-14
keys_handler: verify HMAC in constant-time

ueno:
It is generally recommended that HMAC verification should be done
using a constant-time primitive so it does not involve timing side
channel.  This uses openssl::memcmp as suggested in the crate
documentation:
https://docs.rs/openssl/0.10.36/openssl/sign/index.html#examples


lkatalin (approved):
Thank you for this improvement and the documented reference, @ueno !

ueno:
Thanks for the review. I guess we should cover this part with unit tests, but at least confirmed that it doesn't break the existing behavior (by seeing "HMAC check passed").

Issue #247, opened 2021-09-14
`main` exits in non-fatal occasions


ueno:
I'm trying to test the Rust agent on Fedora CoreOS VM (see the current [instruction](https://github.com/ueno/rust-keylime-fcos34)), and realized that the agent stops running in certain conditions:
- when `payload_script` is non empty, but the path doesn't exist (Python agent just prints a [warning](https://github.com/keylime/keylime/blob/master/keylime/keylime_agent.py#L315))
- any error on executing the payload script (including decryption failure) is treated as fatal and causes the agent to exit
- when `listen_notfications` is `False` (Python agent still enters a [loop](https://github.com/keylime/keylime/blob/master/keylime/keylime_agent.py#L655))

I would suggest reorganizing the mainloop logic as follows:
- do payload script decryption/execution in actix server handler
- always join the running threads at the end of `main`

PR #249, opened 2021-09-22
Retain the main loop running after payload execution

ueno:
Fixes: #247

Still a draft as I haven't checked every error case that we can ignore.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -261,6 +261,52 @@ pub(crate) fn optional_unzip_payload(unzipped: &str) -> Result<()> {
     Ok(())
 }
 
+async fn run_encrypted_payload(
+    symm_key: Arc<Mutex<SymmKey>>,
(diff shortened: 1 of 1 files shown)


lkatalin:
This is much better. Thanks for making this change.

lkatalin:
@ueno Would you be open to merging this version? I think it may help with #262 (other fixes are also needed, which I am working on).

ueno:
@lkatalin sure, just rebased against master.

lkatalin (approved):

lkatalin:
> @lkatalin sure, just rebased against master.

@ueno Thanks! GHA was down for a bit and it looks like the tests got stuck; I tried to re-trigger it from this end but it hasn't helped, so if you get a chance do you mind re-pushing to see if that fixes it?

ueno:
Thanks for the review.

Issue #250, opened 2021-09-22
Promote Daiki Ueno to maintainer


lukehinds:
Daiki Ueno has made some great contributions as of late and has a lot of experience maintaining low level crypto libraries. I therefore think it makes sense to promote Daiki to a maintainer in rust keylime.

Maintainers please respond with +1's or feedback.

https://github.com/ueno

lukehinds:
Cc @lkatalin @ashcrow @mpeters @puiterwijk

lkatalin:
+1 I agree!

PR #252, opened 2021-09-27
Add Ueno as a maintainer and set codeowners

lukehinds:
codeowners also set, which allows for better branch protections

Resolves: #250

Signed-off-by: Luke Hinds <lhinds@redhat.com>

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+@keylime/rust-team
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -4,3 +4,4 @@
 @lkatalin
 @puiterwijk
 @mpeters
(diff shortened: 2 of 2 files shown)


mpeters (approved):
lgtm

lkatalin (approved):

lkatalin:
This looks like one of those cases where an unrelated clippy error is causing the CI to fail. I can put in a fix for it pretty quick, I think.

lukehinds:
> This looks like one of those cases where an unrelated clippy error is causing the CI to fail. I can put in a fix for it pretty quick, I think.

got for it

lkatalin:
@lukehinds You're good to rebase now and it should fix CI.

PR #253, opened 2021-09-27
Fix clippy errors, update to newest TSS-ESAPI

### LINKED ISSUE #239: Remove temporary dependency on bug fix branch for ESAPI

lkatalin:

Fixes #239

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ serde_json = { version = "1.0", features = ["raw_value"] }
 tempfile = "3.0.4"
 tokio = {version = "0.2", features = ["full"]}
 tokio-io = "0.1"
-#tss-esapi = "5.0"
-tss-esapi = { git = "https://github.com/puiterwijk/rust-tss-esapi.git", branch = "keylime" }
+#tss-esapi = "6.1"
+tss-esapi = { git = "https://github.com/parallaxsecond/rust-tss-esapi.git", branch = "main" }
 thiserror = "1.0"
 zmq = "0.9.2"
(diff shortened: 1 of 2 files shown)


lukehinds (approved):
feel free to merge when test passes,

Cargo.toml
```diff
@@ -28,8 +28,8 @@ serde_json = { version = "1.0", features = ["raw_value"] }
 tempfile = "3.0.4"
 tokio = {version = "0.2", features = ["full"]}
 tokio-io = "0.1"
-#tss-esapi = "5.0"
-tss-esapi = { git = "https://github.com/puiterwijk/rust-tss-esapi.git", branch = "keylime" }
+#tss-esapi = "6.1"
+tss-esapi = { git = "https://github.com/parallaxsecond/rust-tss-esapi.git", branch = "main" }
```
ueno, line 32: Can't we use the `tss-esapi` crate from crates.io (it would make Fedora packaging a little easier)?
I see the patches in the keylime branch had been merged in the 6.0 release.

lukehinds: This makes sense to me. I can see they have a 6.1 in crates, maybe we should update to that at the same time.

lkatalin: @ueno @lukehinds I agree - I can double check, but the released version was still giving me the clippy error yesterday. It is possible the clippy update came out after the 6.x release and so it is not yet fixed there. But I can look again in case there was something wrong on my end. We do want to use the crates.io version asap.

Issue #267, opened 2021-10-07
AES-GCM key length


ueno:
While there is a [special case](https://github.com/keylime/rust-keylime/blob/master/src/main.rs#L140) to handle different key lengths for AES-GCM, `/keys/ukey` (and `/keys/vkey`) actually returns error (at [`SymmKey::from_vec`](https://github.com/keylime/rust-keylime/blob/master/src/common.rs#L87)) if the key length does not match `KEY_LEN` (= 32), so only AES-256-GCM is allowed.

Should it support AES-128-GCM as well, or could the special case be removed if it's not used in practice on the protocol?

Issue #94, opened 2020-09-24
Integration Tests


lukehinds:
Alongside unit tests, we need some form of functional / Integration tests, where we test core functionality against the Keylime registrar and verifier.

lkatalin:
Here are four commands that are a good baseline for testing:

### “base” node identity
`keylime_tenant -c update -t 127.0.0.1 -v 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --no-verifier-check --cert /var/lib/keylime/ca/d432fbb3-d2f1-4a97-9ef7-75bd81c00000`

### “base” + measured boot
`keylime_tenant -c update -t 127.0.0.1 -v 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --no-verifier-check --cert /var/lib/keylime/ca/d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --mb_refstate /root/mb_refstate.txt`
where `mb_refstate.txt` can contain just `{}`

### “base” + measured boot + IMA
`keylime_tenant -c update -t 127.0.0.1 -v 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --no-verifier-check --cert /var/lib/keylime/ca/d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --mb_refstate /root/mb_refstate.txt --allowlist /root/allowlist.txt --exclude /root/excludes.txt`

### “base” + measure boot + IMA + autorun
`keylime_tenant -c update -t 127.0.0.1 -v 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --no-verifier-check --cert /var/lib/keylime/ca/d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --mb_refstate /root/mb_refstate.txt --allowlist /root/allowlist.txt --exclude /root/excludes.txt --include /root/senddir`
where `senddir` is a (non-zipped) directory containing, for example, some script `autorun.sh`

THS-on:
> Which of the custom tests do we want to run as part of the Rust agent CI most urgently?

1. code style and compile tests
2. Unit tests
3. SP integration tests to see that nothing got broken
4. Basic automatic security analysis (e.g. cargo audit, are there tools like CodeQL for Rust?)

>  What do we want to see from integration testing in the near future?

[...]

PR #268, opened 2021-10-07
Some refactoring on keys and crypto usage

ueno:
This adds a few changes to refactor keys and crypto usage:
- Use `actix_web::web::Json` for type safe payload parsing
- Move crypto operations to the `crypto` module for easier error handling
- Use `Option` to represent the combined key
- Use `Condvar` to wait for the combined key
- Support 128-bit key AES for payload encryption

Fixes: #267


src/crypto.rs
```diff
@@ -168,6 +169,70 @@ pub(crate) fn asym_verify(
...
+    // Decrypt and truncate the buffer
+    let decrypted_len = decrypter.decrypt(data, &mut decrypted)?;
+    decrypted.truncate(decrypted_len);
+
+    Ok(decrypted)
+}
+
+pub(crate) fn compute_hmac(key: &[u8], data: &[u8]) -> Result<Vec<u8>> {
+    let pkey = PKey::hmac(key)?;
+    let mut signer = Signer::new(MessageDigest::sha384(), &pkey)?;
```
lkatalin, line 195: I should have done this in the original version, but maybe it would be good to document why we are using these algos (they are expected on the Python side) - with refs to the Python files / line numbers. I can also do this.

lkatalin (commented):
These look like good changes so far. 

I had combined the U and V handlers into one to remove some of the code duplication, but in this version we don't have to check for which key it is and that also helps make it more concise and readable. I am not sure which version ends up better, so I am okay with this change.

ueno:
@lkatalin could you take a look when you have time? I think this is almost ready for review.

src/crypto.rs
```diff
@@ -166,9 +117,103 @@ pub(crate) fn asym_verify(
...
+ * Output: signed HMAC result
+ *
+ * Sign message and return HMAC result string
+ */
+pub(crate) fn compute_hmac(key: &[u8], data: &[u8]) -> Result<Vec<u8>> {
+    let pkey = PKey::hmac(key)?;
+    // SHA-384 is used as the underlying hash algorithm.
+    //
+    // Reference:
+    // https://keylime-docs.readthedocs.io/en/latest/rest_apis.html#post--v1.0-keys-ukey
```
lkatalin, line 163: Awesome, thanks! I didn't even know about these docs.

ueno:
@lkatalin could you take a look again?

src/keys_handler.rs
```diff
@@ -330,11 +274,42 @@ mod tests {
...
+    #[cfg(feature = "testing")]
+    #[actix_rt::test]
+    async fn test_u_or_v_key_short() {
+        test_u_or_v_key(AES_128_KEY_LEN, None).await;
+    }
+
+    #[cfg(feature = "testing")]
+    #[actix_rt::test]
+    async fn test_u_or_v_key_long() {
+        test_u_or_v_key(AES_128_KEY_LEN, None).await;
```
lkatalin, line 298: Should this one be `AES_256_KEY_LEN`?

src/keys_handler.rs
```diff
@@ -330,11 +274,42 @@ mod tests {
             .set_json(&vkey)
             .to_request();
 
-        let resp = test::call_service(&mut app, req).await;
+        let resp = test::call_service(&mut app, req).await; //#[allow_ci]
```
lkatalin, line 277: Nit: looks like we don't need `//#[allow_ci]` here.

lkatalin (commented):
Thanks @ueno , this LGTM in terms of the code. 

I'm having a problem, though, where I test this specific branch with a tenant `add` command and it does not work for me. I get:

```
2021-12-15 21:00:05.097 - keylime.tenant - INFO - Quote from 127.0.0.1 validated
2021-12-15 21:00:05.107 - keylime.tenant - ERROR - Response code 200: Success
/usr/lib/python3.9/site-packages/urllib3/connectionpool.py:981: InsecureRequestWarning: Unverified HTTPS request is being made to host '127.0.0.1'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  warnings.warn(
2021-12-15 21:00:05.841 - keylime.tenant - INFO - Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 stopped
  File "/usr/local/bin/keylime_tenant", line 33, in <module>
    sys.exit(load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_tenant')())
  File "/usr/local/lib/python3.9/site-packages/keylime/cmd/tenant.py", line 15, in main
    tenant.main()
  File "/usr/local/lib/python3.9/site-packages/keylime/tenant.py", line 1359, in main
    mytenant.do_quote()
  File "/usr/local/lib/python3.9/site-packages/keylime/tenant.py", line 1057, in do_quote
    raise e
  File "/usr/local/lib/python3.9/site-packages/keylime/tenant.py", line 1053, in do_quote
    raise UserError(
keylime.tenant.UserError: Posting of Encrypted U to the Cloud Agent failed with response code 400 ()
```

You can see the quote is validated, but then something goes wrong. I double checked this is not happening for me with the current master branch. However... it is still possible that this is due to something in my environment, as I sometimes have some glitches.

Would you be able to double check whether you can reproduce this error or not? The command I'm using for the tenant is 
```
keylime_tenant -v 127.0.0.1 -t 127.0.0.1 --uuid d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --allowlist /root/allowlist.txt --include /root/senddir --cert /root/ca --exclude /root/excludes.txt -c add
```

This may not have to do with these changes, but I would like to be sure before merging.

lkatalin:
@ueno Going commit by commit, I don't get the `400` error on the first commit (Remove unused imports `f94290bb83abda9ae60af78ca0dbd1afdd9de64b`), but I do start getting it with the second commit (keys_handler: Make use of type safe payload deserialization `ffc6316b3e7094e4c9d0215ce8e622e25c21514f`).

lkatalin:
This also points to a need for #94

lkatalin:
If I were to hazard a guess, I'd say it has to do with using `key: web::Json<KeylimeUKey>,` in `u_key()`, which it can't interpret for some reason.

ueno:
> If I were to hazard a guess, I'd say it has to do with using `key: web::Json<KeylimeUKey>,` in `u_key()`, which it can't interpret for some reason.

Thanks for looking into it; yes, that seems like the case. If I run the agent with `RUST_LOG=actix_web=debug`, I see:
```console
 DEBUG actix_web::types::json > Failed to deserialize Json from payload. Request path: /v1.0/keys/ukey
```
Let me try to fix it.

ueno:
The reason seems to be that the Content-Type: header is missing in the HTTP request, while actix-web v3 [requires](https://github.com/actix/actix-web/blob/web-v3.3.3/src/types/json.rs#L339) one. So the only way forward for now would be to do the deserialization manually. I'll also open an issue in Python keylime to add Content-Type header for the requests.

THS-on:
The content type should be set in the Python project: https://github.com/keylime/keylime/blob/883085d6a4bcea3012729014d5b8e15ecd65fc7c/keylime/web_util.py#L138

Can you check if the header is not set?

ueno:
@THS-on here is the packet [capture](https://github.com/keylime/rust-keylime/files/7911237/ukey.txt.gz) taken with wireshark.

THS-on:
Yeah, what I thought was that the verifier or registrar forgot to set the headers, but in this case it is the tenant.

Can you check if this fixes it for you?  https://github.com/keylime/keylime/pull/845

THS-on:
> So the only way forward for now would be to do the deserialization manually.

I would wait for the fixes for adding the correct content type are merged and not trying to work around the current bug.

ueno:
> I would wait for the fixes for adding the correct content type are merged and not trying to work around the current bug.

I agree in general, but as the workaround needs only a couple of [lines](https://github.com/keylime/rust-keylime/pull/268/files#diff-3721aaaab998a2460958b4fa67e0305dbb3d16fba8816e2544da8ce0d0dc2e4eR82), I'd rather document it for now and create a "debt" issue to track it.

lkatalin:
Can confirm this new version works now with local testing. Thanks @ueno and @THS-on !

@ueno Do you still have a workaround in the code that can be removed? Or should I merge after CI passes?

ueno:
> @ueno Do you still have a workaround in the code that can be removed? Or should I merge after CI passes?

@lkatalin I think it's ready to merge. The workaround is marked with a comment so we can remove it after the new (python) keylime) release. I'll create a github issue for that.

lkatalin:
> > @ueno Do you still have a workaround in the code that can be removed? Or should I merge after CI passes?
> 
> @lkatalin I think it's ready to merge. The workaround is marked with a comment so we can remove it after the new (python) keylime release. I'll create a github issue for that.

Awesome, thank you!

lkatalin (approved):

PR #273, opened 2021-10-15
keys_handler: Add test with encrypted payload

ueno:
This adds a test to actually decrypt and run encrypted payload through the actix handlers.
The first 7 commits are from #268, so please ignore.


ueno:
Let's do this as part of #268, as this doesn't change any functional behavior.

ueno:
@lkatalin yes, I included that change in #268.

PR #279, opened 2021-11-16
Run Fedora tests on unified Keylime test container

mbestavros:
Previously, `rust-keylime` built and ran tests on a container in `tests/`. With https://github.com/keylime/keylime/pull/788, there's now a test container on Quay that should work for `rust-keylime` as well. This PR runs the existing test suite on that new container.

Signed-off-by: Mark Bestavros <mbestavr@redhat.com>


mpeters (approved):
lgtm

PR #282, opened 2021-12-10
Code refactor to use updated tss-esapi

### LINKED ISSUE #234: Multiple PCR reads necessary when many PCRs requested
According to @puiterwijk : if more than 8 PCR datas are requested (including e.g. 7 sha256 and 2 sha1 values), a single pcr_read won't suffice, and we'll need to get multiple reads, and combine the results

lkatalin:
Fixes #234 

At long last, we have a fix for the > 8 PCR read bug. This bug fix was made in the `rust-tss-esapi` and this PR brings our code up to date with theirs.

I think the goal will be using `rust-tss-esapi` realease `7.0.0` as our dependency because this will be the release that incorporates the PCR read bug fix. There are more changes coming in for that release so we will probably have to make further changes also. Until then the dependency is on the `rust-tss-esapi` repo so that we can keep up.

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -25,7 +25,13 @@ use openssl::{
 use flate2::{write::ZlibEncoder, Compression};
 
 use tss_esapi::{
-    abstraction::{ak, cipher::Cipher, ek, DefaultKey},
+    abstraction::{
+        ak,
+        cipher::Cipher,
+        ek,
+        pcr::{read_all, PcrData},
+        DefaultKey,
+    },
     attributes::session::SessionAttributesBuilder,
     constants::{
         session_type::SessionType,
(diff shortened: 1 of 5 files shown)


src/tpm.rs
```diff
@@ -613,10 +605,10 @@ fn hash_alg_to_message_digest(
 fn check_if_pcr_data_and_attestation_match(
     hash_algo: HashingAlgorithm,
     pcr_data: &PcrData,
-    attestation: &TPM2B_ATTEST,
+    attestation: Attest,
 ) -> Result<bool> {
-    let pcr_data = TPML_DIGEST::try_from(pcr_data.clone())?;
-    let attestation = unmarshal_tpms_attest(&attestation.attestationData)?;
+    let pcr_data = Vec::<TPML_DIGEST>::try_from(pcr_data.clone())?;
+    let attestation = TPMS_ATTEST::from(attestation);
```
ueno, line 611: I suspect this function could be refactored to use the public methods on the `Attest` structure, something like:
```rust
fn check_if_pcr_data_and_attestation_match(
    hash_algo: HashingAlgorithm,
    pcr_data: &PcrData,
    attestation: &Attest,
) -> Result<bool> {
    let pcr_data = Vec::<TPML_DIGEST>::try_from(pcr_data.clone())?;

    let quote = match attestation.attested() {
        AttestInfo::Quote { info } => info,
        _ => {
            return Err(KeylimeError::Other(format!(
                "Expected attestation type TPM2_ST_ATTEST_QUOTE, got {:?}",
                attestation.attestation_type()
            )));
        }
    };

    let attested_pcr = quote.pcr_digest().value();
```
Then you could omit a `clone` on Attest, marshalling to `TPMS_ATTEST`, and `unsafe` block around `attestation.attested.quote`.

ueno (approved):
Looks great to me!

src/tpm.rs
lkatalin: Thanks @ueno , this was a good suggestion! I've made the update.

lkatalin:
@lukehinds Should we go ahead and merge this?

PR #289, opened 2021-12-28
tpm: remove special handling for PCR10

THS-on:
IMA now works with all hash algorithms: https://github.com/keylime/keylime/pull/803

The agent seems to currently use always sha256, but this will be changed in a separate PR.


lkatalin (approved):
Thank you for making this change! I'm very excited that we don't need this idiosyncratic handling anymore. I tried this out in my local test setup and it validated the quote so I will go ahead and approve it.

PR #290, opened 2021-12-28
quotes_handler: send the TPM2 UEFI eventlog if PCR0 is in mask

THS-on:
The agent now also sends the UEFI eventlog for measured boot attestation.

I've also moved our serialization functions to a new module because it can be shared between quotes handler and the registration process.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -10,7 +10,7 @@ use std::convert::TryFrom;
 use std::env;
 use std::fs::File;
 use std::path::{Path, PathBuf};
-use tss_esapi::utils::TpmsContext;
+use tss_esapi::{structures::PcrSlot, utils::TpmsContext};
 
 /*
  * Constants and static variables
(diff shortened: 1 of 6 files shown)


lkatalin:
Thanks, this looks good! Out of curiosity I have the same question, is this already implemented on the Python side also?

THS-on:
> Out of curiosity I have the same question, is this already implemented on the Python side also?

Yes that mimics the Python implementation: https://github.com/keylime/keylime/blob/ef60e84ad05c9d551e7c5ff9be981151b43c9fd1/keylime/keylime_agent.py#L180-L186

lkatalin:
> > Out of curiosity I have the same question, is this already implemented on the Python side also?
> 
> Yes that mimics the Python implementation: https://github.com/keylime/keylime/blob/ef60e84ad05c9d551e7c5ff9be981151b43c9fd1/keylime/keylime_agent.py#L180-L186

Perfect. Thanks!

lkatalin (approved):

PR #286, opened 2021-12-17
cmd_exec: Fix PATH envvar construction

ueno:
When PATH is non-empty, TPM_TOOLS_PATH ("/usr/local/bin/") was
appended without a path separator, resulting in a corrupted PATH
value: "$PATH/usr/local/bin/".

Signed-off-by: Daiki Ueno <dueno@redhat.com>

Issue #220, opened 2021-07-07
Audit for extraneous code


lkatalin:
We want to make sure we haven't written any code that is unused.

PR #291, opened 2021-12-28
secure_mount: fix tmpfs creation for secure mount point

THS-on:
Using `cmd_exec::run` errors with:
```
 INFO  keylime_agent::quotes_handler  > GET integrity quote returning 200 response
 INFO  keylime_agent::keys_handler    > Received vkey
 INFO  keylime_agent::keys_handler    > HMAC check passed
 INFO  keylime_agent::keys_handler    > Successfully derived symmetric payload decryption key
 INFO  keylime_agent                  > Successfully decrypted payload
 INFO  keylime_agent::secure_mount    > secure storage location /tmp/secure not mounted.
 INFO  keylime_agent::secure_mount    > Directory "/tmp/secure" created.
 INFO  keylime_agent::secure_mount    > Mounting secure storage location /tmp/secure on tmpfs.
 INFO  keylime_agent::common          > Changed file /tmp/secure owner to root.
 INFO  keylime_agent::secure_mount    > Changed path /tmp/secure owner to root.
Error: Other("cannot get execution status code for Error type IO error: No such file or directory (os error 2)")
```
We now call mount directly using `Command` instead of the retry abstraction. `cmd_exec` is now unused and can be removed because we interact with the TPM directly using the tss bindings.

The `WORK_DIR` is changed to `/var/lib/keylime` to be consistent with the Python agent.


THS-on:
I see that #286 should also fix this issue, but I think that we don't need `cmd_exec` just for the calls to `mount`.

ueno:
Yes, and this looks much cleaner; maybe we could also remove `cmd_exec.rs`, given the agent doesn't rely on tpm2-tools?

THS-on:
@ueno Yeah this PR also removes `cmd_exec.rs`.

ueno (approved):

lkatalin (approved):
@THS-on This change was on my to-do list. Thank you so much.

lkatalin:
Looks good and related to #220

Issue #164, opened 2021-03-04
Load config only once


lkatalin:
From @ashcrow on #161 : I think we should eventually rework the configuration system so that we load the configuration once and then reload when we receive the proper signal. My reasoning is that the current set up means that if the file changes while the agent is running then the result could be old and new variables in use based on when they have been read.

lkatalin:
These are two separate issues: 1) only load it once in the code and pass reference to config obj around; 2) reload if conf file is updated.

lkatalin:
Closing in favor of #390

PR #292, opened 2021-12-29
Allow the hash, signing and encryption algorithm to be changed

THS-on:
The user can now specify the different algorithms in the `keylime.conf`.
Following configuration options are now supported:

* `tpm_hash_alg`
* `tpm_encryption_alg`
* `tpm_signing_alg`

The `hash.rs` module was renamed to `algorithms.rs` and implements all the necessary conversions to the different types (str, String, tss specific). I've also tried to move all `config_get(..)` calls that I encountered to `main.rs`, such that the configuration is only read on startup.

Note that this PR is on top of #289.


THS-on:
For the internal algorithm enums I've implemented only `TryFrom` for converting `&str` to the enum. 
I know there is also `FromStr`, but I don't know if we should implement both.

lkatalin (approved):
Thanks @THS-on , this is great and again necessary. I tested locally so LGTM.

lkatalin:
Also related to #164

Issue #235, opened 2021-07-14
Add capability to retain attestation artifacts


lkatalin:
- Includes the AIK-encrypted tpm quote and the retrieved boot/IMA logs
- Make an auditable record of attestation for each machine being verified
- Problem: AIKs change between agent runs

This is an enhancement being planned for Keylime as a whole and requires input from the community.

PR #293, opened 2021-12-30
Try to load AK from disk instead of always creating a new one

THS-on:
This PR is build on top of #292 because the new algorithms module is used.

The agent now creates a persistent file called `tpmdata.json` where it stores the context of the AK and the currently used algorithms. On startup the agent tries to load that file and to load the AK context. If that fails the agent just generates a new AK.
This has the major advantage that the agent can be restarted without causing always and attestation failure.

We probably also could do that with the EK, but I do not see much value in doing that.

Related #235


lkatalin:
@THS-on I'll have to give this one a bit of a closer look as I'm less familiar with this functionality. Is there a corresponding PR on the Python side for reference? My understanding from #235 was that this was a to-do in both Rust and Python, but I'm not sure.

THS-on:
>  Is there a corresponding PR on the Python side for reference? 

No because the Python agent already handles the EK and AK quite different compared to the rust agent. I think it does not make sense to rework the TPM handling in the Python agent unless strictly necessary. 

This is mostly a convince feature that the agent can reuse the AK between starts.

lkatalin:
> > Is there a corresponding PR on the Python side for reference?
> 
> No because the Python agent already handles the EK and AK quite different compared to the rust agent. I think it does not make sense to rework the TPM handling in the Python agent unless strictly necessary.
> 
> This is mostly a convince feature that the agent can reuse the AK between starts.

Okay. I think it LGTM but it could help to have a bit more high-level context, just for posterity, about why this is a safe change. Is it related to this [enhancement](https://github.com/keylime/enhancements/blob/master/1_persist_agent_restart.md)?

THS-on:
This is a save change because we don't have any lifetime guarantees on our AK, we just care about the attributes make it secure for signing quotes. An attacker could already implement this to reuse AKs.
The TPM store operation encrypts the private data with a (symmetric?) key before dumping it to disk, so that only the TPM can actually do something useful with that data. The load operation then decrypts the private data again inside the TPM.

Reboot detection should be implemented using the clock data in the quote and not by changing the AK every time.

> Is it related to this enhancement?

Yes it seems it is related to this enhancement.

In the future this will allow agents to register only with an AK (that is for example verified by another third party to be trusted)  and don't have to do make/activate credential every time during registration.

lkatalin:
Thanks @THS-on , that's good information to have. Especially that the data is encrypted.

lkatalin (approved):
This LGTM

PR #295, opened 2022-01-10
ima_emulator: Fix running hash calculation on resumption

ansasaki:
Previously, the emulator could not be restarted after stopped because
the running hash calculation was incorrect and would never match the PCR
10 content.  This fixes the extension calculation allowing the emulator
to be restarted without resetting the software TPM.


src/ima_emulator.rs
```diff
@@ -82,14 +83,14 @@ fn ml_extend(
             }
             Some(search_hash) => {
                 let mut hasher = openssl::sha::Sha1::new();
-                hasher.update(START_HASH);
+                hasher.update(&running_hash);
                 hasher.update(&template_hash);
-                let running_hash: Vec<u8> = hasher.finish().into();
-                let running_hash = Digest::try_from(running_hash)?;
+                running_hash = hasher.finish().into();
+                let digest = Digest::try_from(running_hash.clone())?;
```
ueno, line 89: Using `.as_slice` instead of `.clone` might avoid making an extra copy?

ueno (approved):
Thanks; it's a good finding and the change looks good to me!

src/ima_emulator.rs
ansasaki: Yes, I agree. I'll fix this.

Issue #233, opened 2021-07-14
Audit private / public functions


lkatalin:
Do all our public functions need to be public?

puiterwijk:
Short answer: nope 😄.
Longer answer: still nope. Though I think that we're compiling as a binary (non-library) crate anyway, so it doesn't really matter.

lkatalin:
Very low priority issue, I agree, but for the sake of dotting i's it would be nice to do it.

PR #296, opened 2022-01-13
move global configuration into one single struct

THS-on:
This moves the entire used configuration into a struct called `KeylimeConfig` and makes the all the `config_get` functions private.

The goal of this change is to load the configuration only once during startup and to keep most of the functions behaviour independent from global state. I don't think that partial config reload during execution makes sense, because in most cases a change would require a complete restart of the agent (generating new AK with different algorithm for example).

The config struct is currently passed by reference where possible. It might be required to use an Arc instead if we allow it to be shared between threads.

We should discuss before this gets merged, when functions take the config struct and when they are just taking single arguments for values that are in the config struct. My rough idea is the following:
 
- If it is an abstraction of an TPM interaction: use tss specific types where possible and let the caller do the conversion
- If it is in the main module: you can pass the config struct if it uses more than one value
- If it is not in the main module: pass only the data required as arguments not the complete struct  (the current `revocation` module would violate this rule)

Related #233, #220, #164

### DIFF
--- a/src/revocation.rs
+++ b/src/revocation.rs
@@ -4,7 +4,7 @@
 #[macro_use]
 use log::*;
 
-use crate::common::{config_get, REV_CERT};
+use crate::common::{KeylimeConfig, REV_CERT};
 use crate::crypto;
 use crate::error::*;
 use crate::secure_mount;
@@ -61,7 +61,10 @@ pub(crate) fn run_action(
 /// An OK result indicates all actions were run successfully.
 /// Otherwise, an Error will be returned from the first action that
 /// did not run successfully.
-pub(crate) fn run_revocation_actions(json: Value) -> Result<Vec<Output>> {
+pub(crate) fn run_revocation_actions(
+    json: Value,
+    secure_size: &str,
+) -> Result<Vec<Output>> {
     #[cfg(not(test))]
     pretty_env_logger::init();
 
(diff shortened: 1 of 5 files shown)


src/revocation.rs
```diff
@@ -242,8 +247,8 @@ mod tests {
         );
         let json_str = std::fs::read_to_string(json_file).unwrap(); //#[allow_ci]
         let json = serde_json::from_str(&json_str).unwrap(); //#[allow_ci]
-
-        let outputs = run_revocation_actions(json);
+        let secure_size = "1m";
```
lukehinds, line 250: Any reason for changing this to a hard coded value rather than grabbing it from config?

Note you also have the same on line # 273

lukehinds: just noted its a WIP, so ignore the above if N/A

THS-on: Not really. I should probably add a KeylimeConfig struct for the tests.

THS-on: I've now implemented the Dafault trait for the KeylimeConfig struct and removed the hard-coded values.

ashcrow (commented):
Looks good. Thanks for doing this!

lkatalin (approved):
This is a great idea. Thanks.

Issue #275, opened 2021-10-17
Mismatches in revocation notification signature verification


ueno:
While the Python agent uses RSA-PSS with SHA-256 (and the maximum salt length depending on the RSA key length):
https://github.com/keylime/keylime/blob/098d9c1d0d76b424ab7c3631fec04cf40df323f0/keylime/cloud_verifier_common.py#L395
https://github.com/keylime/keylime/blob/098d9c1d0d76b424ab7c3631fec04cf40df323f0/keylime/crypto.py#L71

the Rust agent assumes a raw RSA signature without padding:
https://github.com/keylime/rust-keylime/blob/8d5a468824dc609c21ffb0932213fe009116eeee/src/revocation.rs#L207
https://github.com/keylime/rust-keylime/blob/8d5a468824dc609c21ffb0932213fe009116eeee/src/crypto.rs#L166

Also the signature itself should be encoded in base64.

PR #297, opened 2022-01-17
crypto: Fix algorithms used for revocation signature

ansasaki:
The revocation signature uses RSA-PSS and is expected to be base64
encoded.

Fixes: #275


src/crypto.rs
```diff
@@ -164,8 +164,12 @@ pub(crate) fn asym_verify(
     signature: &str,
 ) -> Result<bool> {
     let mut verifier = Verifier::new(MessageDigest::sha256(), keypair)?;
+    verifier.set_rsa_padding(Padding::PKCS1_PSS)?;
+    verifier.set_rsa_mgf1_md(MessageDigest::sha256())?;
+    verifier
+        .set_rsa_pss_saltlen(openssl::sign::RsaPssSaltlen::MAXIMUM_LENGTH)?;
     verifier.update(message.as_bytes())?;
-    Ok(verifier.verify(signature.as_bytes())?)
+    Ok(verifier.verify(&base64::decode(signature.as_bytes()).unwrap())?)
```
ueno, line 172: I think calls that may end up with a panic are prohibited in this project (you could add `// #[allow_ci]` comment). In this case, however, as the function returns `Result`, I guess it could be simply written with `?`:
```suggestion
    Ok(verifier.verify(&base64::decode(signature.as_bytes())?)?)
```

ueno (approved):
Other than the comment, it looks good to me.

src/crypto.rs
ansasaki: OK, I'll make the change as suggested.

src/crypto.rs
```diff
@@ -244,4 +248,29 @@ mod tests {
...
+    fn test_asym_verify() {
+        // Import test keypair
+        let rsa_key_path = Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("test-data")
+            .join("test-rsa.pem");
+
+        // Get RSA keys
+        let contents = fs::read_to_string(rsa_key_path);
+        let private =
+            PKey::private_key_from_pem(contents.unwrap().as_bytes()).unwrap();
```
lkatalin, line 262:
```suggestion
            PKey::private_key_from_pem(contents.unwrap().as_bytes()).unwrap(); // #[allow_ci]
```

src/crypto.rs
```diff
@@ -244,4 +248,29 @@ mod tests {
...
+        let public = pkey_pub_from_priv(private).unwrap();
```
lkatalin, line 263:
```suggestion
        let public = pkey_pub_from_priv(private).unwrap(); // #[allow_ci]
```

src/crypto.rs
```diff
@@ -244,4 +248,29 @@ mod tests {
...
+
+        let message = String::from("Hello World!");
+
+        // Get known valid signature
+        let signature_path = Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("test-data")
+            .join("test-rsa.sig");
+
+        let signature = fs::read_to_string(signature_path).unwrap();
```
lkatalin, line 272:
```suggestion
        let signature = fs::read_to_string(signature_path).unwrap(); // #[allow_ci]
```

src/crypto.rs
```diff
@@ -244,4 +248,29 @@ mod tests {
...
+
+        assert!(asym_verify(&public, &message, &signature).unwrap())
```
lkatalin, line 274:
```suggestion
        assert!(asym_verify(&public, &message, &signature).unwrap()) // #[allow_ci]
```

lkatalin (commented):
@ansasaki If you add `// #[allow_ci]` on any lines that use an unavoidable `unwrap()` (like in the test functions), it should pass CI. That isn't a Rust thing, it's a specific test we wrote for this repo to avoid unnecessary `unwrap()`s.

Issue #300, opened 2022-01-20
Use revocation certificate from the path specified in config file


ansasaki:
The rust agent uses an hard-coded path for the revocation certificate instead of obtaining the path from the configuration file:
https://github.com/keylime/rust-keylime/blob/2c312e2d718ec994926e6c7621211cc7e7af1b16/src/revocation.rs#L129

Following the python agent behavior, the rust agent should obtain the certificate path from the configuration file:
https://github.com/keylime/keylime/blob/c58b19bc8a30691f61b12d272f227df3b251f25d/keylime/keylime_agent.py#L510-L517

PR #302, opened 2022-01-24
Use revocation certificate according to configuration

ansasaki:
The certificate to be used can be set through the ``revocation_cert`` entry in ``cloud_agent`` section of the configuration file.

Following the python agent behavior:

- If the ``revocation_cert`` is set as ``default``, use the default certificate path
- If the ``revocation_cert`` is an absolute path, use the provided path to obtain the certificate
- If the ``revocation_cert`` is a relative path, expand the path from the ``WORK_DIR``
- If the ``revocation_cert`` is empty, return an error

Fixes: #300

### DIFF
--- a/src/revocation.rs
+++ b/src/revocation.rs
@@ -4,14 +4,14 @@
 #[macro_use]
 use log::*;
 
-use crate::common::{KeylimeConfig, REV_CERT};
+use crate::common::{KeylimeConfig, REV_CERT, WORK_DIR};
 use crate::crypto;
 use crate::error::*;
 use crate::secure_mount;
 
 use std::convert::TryInto;
 use std::io::Write;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::{Command, Output, Stdio};
 
 use serde_json::Value;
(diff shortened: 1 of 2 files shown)


src/revocation.rs
```diff
@@ -274,4 +309,45 @@ mod tests {
...
+        let mut expected = PathBuf::from(WORK_DIR);
+        expected.push("secure/unzipped/");
+        expected.push(REV_CERT);
+        assert_eq!(*revocation_cert_path, expected);
+    }
+
+    #[test]
+    fn get_revocation_cert_path_absolute() {
+        let mut test_config = KeylimeConfig::default();
+        test_config.revocation_cert = String::from("/test/cert.crt");
```
lkatalin, line 327: I looked into the `clippy` error on this that is failing CI. If you initialize this way instead, it should get rid of the error:

```
let test_config = KeylimeConfig {
            revocation_cert: String::from("/test/cert.crt"),
            ..Default::default()
        };
```
And of course run `cargo fmt` after!

src/revocation.rs
```diff
@@ -274,4 +309,45 @@ mod tests {
...
+        let revocation_cert_path =
+            get_revocation_cert_path(&test_config).unwrap();
+        assert_eq!(revocation_cert_path, PathBuf::from("/test/cert.crt"));
+    }
+
+    #[test]
+    fn get_revocation_cert_path_relative() {
+        let mut test_config = KeylimeConfig::default();
+        test_config.revocation_cert = String::from("cert.crt");
```
lkatalin, line 336: Same comment as above.

lkatalin (commented):
Once the `KeylimeConfig` initialization is changed slightly and this is rebased on master, I think it should pass CI. The code itself LGTM and I tested it locally.

src/revocation.rs
ansasaki: OK, I'll modify as suggested.

src/revocation.rs
ansasaki: OK!

src/revocation.rs
```diff
@@ -274,4 +309,51 @@ mod tests {
         let outputs = run_revocation_actions(json, &test_config.secure_size);
         assert!(outputs.is_err());
     }
+
+    #[test]
+    fn get_revocation_cert_path_default() {
+        let test_config = KeylimeConfig::default();
+        let revocation_cert_path =
+            get_revocation_cert_path(&test_config).unwrap();
```
lkatalin, line 317: Looks like the CI is failing on a few of these `unwrap()`s still.

src/revocation.rs
```diff
@@ -274,4 +309,51 @@ mod tests {
...
+    }
+
+    #[test]
+    fn get_revocation_cert_path_absolute() {
+        let mut test_config = KeylimeConfig {
+            revocation_cert: String::from("/test/cert.crt"),
+            ..Default::default()
+        };
+        let revocation_cert_path =
+            get_revocation_cert_path(&test_config).unwrap();
```
lkatalin, line 331: Here as well.

src/revocation.rs
```diff
@@ -274,4 +309,51 @@ mod tests {
...
+    }
+
+    #[test]
+    fn get_revocation_cert_path_relative() {
+        let mut test_config = KeylimeConfig {
+            revocation_cert: String::from("cert.crt"),
+            ..Default::default()
+        };
+        let revocation_cert_path =
+            get_revocation_cert_path(&test_config).unwrap();
```
lkatalin, line 342: And here.

src/revocation.rs
ansasaki: My bad, I'll fix this

lkatalin (approved):
Thanks @ansasaki

PR #308, opened 2022-01-27
Use Path and PathBuf consistently to represent paths

ueno:
We previously used different types (&str, String, etc) to represent
filesystem paths and converted between them and Path back and forth.

This standardizes those to use Path and PathBuf consistently.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -10,6 +10,7 @@ use std::convert::TryFrom;
 use std::env;
 use std::ffi::CString;
 use std::fs::File;
+use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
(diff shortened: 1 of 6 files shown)


THS-on:
Do we want to move our constant paths like `/var/lib/keylime` also to Path/PathBuf?
https://github.com/keylime/rust-keylime/blob/94a19e236b2b3dd32543cf26aa555db2dbdeddd4/src/common.rs#L26-L34

ueno:
> Do we want to move our constant paths like `/var/lib/keylime` also to Path/PathBuf?

It's tempting, but I suspect it might not be possible because there are no Path/PathBuf creation methods which can be used as a constant expression.

lkatalin (approved):
This is great.

Issue #271, opened 2021-10-14
Add systemd unit file and expand documentation for deployment


ueno:
It would be nice to port the [systemd unit file](https://github.com/keylime/keylime/blob/master/services/keylime_agent.service) currently only hosted in the main keylime repository, and/or expand the README.md to cover steps to deploy the service in production environment.

Inspired by the [comment](https://github.com/coreos/fedora-coreos-tracker/issues/982#issuecomment-943283946) from @lucab.

PR #309, opened 2022-01-27
Install systemd unit file

ueno:
Also add a Makefile that wraps cargo and install commands.

Fixes: #271

### DIFF
--- a/dist/systemd/system/keylime_agent.service
+++ b/dist/systemd/system/keylime_agent.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=The Keylime compute agent
+
+[Service]
+# To run the agent as a non-root user, setup the environment following
+# the procedure at:
+# https://github.com/keylime/keylime/blob/master/agent_non_root_condition.sh
+# and uncomment the following lines:
+# User=keylime
+# Group=tss
+ExecStart=/usr/bin/keylime_agent
+KillSignal=SIGINT
+
+[Install]
+WantedBy=default.target
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 Keylime Authors
+
+RELEASE ?= 0
+TARGETDIR ?= target
+
+ifeq ($(RELEASE),1)
+        PROFILE ?= release
+        CARGO_ARGS = --release
+else
+        PROFILE ?= debug
+        CARGO_ARGS =
+endif
+
+systemdsystemunitdir := $(shell pkg-config systemd --variable=systemdsystemunitdir)
+
+.PHONY: all
+
+.PHONY: build
+build:
+	cargo build --target-dir="${TARGETDIR}" ${CARGO_ARGS}
+
+.PHONY: install
+install: build
+	install -D -t ${DESTDIR}/usr/bin "${TARGETDIR}/${PROFILE}/keylime_agent"
+	install -D -t ${DESTDIR}/usr/bin "${TARGETDIR}/${PROFILE}/keylime_ima_emulator"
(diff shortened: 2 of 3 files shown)


dist/systemd/system/keylime_agent.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=The Keylime compute agent
+
+[Service]
+ExecStart=/usr/bin/keylime_agent
```
sergio-correia, line 5: Do we want to specify a (non-root) user here?

THS-on: Yes it is a good idea to run keylime not as root by default. Example on what we do for the Debian package:
https://github.com/utkarsh2102/python-keylime/blob/master/debian/python3-keylime-agent.keylime_agent.service

Also the mount for `/var/lib/keylime/secure` can then created by systemd:
https://github.com/utkarsh2102/python-keylime/blob/master/debian/python3-keylime-agent.var-lib-keylime-secure.mount

sergio-correia: @THS-on: nice! Can't (some of) these changes be upstreamed?

THS-on: @sergio-correia the current installer.sh script just assumes that you run everything as root which is fine because it is mostly used for testing and CI. It is up to the distributions on how to package keylime and if for example they want to create a new user etc.

If there is interest I see no reason why not to include those as an example on how to run Keylime without root.

lkatalin (commented):
I'll defer to others on the non-root user issue. I also think it may be good to have a few lines of documentation about these added files and how to invoke the makefile to do different things.

lkatalin (changes requested):
See above

lkatalin:
Related: https://github.com/keylime/keylime/issues/864

dist/systemd/system/keylime_agent.service
lkatalin: @THS-on @sergio-correia @ueno Was I reading this correctly that we are still waiting for a change on this PR re: a non-root user?

THS-on: No, what the Debian package currently does is not ideal and will not work for packages  actually shipped in distributions. So we can merge it as currently is.

See discussion here: https://github.com/keylime/keylime/issues/864

lkatalin:
> I'll defer to others on the non-root user issue. I also think it may be good to have a few lines of documentation about these added files and how to invoke the makefile to do different things.

@ueno Is it possible to add a few lines of documentation in the `README.md` for these changes?

ueno:
@lkatalin sorry for the delay; added some comments in README.md and keylime_agent.service.

dist/systemd/system/keylime_agent.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=The Keylime compute agent
+
+[Service]
+# To run the agent as a non-root user, setup the environment following
+# the procedure at:
+# https://github.com/keylime/keylime/blob/master/agent_non_root_condition.sh
+# and comment out the following lines:
```
lkatalin, line 8:
```suggestion
# and uncomment the following lines:
```

lkatalin:
> @lkatalin sorry for the delay; added some comments in README.md and keylime_agent.service.

@ueno No problem, thank you! Looks great.

lkatalin:
@ueno Sorry I've tried to make one small change in wording without asking you to re-push, but it seems I've mucked it up and can't revert the commit as I don't have force-push access to your branch. Do you think you could re-push it when you get a chance? Then I will be able to merge it.

ueno:
@lkatalin thanks, merged the fix and force-pushed

lkatalin (approved):
Thank you, @ueno !

PR #316, opened 2022-01-28
Inital support for API version 2.0

### LINKED ISSUE #311: Bump API version
Keylime is dropping support for API v1.0 https://github.com/keylime/keylime/pull/862

We should make sure the Rust agent is compliant with any 2.0 requirements and bump the [API version](https://github.com/keylime/rust-keylime/blob/318dfdec5ce30c520e5f17bc2b03c422a198a1a4/src/common.rs#L22).

THS-on:
This implements the following changes for the 2.0 API:

* dropping zlib compression in quotes
* The agent now runs a HTTPS server which validates the connections against the Keylime CA using mTLS (fixes https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r)
* API prefix is updated to `v2.0`

Note that we currently do not persist the transport key and mtls certificate to disc which means that after a restart attestation of the agent will fail because the agent. This will be addressed in a separate PR which extends https://github.com/keylime/rust-keylime/pull/293 to a more general persistence file.

Mostly resolves #311


lkatalin:
> Note that we currently do not persist the transport key and mtls certificate to disc which means that after a restart attestation of the agent will fail because the agent. This will be addressed in a separate PR which extends #293 to a more general persistence file.

Should we wait for the additional PR extending #293 to come in first, so that we don't get new failures with this change?

THS-on:
I would still merge this before I have implemented the other change to allow the agent to work with API v2.0,

Not implementing this breaks nothing other than restarting the agent mid attestation will produce an failure, which is currently a nice to have feature anyway.

src/crypto.rs
```diff
@@ -69,6 +80,51 @@ pub(crate) fn pkey_pub_from_priv(
...
+    let valid_from = Asn1Time::days_from_now(0)?;
+    let valid_to = Asn1Time::days_from_now(356)?;
+
+    let mut builder = X509::builder()?;
+    builder.set_version(2)?;
+    builder.set_subject_name(&name)?;
+    builder.set_issuer_name(&name)?;
+    builder.set_not_before(&valid_from)?;
+    builder.set_not_after(&valid_to)?;
+    builder.set_pubkey(key)?;
```
ueno, line 97: Although this use-case is a bit special, I wonder if it would make sense to add some common restrictions such as (E)KU, SAN, etc?

ueno (approved):

src/crypto.rs
THS-on: SAN does not make sense because we ignore it on the server side. We do this because we don't know on which IP or address we use to connect to the agent beforehand.

What attributes would you set for (E)KU?

src/crypto.rs
```diff
@@ -36,6 +41,12 @@ pub(crate) fn import_x509(input_key_path: &Path) -> Result<PKey<Public>> {
     cert.public_key().map_err(Error::Crypto)
 }
 
+pub(crate) fn load_x509(input_cert_path: &str) -> Result<X509> {
```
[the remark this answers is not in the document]
THS-on: Yeah it makes sense to unify those functions.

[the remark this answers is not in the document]
THS-on: Done :)

lkatalin:
> I would still merge this before I have implemented the other change to allow the agent to work with API v2.0,
> 
> Not implementing this breaks nothing other than restarting the agent mid attestation will produce an failure, which is currently a nice to have feature anyway.

Sounds good, I will give it another look shortly.

src/crypto.rs
```diff
@@ -19,10 +24,9 @@ use crate::{
     Error, Result, AES_128_KEY_LEN, AES_256_KEY_LEN, AES_BLOCK_SIZE,
 };
 
-// Reads an X509 cert chain (provided by the tenant with the --cert command) and outputs
-// its public key.
-pub(crate) fn import_x509(input_key_path: &Path) -> Result<PKey<Public>> {
-    let contents = fs::read_to_string(input_key_path)?;
+// Read a X509 cert or cert chain and outputs the first certificate
```
lkatalin, line 27: What do you mean "first" certificate? Is this the leaf?

lkatalin (commented):
This looks pretty good to me, I'm just doing some final testing. @ueno did you have anything else you wanted to see re: the (E)KU or should we merge if testing works on my end?

lkatalin (approved):
Local tests passed so LGTM

src/crypto.rs
THS-on: We error if there find more than one, so we expect exactly one leaf

lkatalin:
Since there are two approvals, I will go ahead and merge this. We can make any extra changes in a new PR.

PR #329, opened 2022-02-15
Depend on Rust-TSS-ESAPI 7.0.0 stable

lkatalin:
The [upstream issue](https://github.com/parallaxsecond/rust-tss-esapi/issues/290) is now closed. :tada:


lkatalin:
@ueno Since you had trouble with `7.0.0-beta1`, would you be able to confirm this works for you?

ueno (approved):
Thanks, it compiles and works OK with a quick test.

Issue #298, opened 2022-01-19
Add a way to compile out ZeroMQ support


ueno:
As noted in #199 libzmq brings in quite a few dependencies, while this feature is not always desired. It might be worth considering a Cargo feature to compile out the support. This could be done something like in Cargo.toml:
```toml
[features]
default = ["revocation"]
revocation = ["zmq"]
```
in the source, using the [cfg_if](https://docs.rs/cfg-if/1.0.0/cfg_if/) crate:
```rust
cfg_if::cfg_if! {
    if #[cfg(feature = "revocation")] {
        ...
    }
}
```
and then you could compile the crate with `--no-default-features`.

Additional note:
- Proper support for revocation events (including actions) through alternative means would need to wait for keylime/enhancements#54 and keylime/enhancements#56 being implemented, which is out of scope of this issue
- Making it a dynamically loadable [plugin](https://michael-f-bryan.github.io/rust-ffi-guide/dynamic_loading.html) might be an option, though I guess it would require significant work

lkatalin:
I like the idea of having this as a feature.

Re: the dynamic plugin, I thought this was already the approach we're taking since each revocation action takes some type of JSON input? [Example](https://github.com/keylime/rust-keylime/blob/master/src/revocation.rs#L20)

lkatalin:
@ueno has mentioned that the way forward on this is by implementing the proposal in https://github.com/keylime/enhancements/blob/master/53_agent_local_revocation.md. @ansasaki is this something that you are planning to implement?

Do we still also want to have revocation as an optional feature as well? It could make sense.

ueno:
> Re: the dynamic plugin, I thought this was already the approach we're taking since each revocation action takes some type of JSON input? [Example](https://github.com/keylime/rust-keylime/blob/master/src/revocation.rs#L20)

To be clear, I meant to dynamically load libzmq itself, not the revocation actions.

> Do we still also want to have revocation as an optional feature as well? It could make sense.

I believe this is still necessary, because libzmq and its dependencies are not always available (e.g., on CentOS/RHEL) nor usable under FIPS restrictions.

ansasaki:
> @ueno has mentioned that the way forward on this is by implementing the proposal in https://github.com/keylime/enhancements/blob/master/53_agent_local_revocation.md. @ansasaki is this something that you are planning to implement?

Yes, I'm planning to implement that enhancement

lkatalin:
[...]

PR #331, opened 2022-02-15
Add REST endpoint to receive revocation notifications

ansasaki:
This implements the agent side of the enhancement proposal: https://github.com/keylime/enhancements/blob/master/53_agent_local_revocation.md

Add the `/notifications/revocation` endpoint to the agent which handles revocation notifications delivered by POST requests.
The format of the notification message is the same used for notifications delivered via ZeroMQ.

This also makes ZeroMQ an optional dependency enabled by default. It is possible to compile without `zmq` by passing the `--no-default-features` options to cargo when building:
```
$ cargo build --no-default-features
```
Fixes: #298


lkatalin (commented):
This LGTM! Thanks @ansasaki  for making this change so quickly.

According to the enhancement it seems like the API version should be bumped - would this break communication with the Python components? I can't remember if they specify a specific API version or "at least" a specific version.

Similar point about the config file, there is a change mentioned in the enhancement but that needs to change in the Python repo to actually be used by the agent.

Is there any downside of potentially having both the REST revocation listener and the zeromq revocation service on at the same time? Should we only enable one at a time? I guess since the revocation message has to be signed, it should make no difference as long as the verifier is behaving in a sane way.

THS-on:
> According to the enhancement it seems like the API version should be bumped - would this break communication with the Python components? I can't remember if they specify a specific API version or "at least" a specific version.

The API version should be increased to `2.1`, because it is only an addition to the API. On the Python side the the connection should still work, because currently we have no checks for that in place. The API on the server side should be increased once https://github.com/keylime/keylime/pull/795 is merged.

ansasaki:
> I'll add a commit to bump the API version.

In the end I amended the commit which introduces the new API to bump the version.

lkatalin:
> > I'll add a commit to bump the API version.
> 
> In the end I amended the commit which introduces the new API to bump the version.

@ansasaki Perfect, thanks!

lkatalin (commented):
LGTM except small nit / confusing on the first commit's message: "Also fix the comment for the behavior when the revocation_cert from the configuration file is empty." I don't see the comment there, did it end up going in a different commit?

ansasaki:
> LGTM except small nit / confusing on the first commit's message: "Also fix the comment for the behavior when the revocation_cert from the configuration file is empty." I don't see the comment there, did it end up going in a different commit?

Oh, sorry about that, the change was included in a commit (ab7ed3c7aa773f6cf9c3605c79ca475bde117fc3) in #321.

I'll amend the message to remove that part.

lkatalin (approved):

PR #334, opened 2022-02-28
Update actix-web to 4, remove tokio 0.2 dependencies

ueno:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,97 +4,57 @@ version = 3
 
 [[package]]
 name = "actix-codec"
-version = "0.3.0"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "78d1833b3838dbe990df0f1f87baf640cf6146e898166afe401839d1b001e570"
+checksum = "57a7559404a7f3573127aab53c08ce37a6c6a315c374a31070f3c91cd1b4a7fe"
 dependencies = [
  "bitflags",
- "bytes 0.5.6",
+ "bytes",
  "futures-core",
  "futures-sink",
  "log",
- "pin-project 0.4.29",
- "tokio 0.2.25",
- "tokio-util 0.3.1",
-]
-
-[[package]]
-name = "actix-connect"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "177837a10863f15ba8d3ae3ec12fac1099099529ed20083a27fdfe247381d0dc"
-dependencies = [
- "actix-codec",
- "actix-rt 1.1.1",
- "actix-service",
- "actix-utils",
- "derive_more",
- "either",
- "futures-util",
- "http",
- "log",
- "openssl",
- "tokio-openssl",
- "trust-dns-proto",
- "trust-dns-resolver",
+ "memchr",
+ "pin-project-lite",
+ "tokio",
+ "tokio-util 0.7.1",
 ]
 
 [[package]]
 name = "actix-http"
-version = "2.2.2"
+version = "3.0.4"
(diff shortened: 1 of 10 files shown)


lkatalin:
I am testing this on top of #342 . For some reason, it seems to hang indefinitely after agent activation, so the agent does not seem to receive the requests for identity or integrity quote. @ueno are you able to get through the whole attestation process with these changes?

ueno:
@lkatalin obviously I haven't tried the actual attestation, sorry about that. After going through the [migration guide](https://github.com/actix/actix-web/blob/master/actix-web/MIGRATION-4.0.md), it seems that we need to explicitly spawn the Actix server and revocation notifier threads, something like:
```rust
let server_task = rt::spawn(actix_server).map_err(Error::from);
let worker_task =
    rt::spawn(worker(symm_key, symm_key_cvar, payload, config.clone()))
        .map_err(Error::from);

let result = try_join!(server_task, worker_task);
```
I'll update the PR after I manage to fix other issues (e.g., `ResponseBody` and error handlers have been reworked). Until then I'm marking this as draft.

ueno:
I still haven't gone through proper attestation scenario, but rebased against the master.
Tagging @ansasaki for awareness as it touches `errors_handler.rs`.

ansasaki (commented):
The changes LGTM, the only thing to fix are the clippy errors (unneeded `return` at the end of the scope and unneeded mutable reference).

After those are fixed and the CI passes, I will run some additional tests against the python services and then I can approve the PR.

BTW, with this merged we could retry switching to the `zeromq` crate to drop some extra dependencies.

lkatalin:
@ueno Thanks for the update, I just tested this new version and the quote validation works now!

ueno:
@ansasaki I talked to the actix-web devs on the linked issue. The suggestion was, if we ever want to handle graceful shutdown with `ctrl-c`:
1. disable the default signal handling of `HttpServer`
1. create a new task waiting for it through `tokio::signal::ctrl_c`
1. once `ctrl-c` is received propagate it to all tasks and also call `ServerHandle::stop`

Given the complexity of the solution and the fact that we haven't cared about graceful shutdown of the agent, I suppose we could work around the issue only with (1), which I've incorporated in the latest commit. Maybe we could file a follow-up issue to implement it properly.

ansasaki:
> @ansasaki I talked to the actix-web devs on the linked issue. The suggestion was, if we ever want to handle graceful shutdown with `ctrl-c`:
> 
>     1. disable the default signal handling of `HttpServer`
> 
>     2. create a new task waiting for it through `tokio::signal::ctrl_c`
> 
>     3. once `ctrl-c` is received propagate it to all tasks and also call `ServerHandle::stop`
> 
> 
> Given the complexity of the solution and the fact that we haven't cared about graceful shutdown of the agent, I suppose we could work around the issue only with (1), which I've incorporated in the latest commit. Maybe we could file a follow-up issue to implement it properly.

OK, sounds reasonable. Let's create the follow-up issue for the graceful shutdown and unblock this to be merged.

keylime.conf
```diff
@@ -13,7 +13,7 @@ receive_revocation_port = 8992
 #=============================================================================
 
 # The binding address and port for the agent server
-cloudagent_ip = 127.0.0.1
+cloudagent_ip = 0.0.0.0
```
ansasaki, line 16: I think the `keylime.conf` file changes were not intentional. Could you confirm?

ansasaki (changes requested):
I think this could be merged after the `keylime.conf` changes are reverted and the workaround which disables the default signal handlers with `disable_signals()` is added (and the follow-up issue to address this is created).

keylime.conf
ueno: Right, it was not intentional.

ansasaki (approved):
LGTM!

Issue #333, opened 2022-02-22
Implement use of KEYLIME_DIR env var


lkatalin:
Use the env variable `KEYLIME_DIR` to set the current working directory. The default is `/var/lib/keylime`. But, during e2e testing, it should be set to something like `/root/keylime`. I believe the e2e tests would have to set this var, it isn't set in the agent; but the agent needs to recognize it and choose locations like [where to look for the CA cert](https://github.com/keylime/rust-keylime/blob/master/src/common.rs#L34) based on it.

ansasaki:
Can I assign this to myself?

lkatalin:
> Can I assign this to myself?

@ansasaki Absolutely! Thanks for volunteering.

PR #335, opened 2022-03-01
Set working directory via KEYLIME_DIR env variable

ansasaki:
The working directory can be set via the KEYLIME_DIR environment
variable or keylime_dir configuration option.

When the working directory is set, the expected location for the
certificates and secure mount are also modified and are relative to the
location set via KEYLIME_DIR.

The DEFAULT_CA_PATH is now relative to the working directory.

Consolidate the conditional change of the secure mount location when
testing to a single location.

Modify the revocation tests to use a temporary directory as the working
directory.

Fixes: #333

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -28,16 +28,20 @@ registrar_port = 8890
 # The name of the RSA key that Keylime should use for protecting shares of U/V.
 rsa_keyname = tci_rsa_key
 
+# The keylime working directory.  Can be overriden by setting the KEYLIME_DIR
+# environment variable. The default value is /var/lib/keylime
+# keylime_dir = /var/lib/keylime
+
 # The CA that signs the client certificates of the tenant and verifier.
-# If set to default it tries to use /var/lib/keylime/cv_ca/cacert.crt
+# If set to default it tries to use $keylime_dir/cv_ca/cacert.crt
 keylime_ca = default
 
 # The name that should be used for the encryption key, placed in the
-# /var/lib/keylime/secure/ directory.
+# $keylime_dir/secure/ directory.
 enc_keyname = derived_tci_key
 
 # The name that should be used for the optional decrypted payload, placed in
-# the /var/lib/keylime/secure directory.
+# the $keylime_dir/secure directory.
 dec_payload_file = decrypted_payload
 
 # The size of the memory-backed tmpfs partition where Keylime stores crypto keys.
@@ -52,7 +56,7 @@ tpm_ownerpassword = keylime
 
 # Whether to allow the cloud_agent to automatically extract a zip file in
 # the delivered payload after it has been decrypted, or not. Defaults to "true".
-# After decryption, the archive will be unzipped to a directory in /var/lib/keylime/secure.
+# After decryption, the archive will be unzipped to a directory in $keylime_dir/secure.
 # Note: the limits on the size of the tmpfs partition set above with the 'secure_size'
 # option will affect this.
 extract_payload_zip = True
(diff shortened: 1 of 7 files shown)


mpeters:
> One thing I notice is the changes to `keylime.conf`. Since this file is always installed from Python Keylime, any changes to it from the Rust repo are essentially a no-op. Longer term I think this conf file should live upstream to both Python and Rust Keylime, but for now it seems to make more sense to include any changes to it in the Python repo instead and mirror them here. What do you think about this?

I don't think this is necessary. Long term and for most users they will install keylime and the rust agent via their distro's packaging system, so it's really a packaging concern (as most default configuration issues are). It's also going to be uncommon to install both keylime and the rust agent on the same host (though not impossible if you have keylime monitoring itself).

The keylime.conf shipped by either keylime or rust-keylime are really for development, not actual end users.

lkatalin:
> > One thing I notice is the changes to `keylime.conf`. Since this file is always installed from Python Keylime, any changes to it from the Rust repo are essentially a no-op. Longer term I think this conf file should live upstream to both Python and Rust Keylime, but for now it seems to make more sense to include any changes to it in the Python repo instead and mirror them here. What do you think about this?
> 
> I don't think this is necessary. Long term and for most users they will install keylime and the rust agent via their distro's packaging system, so it's really a packaging concern (as most default configuration issues are). It's also going to be uncommon to install both keylime and the rust agent on the same host (though not impossible if you have keylime monitoring itself).
> 
> The keylime.conf shipped by either keylime or rust-keylime are really for development, not actual end users.

Thanks @mpeters . I now see your point that it wouldn't be a normal flow for most users. A lesser but existing concern for me is that I want to make development straightforward as well. Maybe I missed this, is our current process to mirror any changes to conf from the Python repo here as they happen? I want to make sure the two conf files don't diverge, even if the Rust one has "extra" stuff.

ansasaki:
> One thing I notice is the changes to `keylime.conf`. Since this file is always installed from Python Keylime, any changes to it from the Rust repo are essentially a no-op. Longer term I think this conf file should live upstream to both Python and Rust Keylime, but for now it seems to make more sense to include any changes to it in the Python repo instead and mirror them here. What do you think about this?

> A lesser but existing concern for me is that I want to make development straightforward as well. Maybe I missed this, is our current process to mirror any changes to conf from the Python repo here as they happen? I want to make sure the two conf files don't diverge, even if the Rust one has "extra" stuff.

I agree with @mpeters, that most of the users will get their configuration files from their distribution packaging system (and will probably customize them), but also agree that would be nice to have configuration files in sync. I understand this is difficult, so I would propose all used configuration options to have defaults, to remove the requirement to have a configuration file to run minimally (for example for testing purposes). This way extra options added to one side would not affect the other side (e.g. an option added to rust configuration would have a default that keeps it working when running with the python configuration file). The option added to the configuration file in this PR follows this rule.

With time we can work towards a single configuration file, which will probably happen at the same time as we converge to a single agent implementation.

lkatalin:
> I agree with @mpeters, that most of the users will get their configuration files from their distribution packaging system (and will probably customize them), but also agree that would be nice to have configuration files in sync. I understand this is difficult, so I would propose all used configuration options to have defaults, to remove the requirement to have a configuration file to run minimally (for example for testing purposes). This way extra options added to one side would not affect the other side (e.g. an option added to rust configuration would have a default that keeps it working when running with the python configuration file). The option added to the configuration file in this PR follows this rule.
> 
> With time we can work towards a single configuration file, which will probably happen at the same time as we converge to a single agent implementation.

This is reasonable, thank you @ansasaki . With these defaults in place I won't worry about the config files diverging for the moment. We can wait to consolidate them until the Rust agent can become the primary agent.

I will continue the review and hopefully have more comments soon.

lkatalin (approved):
This LGTM. I was not able to get the Python Keylime CI tests to pass with it yet, but the error could be with the CI tests or something that is unrelated to these changes. I will go ahead and merge and we can make adjustments if needed.

PR #337, opened 2022-03-01
Remove revocation actions naming restriction

ansasaki:
The revocation actions names are no longer required to start with
'local_action_'.

This implements the enhancement amendment added via:
https://github.com/keylime/enhancements/pull/63


lkatalin (approved):
Thanks!

PR #277, opened 2021-10-20
Add support for iterative attestation by returning the end of the IMA measurement log 

stefanberger:
This PR adds support for iterative attestation by returning the end of the IMA measurement log starting at a certain point requested by the validator. This code is a translation of the code the python agent to Rust.

PR #339, opened 2022-03-02
Add support for iterative attestation

ansasaki:
Rebase the changes proposed in #277 on top of current master.

Small changes were added:
- Applied changes suggested by `cargo fmt` and `cargo clippy`
- Avoid using `unwrap()`, `panic!()`, and cloning when possible
- Move added unit test to a module

### DIFF
--- a/src/ima.rs
+++ b/src/ima.rs
@@ -1,0 +1,155 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2021 Keylime Authors
+
+#[macro_use]
+use log::*;
+
+use std::collections::HashSet;
+use std::fs::File;
+use std::io::prelude::*;
+use std::io::Error;
+use std::io::SeekFrom;
+use std::path::Path;
+use std::str;
+
+use tempfile::NamedTempFile;
+
+/// IMAMeasurementList models the IMA measurement lists's last two known
+/// numbers of entries in the log and filesizes at that point
+#[derive(Debug)]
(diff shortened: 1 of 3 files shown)


src/ima.rs
```diff
@@ -0,0 +1,155 @@
...
+        let _ = iml.reset();
+        iml
+    }
+
+    fn reset(&mut self) {
+        self.entries = HashSet::new();
+    }
+
+    fn update(&mut self, num_entries: u64, filesize: u64) {
+        if self.entries.len() > 32 {
```
lkatalin, line 33: What is the significance of the len of entries here?

lkatalin (approved):
This also LGTM.

src/ima.rs
ansasaki: If I understood the code correctly, the structure will keep the latest 32 updates to the offsets to speed up finding the optimal file seek. It is basically a cache of the latest queries results and 32 was an arbitrary chosen number of entries.

The interesting part is that since the used structure is not ordered, the entry dropped when the 33rd query arrives is not the oldest as one would expect, but a random entry. This don't affect the results, but I wonder if that is the expected result when the implementer wrote the code. I believe the algorithm assumes the queries to be random accesses, but in reality I think in most of the cases the query will be closer to the latest entry. I think there is room for improvement.

lkatalin: That actually is very interesting, I was wondering about that! Thank you for this explanation. I wasn't familiar with how this works.

Issue #340, opened 2022-03-02
Resolve API error


lkatalin:
Looks like bumping the API version to `2.1` broke something when communicating with the registrar:

```
INFO  keylime_agent::registrar_agent > Requesting agent registration from http://127.0.0.1:8890/v2.1/agents/d432fbb3-d2f1-4a97-9ef7-75bd81c00000 for d432fbb3-d2f1-4a97-9ef7-75bd81c00000
Error: Registrar { addr: "http://127.0.0.1:8890/v2.1/agents/d432fbb3-d2f1-4a97-9ef7-75bd81c00000", code: 400 }
```

lkatalin:
Fixed in #342

PR #341, opened 2022-03-03
Specify API version of Python components

lkatalin:
Since the Rust agent and Python components can and have diverged
by a minor version, the handlers should expect the API version of
the Python components, not the API version of the Rust agent.

Related to #340, but does not fix it until a change goes in on the Python side as well.

Is there some better method to make sure the `PYTHON_API_VERSION` stays in sync with the Python repo's API version, other than manually? I can't think of one.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -19,7 +19,10 @@ use uuid::Uuid;
 /*
  * Constants and static variables
  */
-pub const API_VERSION: &str = "v2.1";
+// expected API version of verifier, registrar, tenant
+pub const PYTHON_API_VERSION: &str = "v2.0";
+// API version of agent
+pub const AGENT_API_VERSION: &str = "v2.1";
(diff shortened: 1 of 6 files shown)


lkatalin:
Parking this while we sort stuff out on https://github.com/keylime/keylime/pull/907

lkatalin:
> Parking this while we sort stuff out on [keylime/keylime#907](https://github.com/keylime/keylime/pull/907)

Closing this - per the discussion in the link above, the Rust agent is going to follow the API versioning from the Python repo rather than having its own versioning. This is because a newer agent talking to older server-side components is not supported per the original API [enhancement](https://github.com/keylime/enhancements/blob/master/45_api_versioning.md#notesconstraintscaveats):

```
This proposal does not guarantee the ability for a newer client (using say v3.4) to communicate with an older server-side component (using say v3.3).
```

PR #53, opened 2019-03-15
Add KeylimeTpmError support and change return type in tpm.rs

leonjia0112:
Add support for tpm custom error type. This is aim on solving #40 for changing return type.
Implementation details:
1. Implement KeylimeTpmError enum has two contribution
    - error generate from TPM command execution
    - error generate from rust part of TPM interface
2. Implement From trait for Error:
    - `serde_json::error::Error`
    - `std::num::ParseIntError`    
    - `std::string::FromUtf8Error`
    - `std::io::Error`
    - `std::time::SystemTimeError`    
3. Upate return type of `run()` function in a tuple and a custom error
    `Result<(String, String), TpmExecError>`    
4. Support return type of other function in tpm.rs with `KeylimeTpmError`, which has benefit in code efficient and early return efficient

PR #342, opened 2022-03-08
Revert API version to 2.0

lkatalin:
It was updated to 2.1 based on Keylime enhancement #53
(agent-local revocation), but this broke communication
with the Python components. We have [decided](https://github.com/keylime/keylime/pull/907) that the
Rust agent should not have its own API version, but
rather follow API versioning from the Python components.


lkatalin:
@lukehinds @ueno @ansasaki @mpeters Any chance I can get a quick approval on this? It will un-break communication with server-side.

mpeters (approved):

PR #348, opened 2022-03-15
Use shared JSON wrapper for HTTP responses

ansasaki:
Use a generic JSON wrapper implementation for HTTP responses instead of implementing a specific wrapper for each JSON content.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 use std::convert::TryFrom;
 use std::env;
 use std::ffi::CString;
+use std::fmt::Debug;
 use std::fs::File;
 use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
@@ -66,6 +67,26 @@ cfg_if::cfg_if! {
 
         pub(crate) fn ima_ml_path_get() -> PathBuf {
             Path::new(IMA_ML).to_path_buf()
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub(crate) struct JsonWrapper<A> {
+    pub code: u32,
+    pub status: String,
+    pub results: A,
+}
+
+impl<'de, A> JsonWrapper<A>
+where
+    A: Deserialize<'de> + Serialize + Debug,
+{
+    pub(crate) fn new(results: A) -> JsonWrapper<A> {
+        JsonWrapper {
+            code: 200,
+            status: String::from("Success"),
+            results,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/common.rs
```diff
@@ -70,6 +70,23 @@ cfg_if::cfg_if! {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub(crate) struct JsonWrapper<A: Serialize> {
```
THS-on, line 74: I'm actually not sure if `A` needs `Serialize` added explicitly or if the deriving automatically adds that. This has to be checked with someone with deeper Rust knowledge. @ueno any thoughts on this?

ueno: Weak opinion but I think it's a little clearer to explicitly state that `A` is bound to `Serialize` at struct definition; without it one could create a `JsonWrapper` with a non-serializable `results` but this error is caught only when it's actually being serialized.

THS-on: Do we then need to add this for `Deserialize` and `Debug`?

THS-on: I agree. All the other changes LGTM

lkatalin (approved):
This LGTM as well, I can merge it after tests pass.

lkatalin:
@ansasaki It looks like CI is stuck, maybe it would help if you can re-push with some trivial change.

src/common.rs
ansasaki: Searching the documentation, I found that derived `Serialize` and `Deserialize` add the trait bound for generic structure fields, which is probably sufficient in our case:
https://serde.rs/attr-bound.html

Moreover, trying to apply the bounds manually in the structure proved to be complicated due to the required lifetime binding for `Deserialize`. It is probably possible, but I failed to understand how to write the code (I need to study more rust).

Searching in the internet how people solved this, I found that it is somewhat unusual to restrict the traits in the structure for generic fields. Normally such restrictions are applied in the `impl` which defines the behavior.

For all above, I think it is sufficient to leave the derived `Serialize` and `Deserialize` to take care of the traits binding and add the trait restriction to the `impl`. 

@ueno @THS-on What do you think? If you know how to enforce the trait binding in the structure, feel free to suggest.

ps.: The result seems to be that `JsonWrapper::new()` will not accept types without the `Serialize` and `Deserialize` traits, but it would still be possible to manually create a `JsonWrapper` structure containing a structure in `results` without the traits, which is basically the issue @ueno pointed out.

lkatalin:
> @lkatalin Hello, could you re-run the CI?

@ansasaki I've just hit "re-run all checks," but I'm not convinced that it is working. If not, could you make some trivial change and re-push?

PR #353, opened 2022-03-22
keys_handler: Fix test for 256-bits keys

ansasaki:
The test_u_or_v_key_long() was using a 128 bits long key instead of 256
bits.


ueno (approved):

lkatalin (approved):
Great catch.

PR #356, opened 2022-03-30
Use JSON wrapper on error responses

ansasaki:
The response for requests should always be wrapped in the JSON structure, including in error cases.
    
Without this, the rust agent returns malformed responses on error cases, advertising `content-type: application/json` on the header, but sending plaintext on the body, which caused JSON parsing errors.

This was found during implementation of https://github.com/keylime/keylime/pull/947.


lkatalin (approved):
Ah, yes, I think I remember this bug - thank you for the fix!

PR #357, opened 2022-04-01
Rename JSON wrapper methods

ansasaki:
Rename `JsonWrapper::new` as `JsonWrapper::success` to clarify the purpose which is to wrap the content with the JSON structure with success status code.

Also replace `JsonWrapper::bad_request` with `JsonWrapper::error` which receives the error status code as a parameter, allowing wrapping the error message with different error status codes.


lkatalin (approved):
Fantastic.

Issue #343, opened 2022-03-09
Drop privileges of agent process after startup


lkatalin:
Mirror of https://github.com/keylime/keylime/pull/900/

Issue #344, opened 2022-03-09
services: add keylime_agent_secure.mount service


lkatalin:
Mirror of https://github.com/keylime/keylime/pull/903

Issue #317, opened 2022-01-28
Use /proc/self/mountinfo instead of mount for checking secure mount point


THS-on:
The output of `mount` does not escape spaces which might lead to unexpected behavior that fools the agent that the mount is actually there:  https://github.com/keylime/keylime/security/advisories/GHSA-wj36-qcfg-5j52

We fixed this in the Python agent by parsing `/proc/self/mountinfo` instead. I think the rust agent is also affected by this.

Logic from the Python agent can be found here: https://github.com/keylime/keylime/blob/15663f08b5d94ae716cd971b8a1215121cc1acb1/keylime/secure_mount.py#L19-L60

ansasaki:
Fixed via #364

PR #364, opened 2022-04-25
Drop privileges after startup

ansasaki:
After finishing operations that require privileges, drop the privileges to run as a normal user.

This adds support to the `run_as` configuration option to select the `user:group` under which the keylime agent will run.

This also rewrites `check_mount()` to use the information from `/proc/self/mountinfo` instead of the mount output file to check if `secure` is mounted. The implementation was based on python agent.

Fixes: #343, #317, #344

### DIFF
--- a/src/ima.rs
+++ b/src/ima.rs
@@ -25,7 +25,7 @@ impl ImaMeasurementList {
         }
     }
 
-    fn reset(&mut self) {
+    pub(crate) fn reset(&mut self) {
         self.entries = HashSet::new();
     }
 
@@ -57,15 +57,9 @@ impl ImaMeasurementList {
 /// was read and the current number of entries in the file.
 pub(crate) fn read_measurement_list(
     ima_ml: &mut ImaMeasurementList,
-    filename: &Path,
+    ima_file: &mut File,
     nth_entry: u64,
 ) -> IMAError {
-    if !Path::new(filename).exists() {
-        let _ = ima_ml.reset();
-        warn!("IMA measurement list not available: {}", filename.display());
-        return Ok((None, None, None));
-    }
-
     let mut nth_entry = nth_entry;
 
     // Try to find the closest entry to the nth_entry
@@ -73,9 +67,8 @@ pub(crate) fn read_measurement_list(
 
     let mut ml = None;
     let mut filedata = String::new();
-    let mut file = File::open(filename)?;
-    let _ = file.seek(SeekFrom::Start(filesize))?;
-    let _ = file.read_to_string(&mut filedata)?;
+    let _ = ima_file.seek(SeekFrom::Start(filesize))?;
+    let _ = ima_file.read_to_string(&mut filedata)?;
     let mut offset: usize = 0;
 
     loop {
@@ -94,7 +87,7 @@ pub(crate) fn read_measurement_list(
     let _ = ima_ml.update(num_entries, filesize + offset as u64);
 
     match ml {
-        None => read_measurement_list(ima_ml, filename, 0),
+        None => read_measurement_list(ima_ml, ima_file, 0),
         Some(slice) => Ok((
             Some(String::from(slice)),
             Some(nth_entry),
@@ -116,16 +109,18 @@ mod tests {
         tf.write_all(filedata.as_bytes());
         tf.flush();
 
+        let mut ima_file = File::open(tf.path()).unwrap(); //#[allow_ci]
+
         // Request the 2nd entry, which is available
         let (ml, nth_entry, num_entries) =
-            read_measurement_list(&mut ima_ml, tf.path(), 2).unwrap(); //#[allow_ci]
+            read_measurement_list(&mut ima_ml, &mut ima_file, 2).unwrap(); //#[allow_ci]
         assert_eq!(num_entries, Some(3));
         assert_eq!(nth_entry, Some(2));
         assert_eq!(ml.unwrap().find("2-entry").unwrap(), 0); //#[allow_ci]
 
         // Request the 3rd entry, which is not available yet, thus we get an empty list
         let (ml, nth_entry, num_entries) =
-            read_measurement_list(&mut ima_ml, tf.path(), 3).unwrap(); //#[allow_ci]
+            read_measurement_list(&mut ima_ml, &mut ima_file, 3).unwrap(); //#[allow_ci]
         assert_eq!(num_entries, Some(3));
         assert_eq!(nth_entry, Some(3));
         assert_eq!(ml.unwrap().len(), 0); //#[allow_ci]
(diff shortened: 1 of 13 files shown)


src/secure_mount.rs
```diff
@@ -18,36 +22,45 @@ use std::process::Command;
...
-        let tokens: Vec<&str> = line.split(' ').collect();
-
-        if tokens.len() < 3 {
-            continue;
+    let f = fs::File::open(MOUNTINFO)?;
+    let f = BufReader::new(f);
+    let lines = f.lines();
+
+    for line in lines.flatten() {
+        if !line.contains('-') {
```
ueno, line 30: Perhaps this check could be moved after splitting the fields, as filesystem type and volume name may also contain `-`.

src/main.rs
```diff
@@ -131,6 +131,9 @@ pub(crate) fn setup_unzipped(
     config: &KeylimeConfig,
 ) -> Result<(PathBuf, PathBuf, PathBuf)> {
     let work_dir = Path::new(&config.work_dir);
+
+    // This call will simply return the Result containing the path since the directory should
+    // already be mounted
```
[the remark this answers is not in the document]
ansasaki: Yes, I was undecided about the approach to take where `config` is available, which allows to easily reconstruct the paths, but it is better to have consistency. I'll change it to pass the paths.

src/permissions.rs
```diff
@@ -0,0 +1,122 @@
...
+
+pub(crate) fn get_uid() -> uid_t {
+    unsafe { libc::getuid() }
+}
+
+pub(crate) fn get_euid() -> uid_t {
+    unsafe { libc::geteuid() }
+}
+
+fn get_uid_gid_from_name(user_group: &str) -> Result<(uid_t, gid_t)> {
```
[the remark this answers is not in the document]
ansasaki: Thanks for the suggestion, I'll try.

src/secure_mount.rs
ansasaki: Oops, thanks for that, I wasn't paying attention. I changed the position (comparing to the original python implementation) to avoid splitting the string when it is not really parseable.

I think that checking for `-` with surrounding spaces is basically equivalent to check for `-` string after splitting on spaces. I'll try this.

ansasaki: @ueno In the end I removed the initial "validation" as it wasn't really useful (let me know if you disagree). I changed the code to error in case the `-` separator is not found while parsing the lines.

src/keys_handler.rs
```diff
@@ -280,13 +280,22 @@ mod tests {
...
     async fn test_u_or_v_key(key_len: usize, payload: Option<&[u8]>) {
         let test_config = KeylimeConfig::default();
-        let quotedata = web::Data::new(QuoteData::fixture().unwrap()); //#[allow_ci]
+        let mut fixture = QuoteData::fixture().unwrap(); //#[allow_ci]
+
+        // Create temporary working directory and secure mount
+        let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
+        fixture.secure_mount =
+            PathBuf::from(&temp_workdir.path().join("tmpfs-dev"));
+        fs::create_dir(&fixture.secure_mount).unwrap(); //#[allow_ci]
```
ansasaki, line 295: I added this because when the secure mount path is passed directly to `run_encrypted_payload()`, the directory is expect to exist (since the secure mount should be ready from the beginning).

In the previous implementation the `secure_mount` would be created by `secure_mount::mount()`. It took me some time to understand what was happening to make the test to fail.

src/ima.rs
```diff
@@ -57,14 +58,16 @@ impl ImaMeasurementList {
 /// was read and the current number of entries in the file.
 pub(crate) fn read_measurement_list(
     ima_ml: &mut ImaMeasurementList,
-    filename: &Path,
+    ima_file: &Option<Arc<Mutex<File>>>,
```
ueno, line 61: Could you explain why `Mutex` is necessary here? Is it possible that the file is accessed from multiple threads?

ueno: Yeah, I understand that, but in that case I wonder if it might make more sense to handle locking in `quotes_handler.rs` (and if it's not feasible at least have some comment in `ima.rs` why this interface involves locking)?

ansasaki: > Yeah, I understand that, but in that case I wonder if it might make more sense to handle locking in `quotes_handler.rs` (and if it's not feasible at least have some comment in `ima.rs` why this interface involves locking)?

I have to take some time and think if there is a better way and if your suggestion is feasible. Right now I can't see another way to open a file before dropping the privileges and later share the handle in multiple threads without using a mutex on the handle. That would be possible if I can clone the handle in each thread and operate on them independently. The content of the file is accessed only for reading, but the handle is modified by seek operations. I need to test this.

ueno: > Right now I can't see another way to open a file before dropping the privileges and later share the handle in multiple threads without using a mutex on the handle.

I'm not saying that using a mutex is a problem, but suggest calling `.lock()` in `integrity` (in quotes_handler.rs) rather than `read_measurement_list` (in ima.rs). The lock for the other argument (`ima_ml`) is already implemented that [way](https://github.com/keylime/rust-keylime/pull/364/files#diff-e5477179c2c236993dfa1103291d887c44225ca86f91702cba389518fe2eaca4L242).

Here is my [attempt](https://github.com/ueno/rust-keylime/commit/1b0bbabec6423af2f8cc28923c9cba50d6553523) (totally untested).

> That would be possible if I can clone the handle in each thread and operate on them independently.

I am not sure why this mechanism is necessary. Maybe I am missing something but afaics the only code path that could access the file object is `integrity` → `read_measurement_list`, which is even protected by `ima_ml.lock()` so there is only one thread that could take this path.

> The content of the file is accessed only for reading, but the handle is modified by seek operations.

That makes me think that the handle could be generalized as `Read + Seek`, but maybe overkill :-)

dist/systemd/system/installer.sh
```diff
@@ -0,0 +1,57 @@
...
+    echo "Unable to find keylime agent" 1>&2
+    exit 1
+fi
+
+echo "Using keylime directory: ${KEYLIMEDIR}"
+
+# prepare keylime service files and store them in systemd path
+sed "s|KEYLIMEDIR|$KEYLIMEDIR|g" $BASEDIR/keylime_agent.service.template > /etc/systemd/system/keylime_agent.service
+
+# Copy secure mount
```
ueno, line 25: Why not extend the existing [GNUmakefile](https://github.com/keylime/rust-keylime/blob/master/GNUmakefile) and use `install` rather than manual calls to `chmod` and `chown`?

ansasaki: Because I did it without thinking much :)
I will make the changes in the Makefile.

src/ima.rs
ansasaki: Ok, I think I understand now what you were asking. I'll try you approach and move the `.lock()`.

ansasaki: Modified as suggested. I followed the same approach for the measured boot log file.

dist/systemd/system/installer.sh
ansasaki: Removed the installer script completely and added minimal changes to the `GNUmakefile`.

ueno (approved):
Thanks for the update; looks much simpler.

src/ima.rs
```diff
@@ -7,6 +7,7 @@ use std::{
     fs::File,
     io::{prelude::*, Error, SeekFrom},
     path::Path,
+    sync::{Arc, Mutex},
```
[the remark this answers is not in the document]
ansasaki: Right, I'll drop that

ansasaki:
> @ansasaki can we merge this, or do you plan any further update?

I think we can merge, I don't plan to add more changes here.

PR #376, opened 2022-05-16
ima_emulator: Support PCR hash algorithms other than SHA-1

ueno:
This ports the recent changes to keylime_ima_emulator in the Python
stack, which allows PCR hash algorithms other than SHA-1 through the
command-line options.


ueno:
I've only tried this with `--hash_algs sha256` and it seems to work; we probably should add deterministic unit tests with fixtures.

THS-on:
Note that the hashes of ToMToU errors need to be handled differently.

https://github.com/keylime/keylime/blob/37553151b204ace6a1b672b017486f2ed9b17ff4/keylime/ima/ast.py#L371-L376

ueno:
I think the logic is [here](https://github.com/keylime/rust-keylime/pull/376/files#diff-c5136e0a6dc3c65b67d122a69621f5b9057833a037c277995acf9dcb8533fc7fR240-R242); I'll add a comment to be clear.

THS-on:
> I think the logic is [here](https://github.com/keylime/rust-keylime/pull/376/files#diff-c5136e0a6dc3c65b67d122a69621f5b9057833a037c277995acf9dcb8533fc7fR240-R242); I'll add a comment to be clear.

Yes, I missed that.

kkaarreell:
Parameters that we are currently using in tests for Python IMA emulator are `--hash_algs sha256 --ima-hash-alg sha256`.

ueno:
@kkaarreell thanks Karel, I've cherry-picked the commit. The CI is currently failing at:
https://github.com/keylime/rust-keylime/pull/376/files#diff-c5136e0a6dc3c65b67d122a69621f5b9057833a037c277995acf9dcb8533fc7fR87

This code does not exist in the Python IMA emulator, but it checks sanity that the calculated hash value match the template hash in the log file. However, it never matches if `--ima-hash-alg=sha256` is given, because the hash value in the template is hard-coded to be SHA-1:
https://elixir.bootlin.com/linux/v5.18/source/security/integrity/ima/ima_fs.c#L236

That could also mean that, with that option the ToMToU measure will never trigger:
https://github.com/keylime/rust-keylime/pull/376/files#diff-c5136e0a6dc3c65b67d122a69621f5b9057833a037c277995acf9dcb8533fc7fR80

Therefore I suggest using `--hash_algs=sha256 --ima-hash-alg=sha1`.

kkaarreell:
Hi @ueno , I have change it in https://github.com/RedHat-SP-Security/keylime-tests/pull/142
Please rerun packit tests.

Issue #368, opened 2022-05-10
Agent fails to start if `shim.py` is not found


maugustosilva:
# Environment

OS: FC35
`Verifier`/`Registrar`: v6.4.0
`Agent`: b869a6a(master).

# Symptom

```
 WARN  keylime_agent > INSECURE: Keylime is using a software TPM emulator rather than a real hardware TPM.
 WARN  keylime_agent > INSECURE: The security of Keylime is NOT linked to a hardware root of trust.
 WARN  keylime_agent > INSECURE: Only use Keylime in this mode for testing or debugging purposes.
 INFO  keylime_agent > Starting server with API version v2.0...
 ERROR keylime_agent > Could not find python shim at /usr/libexec/keylime/shim.py
Error: Configuration("Could not find python shim at /usr/libexec/keylime/shim.py")
```

# Workaround

```
mkdir -p /usr/libexec/keylime
touch /usr/libexec/keylime/shim.py
```

Perhaps a more operations-friendly option, where the agent can start (albeit with reduced capabilities) could be considered.

lkatalin:
> I propose we make the support for python revocation actions optional, enabled by default for now, and we can remove the support completely later. The idea is to use a mechanism similar to the one used to remove ZeroMQ dependency.

I think this would be a good way to do it.

maugustosilva:
Allow me to suggest an additional refinement to your proposal @ansasaki: In case `listen_notifications = False` under `[cloud_agent]`, then do **not** return error on initialization if `shimp.py` is not in place

ansasaki:
> Ah, I see... notifications via REST API are possible, ok. Being that the case, you're correct: provide a flag on the `[cloud_agent]` section of keylime configuration file to **disable** "python revocation actions", in which case the agent does not look for `shim.py`.

The PR I proposed (#377) adds the  `legacy-python-actions` feature, which can be disabled in compile time to drop the related code, therefore is not a configuration option that can be enabled/disabled later. The plan is to deprecate this feature and ultimately remove it completely when the rust agent is the official agent.

@maugustosilva Do you think it would be necessary to make it a configuration option which can be enabled/disabled?

maugustosilva:
Not at all... since it will be deprecated soon enough, I agree, let's leave it as compile option only.

PR #377, opened 2022-05-16
Make support for legacy python revocation actions optional

ansasaki:
Add the 'legacy-python-actions' feature, enabled by default, to control
whether the agent supports python revocation actions loaded as modules.

Note that this is the only kind of revocation actions script supported
by the python agent implementation, which means that removing this
feature will make existing revocation actions unusable.

Disabling the 'legacy-python-actions' feature will not disable the
support for running stand-alone python revocation action scripts, i.e.
executable python scripts can be used as a revocation actions scripts,
as well as any other executable.

Fixes: #368

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -424,15 +424,19 @@ async fn main() -> Result<()> {
         warn!("INSECURE: Only use Keylime in this mode for testing or debugging purposes.");
     }
 
-    // Verify if the python shim is installed in the expected location
-    let python_shim =
-        PathBuf::from(&config.revocation_actions_dir).join("shim.py");
-    if !python_shim.exists() {
-        error!("Could not find python shim at {}", python_shim.display());
-        return Err(Error::Configuration(format!(
-            "Could not find python shim at {}",
-            python_shim.display()
-        )));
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -424,15 +424,17 @@ async fn main() -> Result<()> {
...
-    // Verify if the python shim is installed in the expected location
-    let python_shim =
-        PathBuf::from(&config.revocation_actions_dir).join("shim.py");
-    if !python_shim.exists() {
-        error!("Could not find python shim at {}", python_shim.display());
-        return Err(Error::Configuration(format!(
-            "Could not find python shim at {}",
-            python_shim.display()
-        )));
+    if cfg!(feature = "legacy-python-actions") {
```
ueno, line 427: What about using [`cfg_if!`](https://docs.rs/cfg-if/1.0.0/cfg_if/) to compile out unnecessary code?

ansasaki: OK, I'll modify the code to use `cfg_if!`

lkatalin (approved):
:+1:

Issue #380, opened 2022-05-18
keylime agent supplies malformed quote to verifier when there is no IMA log in the system


galmasi:
Environment
--

Verifier/Registrar: v6.4.0 standard, running on an Ubuntu 20.04 substrate with docker containers.
Keylime Agent: 
    * rust agent commit c8343943b23a64d9a7b378e1a4bb2251832f76fc (Wed May 11, "CI: Add packit ...")
     * OS: OpenBMC "phosphor" running on a qemu-emulated cortex A7.
     * kernel `5.14.6-yocto-s-dirty-7ae156b` with _IMA disabled_.
     * The problem can be reproduced in any agent where the IMA log is missing/unreadable.

Symptom
--

Keylime verifier fails upon first attestation attempt with the following message:

```
2022-05-18 17:27:23.266 - keylime.cloudverifier - ERROR - ima_measurement_list_entry parameter must be an integer
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/keylime-6.4.0-py3.9.egg/keylime/cloud_verifier_tornado.py", line 810, in invoke_get_quote
    failure = cloud_verifier_common.process_quote_response(agent, json_response['results'], agentAttestState)
  File "/usr/local/lib/python3.9/site-packages/keylime-6.4.0-py3.9.egg/keylime/cloud_verifier_common.py", line 73, in process_quote_response
    raise Exception("ima_measurement_list_entry parameter must be an integer")
Exception: ima_measurement_list_entry parameter must be an integer
```

Root cause
--

The culprit is line 261 in `src/quotes_handler.rs`. This line returns the triplet of ima measurement list, line entry and number of entries in a situation where the IMA list was not found by the agent.

The current implementation returns a "None" in this situation, which translates to `NoneType` on the verifier side. However, the verifier seems to insist on this value *always* being an integer, and throws an uncaught exception when sent a `None` instead.

A trivial fix to the problem on the keylime agent side would be to send replace `None` with `Some(0)` in this situation. Thus:

```
diff --git a/src/quotes_handler.rs b/src/quotes_handler.rs
index 6654882..4206c58 100644
--- a/src/quotes_handler.rs
+++ b/src/quotes_handler.rs
@@ -258,7 +258,7 @@ pub async fn integrity(
                 }
             }
         } else {
-            (None, None, None)
+            (None, Some(0), None)
         };
 
```

This seems to fix the immediate problem. 

Potential PR
--

[...]

PR #381, opened 2022-05-19
Skip serialization of optional fields

ansasaki:
Add the `#[serde(skip_serializing_if = "Option::is_none")]` annotation to
all optional fields serialized using `serde`.  This makes `serde` to skip
serializing optional fields when the value is `None`.

Fixes: #380

### DIFF
--- a/src/keys_handler.rs
+++ b/src/keys_handler.rs
@@ -18,6 +18,7 @@ use std::{convert::TryInto, sync::Arc};
 pub struct KeylimeUKey {
     auth_tag: String,
     encrypted_key: String,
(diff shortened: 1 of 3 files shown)


ansasaki:
@galmasi Hello, could you please check if this solves the issue #380?

galmasi:
Confirming that @ansasaki's commit `6c0a0295b371da82159b916f118ef8e555aef9fe` on the skip_optional branch performs correctly in the qemu-arm development system set up by Angelo Ruocco. That is, in a situation where there is no IMA log to be had, interaction between the rust agent and the keylime server is now correct.

Thanks again for the quick fix.

lkatalin (approved):

PR #384, opened 2022-05-24
quotes_handler: Rewind measured boot log file

ansasaki:
Rewind the file before reading its content.

Fixes: #383 

### DIFF
--- a/src/quotes_handler.rs
+++ b/src/quotes_handler.rs
@@ -221,6 +221,15 @@ pub async fn integrity(
             if let Some(measuredboot_ml_file) = &data.measuredboot_ml_file {
                 let mut ml = Vec::<u8>::new();
                 let mut f = measuredboot_ml_file.lock().unwrap(); //#[allow_ci]
+                if let Err(e) = f.rewind() {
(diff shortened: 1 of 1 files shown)


THS-on:
The measured boot log should never change after boot, so we just could keep it in memory instead of reading it again.

maugustosilva:
I can confirm this single line change indeed fixes the problem. LGTM

ansasaki:
> The measured boot log should never change after boot, so we just could keep it in memory instead of reading it again.

This is probably the best solution. I'll make the changes towards this.

lukehinds:
Just a nit, can the unwrap be error handled instead (I figure rewind is an Option or Result). Perhaps even .except with a meaningful error message.

ansasaki:
After some discussion, it was decided to revert the change to simply add a rewind. There is no clear benefit on holding several KB of data on memory all the time to avoid rewinding and re-reading the file which is on a memory-based file system.

ansasaki:
@lkatalin @lukehinds @ashcrow Could you review/approve this?

lkatalin (approved):

PR #385, opened 2022-05-26
Add support for hash_ek UUID creation

THS-on:
The tss-esapi crate provides now the conversion functions for Public to a more standard key format.
This PR first updates to the newest tss-esapi version and then implements the hash_ek option for UUID creation.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -72,7 +72,7 @@ use std::{
 };
 use tss_esapi::{
     handles::KeyHandle, interface_types::algorithm::AsymmetricAlgorithm,
-    Context,
+    traits::Marshall, Context,
 };
 use uuid::Uuid;
 
@@ -440,8 +440,7 @@ async fn main() -> Result<()> {
     }
 
     // Gather EK values and certs
-    let (ek_handle, ek_cert, ek_tpm2b_pub) =
-        tpm::create_ek(&mut ctx, config.enc_alg.into())?;
+    let ek_result = tpm::create_ek(&mut ctx, config.enc_alg.into())?;
 
     // Try to load persistent TPM data
     let tpm_data = config.tpm_data.clone().and_then(|data|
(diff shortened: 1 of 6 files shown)


THS-on:
@lkatalin @ansasaki this PR is ready for review.

src/main.rs
```diff
@@ -440,8 +440,8 @@ async fn main() -> Result<()> {
     }
 
     // Gather EK values and certs
-    let (ek_handle, ek_cert, ek_tpm2b_pub) =
-        tpm::create_ek(&mut ctx, config.enc_alg.into())?;
+    // let (ek_handle, ek_cert, ek_tpm2b_pub) =
```
lkatalin, line 443: Is this line meant to stay?

lkatalin (approved):

src/main.rs
THS-on: No, thanks for catching it.

src/tpm.rs
```diff
@@ -83,6 +83,15 @@ pub(crate) fn get_tpm2_ctx() -> Result<Context> {
     Context::new(tcti).map_err(|e| e.into())
 }
 
+// Holds the output of create_ek
+#[derive(Clone, Debug)]
+pub struct EKResult {
+    pub key_handle: KeyHandle,
+    pub ek_cert: Option<Vec<u8>>,
+    pub public: tss_esapi::structures::Public,
+    pub tpm2b_pub: Vec<u8>,
```
ueno, line 92: Do we need this member (`tpm2b_pub`), as it could be calculated from `public` on demand?

THS-on: Yes. I even think that we can remove the pub_to_vec function because newer tss versions have a marschall function. See: https://docs.rs/tss-esapi/7.0.1/tss_esapi/structures/enum.Public.html#method.marshall

THS-on: Done.

ueno (approved):
Looks good to me; thanks for the clean up!

Issue #382, opened 2022-05-20
Persist all required keys and certificates to disk


THS-on:
The agent currently only persists the `TpmsContext` of the AK to disk and not the NK and mTLS certificate.
This means that restarting the agent (not the whole system) mid attestation will cause an unnecessary failure.

The following things are required to fix this:

- [ ] Persisting the NK private and public key to disk instead of generating a new one at every startup
- [ ] Persisting the mTLS certificate to disk
- [ ] (Persisting the AK `Private` and `Public` to disk instead of  `TpmsContext` because the context not loadable after a reboot. There is no reason why we need a fresh AK every time.)

PR #391, opened 2022-06-07
persist AK, NK and mTLS certificate to disk

THS-on:
A new file in /var/lib/keylime is created called "agent_data.json which
holds all those keys and certificates. This allows the agent to be
restarted mid attestation without failing due to wrong keys.

Fixes: #382

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ serde_derive = "1.0.80"
 serde_json = { version = "1.0", features = ["raw_value"] }
 static_assertions = "1"
 tempfile = "3.0.4"
-tokio = {version = "1", features = ["full"]}
+tokio = {version = "1.13.1", features = ["full"]}
 tss-esapi = "7.1.0"
 thiserror = "1.0"
 uuid = {version = "0.8", features = ["v4"]}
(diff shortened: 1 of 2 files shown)


THS-on:
> I went back to read your https://github.com/keylime/rust-keylime/pull/293#issuecomment-1007698430 on https://github.com/keylime/rust-keylime/pull/293 about loading the AK from disk, I assume that's applicable here as well?

Yes the `Private` struct returned by `create_ak(..)` is symmetricly encrypted from the TPM.

> I assume that encryption operation was happening with the https://github.com/keylime/rust-keylime/pull/391#diff-bd3201dff7815da0652c4c02d8d1e9f27f63de6cf5c1d332c2f21ff995ce933cL207 function before (could be wrong here) and I don't see that being called anymore in this change. Is that happening somewhere else? Do we not need it?

The old implementation first dumped and then serialized the context with the AK, the new one directly serializes the `Private` struct (TPM2_PRIVATE) which has the advantage that the AK can be loaded after a TPM reset (this is not possible with context objects). 
The encryption happens on AK key creation.

What has changed is that we also dump the NK to disk, which is fine because it is used for transport security and the payload mechanism, not the actual attestation. The worst case that happens if someone gets this key (by having read access to /var/lib/keylime) is that they can decrypt payloads which the agent anyway does and stores under the secure tmp mount with the same permissions.

lkatalin:
@THS-on Thanks for the explanation.

PR #395, opened 2022-06-09
Drop unused rustc-serialize and flate2 dependencies and make wiremock optional

ansasaki:
Drop unused `rustc-serialize` and `flate2` dependencies and make `wiremock` an optional dependency used when the `testing` feature is enabled.


aplanas (approved):

Cargo.toml
```diff
@@ -45,16 +43,16 @@ tss-esapi = "7.1.0"
 thiserror = "1.0"
 uuid = {version = "0.8", features = ["v4"]}
 zmq = {version = "0.9.2", optional = true}
+wiremock = {version = "0.5", optional = true}
```
ueno, line 46: I'd leave a comment why "wiremock" is now a regular dependency, while it is only required for testing (I suppose the reason is that cargo doesn't yet support optional dev-dependencies, but without a comment it's a bit confusing).

aplanas: that is a good point, the documentation (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies) suggest that both are equivalent, and should be supported

ueno: I came across https://github.com/rust-lang/cargo/issues/1596 which might be relevant.

ansasaki: > I'd leave a comment why "wiremock" is now a regular dependency, while it is only required for testing (I suppose the reason is that cargo doesn't yet support optional dev-dependencies, but without a comment it's a bit confusing).

Yes, it is because optional dev-dependencies are not supported. I'll add a comment mentioning https://github.com/rust-lang/cargo/issues/1596.

PR #409, opened 2022-06-17
Handle whitespace in keylime.conf

### LINKED ISSUE #408: Whitespace is not handled for keylime.conf in config_get()
In the function [`config_get()`](https://github.com/keylime/rust-keylime/blob/b834667b4d775065be3d7677e8cb6ad209c43668/src/common.rs#L604), the condition where a key in keylime.conf is present, but the value is whitespace, such as:
```
key:
```
...is not handled. As a result, if a user runs the Keylime agent as root, but does not specify a value for the key `run_as` in keylime.conf, they will get the following error on agent startup:
```
 ERROR keylime_agent::permissions > Invalid parameter format:  cannot be parsed as 'user:group'
 ERROR keylime_agent::permissions > Invalid parameter format:  cannot be parsed as 'user:group'
```
This error arises because, in the function [`try_from()`](https://github.com/keylime/rust-keylime/blob/b834667b4d775065be3d7677e8cb6ad209c43668/src/permissions.rs#L41), a check is performed to ensure that the `run_as` value conforms to the expected format of `user:group`. Since whitespace does not pass this conditional check, the user gets an error that would be appropriate if they had entered a non-conforming value, such as `usergroup`, but which is not appropriate if no value is specified for `run_as` at all.

The expected behavior is that the agent starts successfully if the `run_as` key is not present or not specified in keylime.conf. Instead, the agent bails if `run_as` is present, but left blank.

#409 is a proposed fix.

greyspectrum:
Signed-off-by: greyspectrum <hello@greyspectrum.io>

Closes #408

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -459,11 +459,19 @@ impl KeylimeConfig {
             Ok(s) => bool::from_str(&s.to_lowercase())?,
             Err(_) => ALLOW_PAYLOAD_REV_ACTIONS,
         };
+
         let run_as = if permissions::get_euid() == 0 {
             match config_get(&conf_name, &conf, "cloud_agent", "run_as") {
-                Ok(user_group) => Some(user_group),
+                Ok(user_group) => {
+                    if user_group.is_empty() {
+                        warn!("Cannot drop privileges since 'run_as' is empty in 'cloud_agent' section of keylime.conf.");
+                        None
+                    } else {
+                        Some(user_group)
+                    }
+                }
                 Err(_) => {
-                    warn!("Cannot drop privileges since 'run_as' is empty or missing in 'cloud_agent' section of keylime.conf.");
+                    warn!("Cannot drop privileges since 'run_as' is missing in 'cloud_agent' section of keylime.conf.");
                     None
                 }
             }
@@ -698,7 +706,7 @@ fn config_get(
         }
     };
     let value = match section.get(key) {
-        Some(value) => value,
+        Some(value) => value.trim(),
         None =>
         // TODO: Make Error::Configuration an alternative with data instead of string
         {
@@ -707,6 +715,10 @@ fn config_get(
                 key, conf_name
             )))
         }
+    };
+
+    if value.is_empty() {
+        warn!("Cannot find value for key {} in file {}", key, conf_name);
     };
 
     Ok(value.to_string())
--- a/src/main.rs
+++ b/src/main.rs
@@ -408,7 +408,12 @@ async fn main() -> Result<()> {
     // Drop privileges
     if let Some(user_group) = &config.run_as {
         permissions::chown(user_group, &mount);
-        permissions::run_as(user_group);
+        if let Err(e) = permissions::run_as(user_group) {
+            let message = "The user running the Keylime agent should be set in keylime.conf, using the parameter `run_as`, with the format `user:group`".to_string();
+
+            error!("Configuration error: {}", &message);
+            return Err(Error::Configuration(message));
+        }
     }
 
     info!("Starting server with API version {}...", API_VERSION);


lkatalin:
This seems like a valid error and fix to me insofar as Keylime should not be trying to switch into a `user:group` that is left unspecified. In general it seems like whitespace in the config file should not be read in or treated as if it should be a legitimate value.

@ansasaki I think the error with testing here results from the expectation that setting `payload_script` to empty should be okay ([here](https://github.com/RedHat-SP-Security/keylime-tests/blob/e5e81c634c0221259bfa3aab5afa1a720e5f6592/functional/basic-attestation-without-mtls/main.fmf#L9)), but with this new change that counts as an error. Maybe the test could be updated? I don't know my way around the Packit tests very well so I have been trying to check whether this is the issue but I'm not certain. What do you think is best?

ansasaki:
> This seems like a valid error and fix to me insofar as Keylime should not be trying to switch into a `user:group` that is left unspecified. In general it seems like whitespace in the config file should not be read in or treated as if it should be a legitimate value.

I agree that there is an issue in `run_as` option handling of whitespace, but I disagree that whitespace is invalid for all options. The default configuration file contains whitespaces/empty value in many options, usually with the semantics that it should use the default value. With this change we are effectively changing the configuration file semantics, making required to have a non-whitespace value for all options.

> 
> @ansasaki I think the error with testing here results from the expectation that setting `payload_script` to empty should be okay ([here](https://github.com/RedHat-SP-Security/keylime-tests/blob/e5e81c634c0221259bfa3aab5afa1a720e5f6592/functional/basic-attestation-without-mtls/main.fmf#L9)), but with this new change that counts as an error. Maybe the test could be updated? I don't know my way around the Packit tests very well so I have been trying to check whether this is the issue but I'm not certain. What do you think is best?

The best for me would be to fix the handling of whitespace/empty value instead of making it an invalid value for all options. I would make the semantics that whitespace/empty value means to fallback to the default value for all options (but add a warning message that an empty/whitespace value was found)

lkatalin (approved):
LGTM

lkatalin:
@ansasaki Is this along the lines of what you were thinking?

ansasaki:
> @ansasaki Is this along the lines of what you were thinking?

Couldn't we modify `config_get()` to use `trim()` to remove any bogus white spaces instead of explicitly checking if all chars in the string are white spaces? Then the existing check for empty string would've caught this. Moreover, I think cleaning the white spaces from the strings obtained from the configuration should be appropriate for all options.

lkatalin:
> Then the existing check for empty string would've caught this.

@ansasaki Which check is this? I've tried this change and `trim()` on its own does not fix the error because the config values are still being assigned this way (debug print statement inserted):

```
 INFO  keylime_agent::common > value for key keylime_ca is: default
 INFO  keylime_agent::common > value for key revocation_actions is: 
 INFO  keylime_agent::common > value for key run_as is: 

```

So the code will still call [`run_as`](https://github.com/keylime/rust-keylime/blob/ad8608fd93ad026fe3ebbbecdd12a70b6e5dcf1a/src/main.rs#L411) on the whitespace (which seems like it should not happen) and give the unhelpful error below:

```
ERROR keylime_agent::permissions  > Invalid parameter format:  cannot be parsed as 'user:group'
ERROR keylime_agent::permissions  > Invalid parameter format:  cannot be parsed as 'user:group'
```

Instead the earlier change to check for whitespace before calling `permissions::run_as()`  seemed to fix this.

ansasaki:
> > Then the existing check for empty string would've caught this.
> 
> @ansasaki Which check is this?

I meant this check:
https://github.com/keylime/rust-keylime/blob/5b9072abae219bda0e9a95498b5aef4af5db1bda/src/common.rs#L462-L472

I misread the code and thought it would be caught here, but `config_get()` does not return `Err()` for empty values, only for missing options. 

> I've tried this change and `trim()` on its own does not fix the error because the config values are still being assigned this way (debug print statement inserted):
> 
> ```
>  INFO  keylime_agent::common > value for key keylime_ca is: default
>  INFO  keylime_agent::common > value for key revocation_actions is: 
>  INFO  keylime_agent::common > value for key run_as is: 
> ```
> 
> So the code will still call [`run_as`](https://github.com/keylime/rust-keylime/blob/ad8608fd93ad026fe3ebbbecdd12a70b6e5dcf1a/src/main.rs#L411) on the whitespace (which seems like it should not happen) and give the unhelpful error below:
> 
> ```
> ERROR keylime_agent::permissions  > Invalid parameter format:  cannot be parsed as 'user:group'
> ERROR keylime_agent::permissions  > Invalid parameter format:  cannot be parsed as 'user:group'
> ```
> 
> Instead the earlier change to check for whitespace before calling `permissions::run_as()` seemed to fix this.

Right, `config_get()` doesn't judge if empty is valid or not, so we need to fix both the handling of the empty string in the check I mentioned above and trim the string in `config_get()`. Probably the best is to check `user_group.is_empty()` before returning `Some(user_group)` in the `Ok()` case.

We should also fix the message in the `Err()` case as it is misleading: `Err()` is returned when the option is missing, but not when the option is empty or blank.

lkatalin:
@ansasaki Thanks for checking, makes sense to me.
@greyspectrum Let us know if you have questions.

greyspectrum:
@ansasaki This should look more like what you outlined in your last comment now.

ansasaki (approved):

PR #410, opened 2022-06-21
main: add better error message when mTLS certificate is missing

THS-on:
The error message now also includes what is missing not just:
`IO error: No such file or directory (os error 2)`


THS-on:
The other implementation is cleaner, so I'm closing this.

Issue #407, opened 2022-06-16
TPM quote decoding fails on s390x arch


ansasaki:
When packaging `rust-keylime` on Fedora, the build on s390x machine fails when running the `tpm::quote_encode_decode` test:

```
failures:
---- tpm::quote_encode_decode stdout ----
thread 'tpm::quote_encode_decode' panicked at 'unable to decode quote: Tpm { err: WrapperError(InvalidParam), kind: None, message: "the provided parameter is invalid for that type." }', src/tpm.rs:871:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failures:
    tpm::quote_encode_decode
test result: FAILED. 36 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.80s
     Running `/builddir/build/BUILD/rust-keylime-aed51c7c8c526953e945357594352c3df2ca4ace/target/release/deps/keylime_ima_emulator-3020e777b56dff63`
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
error: test failed, to rerun pass '--bin keylime_agent'
error: Bad exit status from /var/tmp/rpm-tmp.qeHIOg (%check)
RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.qeHIOg (%check)
Child return code was: 1
EXCEPTION: [Error()]
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/mockbuild/trace_decorator.py", line 93, in trace
    result = func(*args, **kw)
  File "/usr/lib/python3.10/site-packages/mockbuild/util.py", line 598, in do_with_status
    raise exception.Error("Command failed: \n # %s\n%s" % (command, output), child.returncode)
```

The culprit seems to be the following `std::mem::transmute` call:
https://github.com/keylime/rust-keylime/blob/25a5ddd2c22ec6201368b0c8c30047f547c60243/src/tpm.rs#L677-L681

Quoting the rust documentation on [transmutes](https://doc.rust-lang.org/stable/nomicon/transmutes.html#transmutes):
> Get out of our way type system! We're going to reinterpret these bits or die trying! Even though this book is all about doing things that are unsafe, I really can't emphasize that you should deeply think about finding Another Way than the operations covered in this section. This is really, truly, the most horribly unsafe thing you can do in Rust. The guardrails here are dental floss.

[...]

PR #411, opened 2022-06-21
tpm: Add serialization functions for structures in quotes

ansasaki:
Add serialization and deserialization functions for the data in quotes
to avoid endianness issues when the arch is big-endian.

The added serialization and deserialization functions will convert the
data endianness as necessary.

Note: the official marshalling and unmarshalling functions cannot be
used directly because the tpm2-tools uses a custom format.

Fixes: #407

### DIFF
--- a/src/tpm.rs
+++ b/src/tpm.rs
@@ -52,15 +52,20 @@ use tss_esapi::{
     tss2_esys::{
         Tss2_MU_TPM2B_PUBLIC_Marshal, Tss2_MU_TPMS_ATTEST_Marshal,
         Tss2_MU_TPMS_ATTEST_Unmarshal, Tss2_MU_TPMT_SIGNATURE_Marshal,
-        TPM2B_ATTEST, TPM2B_PUBLIC, TPML_DIGEST, TPML_PCR_SELECTION,
-        TPMS_ATTEST, TPMS_SCHEME_HASH, TPMT_SIGNATURE, TPMT_SIG_SCHEME,
-        TPMU_SIG_SCHEME,
+        TPM2B_ATTEST, TPM2B_DIGEST, TPM2B_PUBLIC, TPML_DIGEST,
+        TPML_PCR_SELECTION, TPMS_ATTEST, TPMS_PCR_SELECTION,
+        TPMS_SCHEME_HASH, TPMT_SIGNATURE, TPMT_SIG_SCHEME, TPMU_SIG_SCHEME,
     },
     utils::TpmsContext,
     Context,
 };
 
 pub const MAX_NONCE_SIZE: usize = 64;
+pub const TPML_DIGEST_SIZE: usize = std::mem::size_of::<TPML_DIGEST>();
+pub const TPML_PCR_SELECTION_SIZE: usize =
+    std::mem::size_of::<TPML_PCR_SELECTION>();
+pub const TPMS_PCR_SELECTION_SIZE: usize =
+    std::mem::size_of::<TPMS_PCR_SELECTION>();
 
 /*
  * Input: None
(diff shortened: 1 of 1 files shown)


src/tpm.rs
```diff
@@ -123,6 +128,129 @@ pub(crate) fn create_ek(
...
+pub(crate) fn serialize_pcrsel(
+    pcr_selection: &TPML_PCR_SELECTION,
+) -> Vec<u8> {
+    let mut output = Vec::with_capacity(TPML_PCR_SELECTION_SIZE);
+    output.extend(u32::to_le_bytes(pcr_selection.count));
+    for selection in pcr_selection.pcrSelections.iter() {
+        output.extend(selection.hash.to_le_bytes());
+        output.extend(selection.sizeofSelect.to_le_bytes());
+        output.extend(selection.pcrSelect);
+        output.extend([0u8; 1]); // padding to keep the memory aligment
```
ueno, line 143: typo: aligment → alignment

src/tpm.rs
```diff
@@ -123,6 +128,129 @@ pub(crate) fn create_ek(
...
+        return Err(KeylimeError::InvalidRequest);
+    }
+
+    let mut reader = std::io::Cursor::new(pcrsel_vec);
+    let mut count_vec = [0u8; 4];
+
+    let len = reader.read(&mut count_vec)?;
+    if len < count_vec.len() {
+        return Err(KeylimeError::InvalidRequest);
+    }
```
ueno, line 163: Would it make sense to use `read_exact` as it returns dedicated error on unexpected EOF? Something like:
```suggestion
    reader.read_exact(&mut count_vec)?;
```
for other places as well.

src/tpm.rs
```diff
@@ -123,6 +128,129 @@ pub(crate) fn create_ek(
...
+
+    let count = u32::from_le_bytes(count_vec);
+    // Always 1 PCR digest should follow
+    if count != 1 {
+        return Err(KeylimeError::InvalidRequest);
+    }
+
+    let mut pcr_selections: [TPMS_PCR_SELECTION; 16] = [TPMS_PCR_SELECTION {
```
ueno, line 171: `TPMS_PCR_SELECTION` derives from `Default` so I suspect it might be possible to write `TPMS_PCR_SELECTION::default()`?

src/tpm.rs
```diff
@@ -123,6 +128,129 @@ pub(crate) fn create_ek(
...
+    let mut output = Vec::with_capacity(TPML_DIGEST_SIZE);
+    output.extend(u32::to_le_bytes(digest_list.count));
+    for digest in digest_list.digests.iter() {
+        output.extend(digest.size.to_le_bytes());
+        output.extend(digest.buffer);
+    }
+    output
+}
+
+// Deserialize a TPML_PCR_SELECTION from a &[u8] slice.
```
ueno, line 215: typo: TPML_PCR_SELECTION → TPML_DIGEST

src/tpm.rs
```diff
@@ -140,16 +268,14 @@ pub(crate) fn pcrdata_to_vec(
     const DIGEST_SIZE: usize = std::mem::size_of::<TPML_DIGEST>();
 
     let mut pcrsel: TPML_PCR_SELECTION = selection_list.into();
-    pcrsel.count = pcrsel.count.to_le();
-    let pcrsel_vec: [u8; PCRSEL_SIZE] =
-        unsafe { std::mem::transmute(pcrsel) };
+    let pcrsel_vec = serialize_pcrsel(&pcrsel);
 
-    let digest: Vec<TPML_DIGEST> = pcrdata.into();
+    let mut digest: Vec<TPML_DIGEST> = pcrdata.into();
```
ueno, line 273: Is this `mut` needed?

ueno (changes requested):

src/tpm.rs
ansasaki: Nice catch, thanks!

src/tpm.rs
ansasaki: Thanks for the suggestion, I'll replace the reads with `read_exact()`

src/tpm.rs
ansasaki: Thanks!

src/tpm.rs
ansasaki: I'll check, thanks!

ueno (approved):
LGTM!

PR #412, opened 2022-06-22
Show path when fail to load the certificate

aplanas:
When we try to load the mTLS certificate but it is missing, a generic
error is shown: "No such file or directory".

This commit introduces an error! log showing the full path of the
missing certificate.


THS-on:
I implemented something similar in https://github.com/keylime/rust-keylime/pull/410

aplanas:
> I implemented something similar in #410

Ouch! Sorry!

src/main.rs
```diff
@@ -259,7 +259,6 @@ pub(crate) async fn run_encrypted_payload(
     )?;
 
     optional_unzip_payload(&unzipped, config)?;
-
```
ashcrow, line 262: This empty space removal is likely unrelated but I don't see an issue with it :smile:

ashcrow (commented):
Looks good from my POV. I'll defer to @lkatalin for second review and merge.

src/main.rs
aplanas: I think that this comes from "eglot-format", and call "cargo fmt".  Do you want me to remove it?

ashcrow: No, I think it's fine :+1:

lkatalin (approved):
LGTM, thanks!

PR #415, opened 2022-06-24
permissions: Set supplementary groups when dropping privileges

ansasaki:
This allows the agent to run with the user primary group and have the
access given by the supplementary groups.


ansasaki:
@lkatalin @ueno Could you please review/approve this?

src/permissions.rs
```diff
@@ -78,7 +80,12 @@ impl TryFrom<&str> for UserIds {
             )));
         };
 
-        Ok(UserIds { uid, gid })
+        Ok(UserIds {
+            user: user.to_string(),
+            group: group.to_string(),
```
ueno, line 85: I don't see `group` field is used anywhere.

Also just an idea for refactoring: could this struct `UserIds` simply hold `libc::group` and `libc::passwd`?

ueno (approved):

src/permissions.rs
ansasaki: OK, I'll change it

lkatalin:
I don't understand this one enough to approve it, but I am fine if you want to merge since it has two other reviews and passed tests. Thanks for bringing it over from `keylime/keylime`.

PR #417, opened 2022-06-28
main: Use cfg_if to disable ZeroMQ feature code

ansasaki:
Use cfg_if to disable code enabled by 'with_zmq' feature.


ueno (commented):
I wonder what is the benefit of this change: as mentioned in the `cfg_if` [documentation](https://docs.rs/cfg-if/1.0.0/cfg_if/), it is only meaningful when there is a cascade of `#[cfg]` cases.

PR #419, opened 2022-06-29
Load config file less at startup

### LINKED ISSUE #390: Config file is loaded many times at startup
Since `config_get` calls `Ini::load_from_file` every time, the agent opens/closes the keylime.conf 27 times at startup:
```console
$ KEYLIME_CONFIG=$PWD/keylime.conf strace -e trace=openat ./target/debug/keylime_agent |& grep keylime.conf | wc -l
27
```
We probably should pass around the opened `Ini` object to reduce the number of file opening.

avery-blanchard:
This fixes #390 and reduces the number of times the configuration file is loaded during startup. To achieve this, the opened Ini object is passed around rather than being opened with each call to config_get or config_get_env.
@lkatalin

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -636,33 +636,8 @@ fn config_file_get() -> String {
     }
 }
 
-/// Returns revocation ip from keylime.conf if env var not present
-fn revocation_ip_get(conf_name: &String, conf: &Ini) -> Result<String> {
-    config_get_env(
-        conf_name,
-        conf,
-        "general",
-        "receive_revocation_ip",
-        "REVOCATION_IP",
-    )
-}
-
-/// Returns revocation port from keylime.conf if env var not present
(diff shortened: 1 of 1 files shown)


lkatalin:
Sorry I forgot to mention the DCO @avery-blanchard ! You can update your commit(s) to have your sign off with `git commit --amend -s` (this will change the latest commit to have it).

ueno (approved):

lkatalin:
Thanks @avery-blanchard ! Pretty much LGTM.

One nit: In your last commit (removing the two extra functions), you're also including some changes to make `&String` into `&str`. Just for clarity I would either amend your commit message to describe that as well, or alternatively edit your earlier commit to already include this change so that it doesn't show up in the later commit.

lkatalin (approved):

Issue #420, opened 2022-06-29
Cargo command in makefile does not interact well with permissions


lkatalin:
Running `cargo` with `sudo` is not recommended and does not work by default, and running `make install` without `sudo` can result in an error if the current user does not have the right permissions for `/usr/bin/` (which is the case for non-root users). This could frustrate a user trying to use the makefile.

```
[f36-keylime@localhost-live rust-keylime]$ make install
cargo build --target-dir="target" 
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
install -D -t /usr/bin "target/debug/keylime_agent"
install: cannot create regular file '/usr/bin/keylime_agent': Permission denied
make: *** [GNUmakefile:25: install] Error 1

[f36-keylime@localhost-live rust-keylime]$ sudo make install
cargo build --target-dir="target" 
/bin/sh: line 1: cargo: command not found
make: *** [GNUmakefile:21: build] Error 127
```

ueno:
As suggested in README.md, the program building and installation are usually done in two steps, first with "make" as a normal user, and then "make install" as root. This currently doesn't work because of missing dependency declarations. I've filed #421.

PR #421, opened 2022-06-30
GNUmakefile: Make target dependencies explicit

ueno:
Previously, at "make install" the programs are always rebuilt with
cargo, even if they already exist as a result of prior invocation of
"make", which causes a problem if cargo is installed locally for the
user with rustup.  This adds explicit prerequisites to the build
target to avoid that.

Fixes: #420


lkatalin (approved):
Can confirm this seems to fix the error. Thanks @ueno !

PR #422, opened 2022-06-30
keys_handler: Use scopes to drop mutexes before await

ansasaki:
The MutexGuard is not async aware, meaning the mutexes should be
unlocked before any await call.  This change uses scopes to drop the
locked mutexes before calling await.


lkatalin:
Interesting - how was this working before, without the scopes? Was it just keeping the mutex locked indefinitely?

lkatalin (approved):

PR #423, opened 2022-07-12
Add run_as in keylime.conf and avoid executing as root

aplanas:
Add run_as in keylime.conf

If `run_as` parameter is set but the user is missing in the system,
keylime will log an ERROR when trying to drop privileges, but continue
the execution as the current user (usually `root`).  This can be a
security issue, as the agent is running "silently" as a privileged user.
    
This commit stop the execution if an error is found when dropping
privileges for the agent service, and present an `info!` message with
the current user and group.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -411,13 +411,14 @@ async fn main() -> Result<()> {
 
     // Drop privileges
     if let Some(user_group) = &config.run_as {
-        permissions::chown(user_group, &mount);
+        permissions::chown(user_group, &mount)?;
         if let Err(e) = permissions::run_as(user_group) {
             let message = "The user running the Keylime agent should be set in keylime-agent.conf, using the parameter `run_as`, with the format `user:group`".to_string();
 
             error!("Configuration error: {}", &message);
             return Err(Error::Configuration(message));
         }
+        info!("Running the service as {}...", user_group);
     }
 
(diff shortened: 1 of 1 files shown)


lkatalin (approved):
This seems good to me and complementary to #409

lkatalin:
@aplanas Could you resolve the conflict?

aplanas:
@lkatalin, should be done! Do you want me to change the `.expect()` from the `chown` to a `return Err()`? Somehow I see this way clearer and more direct, but I am not sure.

src/main.rs
```diff
@@ -411,13 +411,15 @@ async fn main() -> Result<()> {
 
     // Drop privileges
     if let Some(user_group) = &config.run_as {
-        permissions::chown(user_group, &mount);
+        permissions::chown(user_group, &mount)
+            .expect("Error when changing directory ownership");
```
lkatalin, line 415:
```suggestion
        permissions::chown(user_group, &mount)?;
```
In this case it may be slightly clearer to just return the error from `permissions::chown` directly, as it will print out the reason it failed and the path.

aplanas: Yes, done!

lkatalin:
Thank you, lgtm! Needs one more rebase and it should be good to go.

aplanas:
@lkatalin Rebased, but I did not found any new conflict.

PR #426, opened 2022-07-14
tpm: Respect ek_handle from config

ueno:
Previously it always generated new EK handle even if the "ek_handle"
option is set to use the existing key.

### DIFF
--- a/packit-ci.fmf
+++ b/packit-ci.fmf
@@ -34,6 +34,8 @@
        - /functional/basic-attestation-with-ima-signatures
        - /functional/basic-attestation-without-mtls
        - /functional/basic-attestation-with-unpriviledged-agent
+       - /functional/ek-cert-use-ek_check_script
+       - /functional/ek-cert-use-ek_handle-custom-ca_certs
        - /functional/install-rpm-with-ima-signature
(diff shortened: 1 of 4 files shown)


kkaarreell:
Test log says: `Error: NumParse(ParseIntError { kind: InvalidDigit })`

lkatalin:
@ueno I see some TPM password and owner auth related code - we're not trying to take ownership of the TPM here, right?

ueno:
> @ueno I see some TPM password and owner auth related code - we're not trying to take ownership of the TPM here, right?

Yeah password seems to be needed to access pre-existing EK. For the general decision about ownership, I've filed #429 to make it clear.

lkatalin:
Thanks @ueno. The password seems okay to access the EK as long as the agent doesn't take ownership of the TPM.

kkaarreell:
Hi @ueno @ansasaki @lkatalin 
I think we should also add and document `tpm_ownerpassword` to `keylime.conf` (resp. `keylime-agent.conf`).

lkatalin:
@kkaarreell I opened https://github.com/keylime/rust-keylime/issues/433

PR #443, opened 2022-09-01
Appease clippy

lkatalin:

### DIFF
--- a/src/ima_entry.rs
+++ b/src/ima_entry.rs
@@ -20,7 +20,7 @@ pub trait EncodeLegacy {
     fn encode_legacy(&self, writer: &mut dyn Write) -> Result<()>;
 }
 
-#[derive(PartialEq)]
+#[derive(PartialEq, Eq)]
 pub struct Digest {
     pub algorithm: HashAlgorithm,
     value: Vec<u8>,
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -64,12 +64,10 @@ pub(crate) fn pkey_pub_from_priv(
             .map_err(Error::Crypto)?;
             PKey::from_rsa(rsa).map_err(Error::Crypto)
         }
-        id => {
-            return Err(Error::Other(format!(
-                "pkey_pub_from_priv not yet implemented for key type {:?}",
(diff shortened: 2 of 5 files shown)


src/ima_entry.rs
```diff
@@ -20,7 +20,7 @@ pub trait EncodeLegacy {
     fn encode_legacy(&self, writer: &mut dyn Write) -> Result<()>;
 }
 
-#[derive(PartialEq)]
+#[derive(PartialEq, Eq)]
```
ueno, line 23: Is `PartialEq` still needed, given `Eq` is defined as: `trait Eq: PartialEq`?

ueno (approved):
Looks good to me.

src/ima_entry.rs
lukehinds: I have seen clippy complain about this a few times (even when it breaks code), if you prefer to shut it down this is the ignore option:

https://github.com/sigstore/sigstore-rs/pull/102/files

lkatalin: @ueno I could update this to just derive `Eq`

lkatalin: Looks like the compiler does not like it:
```
[lily@localhost rust-keylime]$ cargo build
   Compiling keylime_agent v0.1.0 (/home/lily/Repos/rust-keylime)
error[E0277]: can't compare `ima_entry::Digest` with `ima_entry::Digest`
  --> src/ima_entry.rs:23:10
   |
23 | #[derive(Eq)]
   |          ^^ no implementation for `ima_entry::Digest == ima_entry::Digest`
   |
   = help: the trait `PartialEq` is not implemented for `ima_entry::Digest`
note: required by a bound in `Eq`
   = note: this error originates in the derive macro `Eq` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider annotating `ima_entry::Digest` with `#[derive(PartialEq)]`
   |
24 | #[derive(PartialEq)]
   |

```

ueno: Thanks for trying; let's keep it as it is.

PR #444, opened 2022-09-07
Add cargo audit to test workflow

avery-blanchard:
Added `cargo audit` to the rust CI workflow. 
There are other options for automatic security analysis, so please let me know your thoughts!
Another option would be to use the [rust audit-check action](https://github.com/marketplace/actions/rust-audit-check) on GitHub. 
Also, this [dependency reviewer action](https://github.com/actions/dependency-review-action) could be used. 
This is related to #94 
@lkatalin

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,6 +20,8 @@ jobs:
       run: cargo fmt --all -- --check
     - name: Check for panics
       run: ./tests/nopanic.ci
+    - name: Run cargo audit
+      run: cargo audit
 
   tests:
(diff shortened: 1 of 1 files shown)


lkatalin (approved):
LGTM, would love thoughts from others though in case it matters which way we implement this.

lkatalin:
@ansasaki @THS-on or anyone else, any thoughts on this, or are we good to merge it?

THS-on:
LGTM. I think we will evaluate the other tools over time and add them if they are useful.

PR #448, opened 2022-09-15
Use the tpm_ownerpassword when generating the EK

ansasaki:
Set the `tpm_ownerpassword` in the TPM context to be used for the Endorsement Hierarchy regardless of the value set in `ek_handle`.

This makes the `tpm_ownerpassword` set in the configuration file to be used to access the TPM when generating the EK handle.

Note: This doesn't change the TPM owner password, only provide it when creating a password session.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -448,17 +448,19 @@ async fn main() -> Result<()> {
         }
     }
 
-    // When the EK handle is given, set auth for the Owner and
-    // Endorsement hierarchies.  Note in the Python implementation,
-    // tpm_ownerpassword option is also used for claiming ownership of
-    // TPM access, which will not be implemented here.
-    if config.ek_handle.is_some() {
-        if let Some(ref v) = config.tpm_ownerpassword {
-            let auth = Auth::try_from(v.as_bytes())?;
-            ctx.tr_set_auth(Hierarchy::Owner.into(), auth.clone())?;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -448,17 +448,14 @@ async fn main() -> Result<()> {
...
-            ctx.tr_set_auth(Hierarchy::Endorsement.into(), auth)?;
-        }
-    }
+    // When the tpm_ownerpassword is given, set auth for the Endorsement hierarchy.
+    // Note in the Python implementation, tpm_ownerpassword option is also used for claiming
+    // ownership of TPM access, which will not be implemented here.
+    if let Some(ref v) = config.tpm_ownerpassword {
+        let auth = Auth::try_from(v.as_bytes())?;
+        ctx.tr_set_auth(Hierarchy::Endorsement.into(), auth)
+            .expect("Failed to set auth for Endorsement Hierarchy");
```
ueno, line 457: Wouldn't it be better to return error, rather than panic?

ueno (approved):

src/main.rs
ansasaki: I will change it to return error.

lkatalin:
> Note: This doesn't change the TPM owner password, only provide it when creating a password session.

Thank you for the updated note.

lkatalin (approved):
LGTM since it does not set the password

PR #449, opened 2022-09-19
Improve configuration and change format to TOML

ansasaki:
This introduces several improvements to the agent configuration.
    
- The file format is modified from INI style files to TOML formatted
  files.
- The configuration options names were modified to follow the
  enhancement proposal:
  - https://github.com/keylime/enhancements/blob/master/72_config_and_simplify_tls.md
- The configuration is constructed using an hierarchical organization
  following the order:
  - The default hardcoded values are loaded
  - If a configuration file is found in /usr/etc/keylime/agent.conf, it
    is loaded and any option set overrides the default values
  - Load configuration snippets from /usr/etc/keylime/agent.conf.d/*,
    overriding the previous values
  - Load the configuration from /etc/keylime/agent.conf, overriding
    previous values
  - Load configuration snippets from /etc/keylime/agent.conf.d/*,
    overriding previous values
  - Finally any configuration option can be overriden via environment
    variable, by using the 'KEYLIME_AGENT_' prefix. For example,
    KEYLIME_AGENT_PORT=9003 will override the 'port' option from the
    'agent' section with the value '9003'.

It is possible to override all file sources with a single file by setting the `KEYLIME_AGENT_CONFIG` environment variable. If it is set, the configuration files and snippets are ignored and all options are loaded from the file pointed by `KEYLIME_AGENT_CONFIG`. The environment variable overrides still apply, meaning it is still possible to override the options from the file pointed by `KEYLIME_AGENT_CONFIG` using the environment variables with the `KEYLIME_AGENT_` prefix.

This was implemented using new dependency crates: `config` and `glob`.

Other changes:
 - Added support to using existing key (optionally protected by a password) and certificate
 - Add the EK public key hash to the stored TPM data (AK) to allow verifying if the TPM where the data is being loaded is the same that generated the data (this is to avoid errors as it is not possible to load the data generated by a TPM using another TPM)


ueno:
> If a configuration file is found in /usr/etc/keylime-agent.conf, it
is loaded and any option set overrides the default values

Perhaps I should have mentioned it earlier but I wonder if `/usr/etc` is a standard place to install such configuration. [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04.html) doesn't specify it and on my systems: `/usr/etc` is only owned by npm on Fedora, and it does not exist on Debian.

ansasaki:
> > If a configuration file is found in /usr/etc/keylime-agent.conf, it
> > is loaded and any option set overrides the default values
> 
> Perhaps I should have mentioned it earlier but I wonder if `/usr/etc` is a standard place to install such configuration. [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04.html) doesn't specify it and on my systems: `/usr/etc` is only owned by npm on Fedora, and it does not exist on Debian.

I can drop this. I just added this path to make it similar to the python configuration.

ueno:
> > > If a configuration file is found in /usr/etc/keylime-agent.conf, it
> > > is loaded and any option set overrides the default values
> > 
> > 
> > Perhaps I should have mentioned it earlier but I wonder if `/usr/etc` is a standard place to install such configuration. [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04.html) doesn't specify it and on my systems: `/usr/etc` is only owned by npm on Fedora, and it does not exist on Debian.
> 
> I can drop this. I just added this path to make it similar to the python configuration.

Was there any discussion on this on the Python side? If not and it's not too late, maybe good to revisit; I would suggest `/usr/share` instead.

ansasaki:
> > > > If a configuration file is found in /usr/etc/keylime-agent.conf, it
> > > > is loaded and any option set overrides the default values
> > > 
> > > 
> > > Perhaps I should have mentioned it earlier but I wonder if `/usr/etc` is a standard place to install such configuration. [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04.html) doesn't specify it and on my systems: `/usr/etc` is only owned by npm on Fedora, and it does not exist on Debian.
> > 
> > 
> > I can drop this. I just added this path to make it similar to the python configuration.
> 
> Was there any discussion on this on the Python side? If not and it's not too late, maybe good to revisit; I would suggest `/usr/share` instead. See also: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_filesystem_layout

I think the main supported of using the `/usr/etc/` was @aplanas who can give the reasoning for it.

aplanas:
> [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04.html) doesn't specify it and on my systems

Depends on the source.  There is a tendency of using it, that is also referenced here:

https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/usr.html

> I think the main supporter of using the `/usr/etc/` was @aplanas who can give the reasoning for it.

I hope that I am not alone here!

`/usr/etc` it is used to store sane defaults configuration for services. This is provided by the distributor and are not mean to be changed by the user nor the admin, for that it is `/etc`.

Ideally the service will know how to merge the snipped in `/etc` together with the version stored in `/usr/etc`, merging both versions (with priority on the one provided by the admin in `/etc`).  The benefits are:

* Once a package gets updated, the new `/usr/etc` will be updated too (no rpmsave, rpmnew).  So if there is a new option or an old one with a safest value, it will be updated with the package
* If the user removes the version in `/etc`, the service will take the default values

A longer explanation, with more examples can be found here: https://kubic.opensuse.org/blog/2019-12-05-usr-etc/

For doing the merge in services that do not support it by default, we are using [libeconf](https://github.com/openSUSE/libeconf), and doing upstream commits to support it.

src/config.rs
```diff
@@ -0,0 +1,725 @@
...
+pub static DEFAULT_PORT: u32 = 9002;
+pub static DEFAULT_CONTACT_IP: &str = "127.0.0.1";
+pub static DEFAULT_CONTACT_PORT: u32 = 9002;
+pub static DEFAULT_REGISTRAR_IP: &str = "127.0.0.1";
+pub static DEFAULT_REGISTRAR_PORT: u32 = 8890;
+pub static DEFAULT_ENABLE_AGENT_MTLS: bool = true;
+pub static DEFAULT_KEYLIME_DIR: &str = "/var/lib/keylime";
+pub static DEFAULT_SERVER_KEY: &str = "server-private.pem";
+pub static DEFAULT_SERVER_CERT: &str = "server-cert.crt";
+pub static DEFAULT_SERVER_KEY_PASSWORD: &str = "";
```
aplanas, line 34: Should be `None` ...?

src/config.rs
```diff
@@ -0,0 +1,725 @@
...
+            revocation_notification_port: Some(
+                DEFAULT_REVOCATION_NOTIFICATION_PORT,
+            ),
+            secure_size: DEFAULT_SECURE_SIZE.to_string(),
+            payload_script: DEFAULT_PAYLOAD_SCRIPT.to_string(),
+            dec_payload_file: DEFAULT_DEC_PAYLOAD_FILE.to_string(),
+            enc_keyname: DEFAULT_ENC_KEYNAME.to_string(),
+            extract_payload_zip: DEFAULT_EXTRACT_PAYLOAD_ZIP,
+            server_key: Some("default".to_string()),
+            server_key_password: Some(
```
aplanas, line 337: ... because this put `Some("")` (there is a password, that is the empty string), that is different from `None` (no password was provided)

ansasaki: The empty string is considered "no password". Check:
https://github.com/ansasaki/rust-keylime/blob/036ab2a392022641a8603fddb405b58dd36430b8/src/crypto.rs#L77-L101
and
https://github.com/ansasaki/rust-keylime/blob/036ab2a392022641a8603fddb405b58dd36430b8/src/crypto.rs#L54-L72

ansasaki:
@lkatalin @ueno @aplanas I think the only pending decision here is about using `/usr/etc` or not.

Please let me know if there are other points to be modified.

Before merging, we need to drop the last commit which temporarily changes the tests repository.

ashcrow:
>  I think the only pending decision here is about using /usr/etc or not.

Quick reminder that `/usr` is read-only on CoreOS.

lkatalin:
I did a first pass and it seemed good to me, I can do another one shortly.

aplanas:
>  the only pending decision here is about using /usr/etc or not

My vote is for `/usr/etc`. We can add `/usr/lib` (like systemd) or `/usr/share` as new search paths later if there are other standards. 

> Quick reminder that `/usr` is read-only on CoreOS.

Exactly. /usr/etc will be deployed with the image, never edited by the user (that should edit /etc)

lkatalin (approved):
This LGTM and I appreciate that the config locations will match the Python version. Hopefully this will help with the CI.

ueno:
@aplanas thanks for the detailed explanation!

> A longer explanation, with more examples can be found here: https://kubic.opensuse.org/blog/2019-12-05-usr-etc/
> 
> For doing the merge in services that do not support it by default, we are using [libeconf](https://github.com/openSUSE/libeconf), and doing upstream commits to support it.

I wasn't aware of the UsrEtc effort. It makes sense from the package update point of view, though other distributions currently don't seem to catch up; as mentioned in the link to the Fedora packaging [guideline](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_filesystem_layout):

> Fedora follows the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/fhs.shtml) with regards to filesystem layout, with exceptions noted below. The FHS defines where files should be placed on the system.
[...]
> Fedora does not allow new directories directly under / or /usr without FPC approval.

That said, I guess it shouldn't be a problem as long as the Fedora package doesn't ship any files under `/usr/etc`.

ansasaki:
> > Fedora follows the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/fhs.shtml) with regards to filesystem layout, with exceptions noted below. The FHS defines where files should be placed on the system.
> > [...]
> > Fedora does not allow new directories directly under / or /usr without FPC approval.
> 
> That said, I guess it shouldn't be a problem as long as the Fedora package doesn't ship any files under `/usr/etc`.

I'll make sure the Fedora package doesn't ship any files there.

aplanas:
> Maybe the location could be configurable at build time

That is a good idea

Issue #441, opened 2022-08-25
Rust agent fails main Keylime CI when getting pub keys or identity quote


lkatalin:
All other 128 tests in [`run_local.sh`](https://github.com/keylime/keylime/blob/master/.ci/run_local.sh) pass with the Rust agent.

Output for `pubkey_get`:
```
Error in test.test_restful.TestRestful.test_071_agent_keys_pubkey_get
  File "/usr/lib/python3.9/site-packages/urllib3/connection.py", line 159, in _new_conn
    conn = connection.create_connection(
  File "/usr/lib/python3.9/site-packages/urllib3/util/connection.py", line 84, in create_connection
    raise err
  File "/usr/lib/python3.9/site-packages/urllib3/util/connection.py", line 74, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.9/site-packages/urllib3/connectionpool.py", line 670, in urlopen
    httplib_response = self._make_request(
  File "/usr/lib/python3.9/site-packages/urllib3/connectionpool.py", line 381, in _make_request
    self._validate_conn(conn)
  File "/usr/lib/python3.9/site-packages/urllib3/connectionpool.py", line 978, in _validate_conn
    conn.connect()
  File "/usr/lib/python3.9/site-packages/urllib3/connection.py", line 309, in connect
    conn = self._new_conn()
  File "/usr/lib/python3.9/site-packages/urllib3/connection.py", line 171, in _new_conn
    raise NewConnectionError(
urllib3.exceptions.NewConnectionError: <urllib3.connection.HTTPSConnection object at 0x7f8b6ed0ce50>: Failed to establish a new connection: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.9/site-packages/requests/adapters.py", line 439, in send
    resp = conn.urlopen(
  File "/usr/lib/python3.9/site-packages/urllib3/connectionpool.py", line 726, in urlopen
    retries = retries.increment(
  File "/usr/lib/python3.9/site-packages/urllib3/util/retry.py", line 439, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
[...]

PR #450, opened 2022-09-19
Add build script to set `feature=testing` cfg

lkatalin:
This is necessary when running the `keylime/keylime` CI with `run_local.sh`, which indicates a test with the env var `KEYLIME_TEST` set to `'true'` (as a string). Otherwise, the `MOUNT_SECURE` flag in `common.rs` is set incorrectly when running this with the CI as opposed to in a Rust-based testing environment.

Related to #441


lkatalin:
Closing as this no longer seems necessary to get the Keylime CI to pass with the Rust agent.

PR #460, opened 2022-09-26
Use cargo tarpaulin with the --all-features flag

avery-blanchard:
When using `cargo tarpaulin` in CI testing, run with the flag `--all-features`. This change increases testing coverage by at least 10%. @lkatalin


lkatalin (approved):
lgtm unless others have objections

PR #481, opened 2022-12-05
Split crates into library and applications

ueno:
This is a start of refactoring the codebase into three sub-crates: keylime, keylime_agent, and keylime_ima_emulator.  The keylime crate provides common facility used by the other crates as a library (currently only IMA parser).


keylime-agent/Cargo.toml
```diff
@@ -0,0 +1,58 @@
+[package]
+authors = ["Keylime Authors"]
+edition = "2018"
```
aplanas, line 3: maybe is time to move it to 2021 like the keylime crate? or there is a dependency on the minimal version used for the compiler?

tests/nopanic.ci
```diff
@@ -6,16 +6,17 @@ Copyright 2021 Keylime Authors
 To prevent CI failing for approved instance of banned words, add a comment: //#[allow_ci]
 '''
 
-import os
+import subprocess
 
 banned = ["unwrap(", "panic!("]
 
-srcs = os.listdir("src")
+res = subprocess.run(["git", "ls-files", "--cached", "--exclude", "*.rs", "--ignored"], capture_output=True)
```
aplanas, line 13: IIUC this is equivalent to `find src/ -name '*.rs'`. I would like to not depend on `git` cli to run tests. To express this in Python it is as:

```python
$ python3
Python 3.10.8 (main, Oct 19 2022, 07:46:20) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pathlib
>>> list(pathlib.Path("src/").glob("**/*.rs"))
[PosixPath('src/algorithms.rs'), PosixPath('src/ima_entry.rs'), PosixPath('src/notifications_handler.rs'), PosixPath('src/crypto.rs'), PosixPath('src/main.rs'), PosixPath('src/version_handler.rs'), PosixPath('src/serialization.rs'), PosixPath('src/keys_handler.rs'), PosixPath('src/config.rs'), PosixPath('src/revocation.rs'), PosixPath('src/error.rs'), PosixPath('src/common.rs'), PosixPath('src/ima.rs'), PosixPath('src/tpm.rs'), PosixPath('src/registrar_agent.rs'), PosixPath('src/secure_mount.rs'), PosixPath('src/errors_handler.rs'), PosixPath('src/quotes_handler.rs'), PosixPath('src/ima_emulator.rs'), PosixPath('src/permissions.rs')]
```

aplanas (commented):
[Seems that the tests fail because of the shim.py move]

tests/nopanic.ci
ueno: A benefit of using `git` is that it allows us to ignore files not tracked in git, so you could run the check in pre-commit phase even if extra files are generated in the build process. I can switch to pathlib for this particular case.

keylime-agent/Cargo.toml
ueno: That sounds like a good idea.

ueno:
> [Seems that the tests fail because of the shim.py move]

I've reverted that change (btw: I would put shim.py in dist/ instead of tests/ as it is installed), though the tests are still failing because they [modify](https://github.com/RedHat-SP-Security/keylime-tests/blob/6e429c381f86ea05eba82ec9578e5cc9de074645/setup/install_upstream_rust_keylime/test.sh#L23) the source code.

aplanas:
> > [Seems that the tests fail because of the shim.py move]
> 
> I've reverted that change (btw: I would put shim.py in dist/ instead of tests/ as it is installed), though the tests are still failing because they [modify](https://github.com/RedHat-SP-Security/keylime-tests/blob/6e429c381f86ea05eba82ec9578e5cc9de074645/setup/install_upstream_rust_keylime/test.sh#L23) the source code.

Ah that is unfortunate ... I wonder what can be the correct approach here. One can be fork the keylime-tests to adapt the tests and add in the PR the address of the forked repo (I think that I saw this approach in other PR)

Another one is compile the agent as debug when testing, and adding a `#[cfg(debug_assertions)]` that will fetch the IMA measurement path from a env var if set.

ansasaki:
> I also tried to move `tpm.rs` into the library crate. The generated documentation is temporarily hosted [here](https://ueno.fedorapeople.org/rust-keylime-docs/keylime/). My idea is to port low-level primitives first, then eventually provide a high-level (protocol) API over them. That way, we could gradually migrate the code out of the agent to the library.
> 
> @lkatalin @ansasaki any thoughts on that?

Thanks for working on this! I think it makes sense to follow this incremental approach and progressively move the code out of the agent and into the library.

> I've reverted that change (btw: I would put shim.py in dist/ instead of tests/ as it is installed), though the tests are still failing because they [modify](https://github.com/RedHat-SP-Security/keylime-tests/blob/6e429c381f86ea05eba82ec9578e5cc9de074645/setup/install_upstream_rust_keylime/test.sh#L23) the source code.

About  the failing test, maybe we could use a branch while we don't have a way to configure the measured boot log path. @kkaarreell Could TMT dynamic reference help us here?

kkaarreell:
> About the failing test, maybe we could use a branch while we don't have a way to configure the measured boot log path. @kkaarreell Could TMT dynamic reference help us here?

I think I have all the required changes in https://github.com/RedHat-SP-Security/keylime-tests/pull/275
To use those updates please (temporarily) modify packit-ci.fmf file and in the `environment` section define variable `RUST_KEYLIME_UPSTREAM_BRANCH: rust_pr_481`

ueno:
> > About the failing test, maybe we could use a branch while we don't have a way to configure the measured boot log path. @kkaarreell Could TMT dynamic reference help us here?
> 
> I think I have all the required changes in [RedHat-SP-Security/keylime-tests#275](https://github.com/RedHat-SP-Security/keylime-tests/pull/275) To use those updates please (temporarily) modify packit-ci.fmf file and in the `environment` section define variable `RUST_KEYLIME_UPSTREAM_BRANCH: rust_pr_481`

Thank you! I'm assuming that you meant `ref: rust_pr_481` under `discover`, instead of `RUST_KEYLIME_UPSTREAM_BRANCH`. With that change, those packit tests are now passing.

kkaarreell:
Right, of course, sorry for that.

lkatalin (approved):

kkaarreell:
> Thank you! I'm assuming that you meant `ref: rust_pr_481` under `discover`, instead of `RUST_KEYLIME_UPSTREAM_BRANCH`. With that change, those packit tests are now passing.

I have merged the test update. You can restore original reference

ueno:
> > Thank you! I'm assuming that you meant `ref: rust_pr_481` under `discover`, instead of `RUST_KEYLIME_UPSTREAM_BRANCH`. With that change, those packit tests are now passing.
> 
> I have merged the test update. You can restore original reference

Thanks; done and merged.

PR #486, opened 2022-12-07
main: only read uuid from KeylimeConfig

THS-on:
The main function also had a local variable called uuid. This variable
contained either the provided uuid or when "hash_ek" was set the sha256 hash
of the EK. We never updated the KeylimeConfig struct which caused a mismatch
between uuids used in the program. Instead we now mutate the KeylimeConfig
when "hash_ek" is set to the correct value.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -508,6 +508,14 @@ async fn main() -> Result<()> {
     // Calculate the SHA-256 hash of the public key in PEM format
     let ek_hash = hash_ek_pubkey(ek_result.public.clone())?;
 
+    // Replace the uuid with the actual EK hash if the option was set.
+    // We cannot do that when the configuration is loaded initially,
+    // because only have later access to the the TPM.
+    config.agent.uuid = match config.agent.uuid.as_str() {
+        "hash_ek" => ek_hash.clone(),
+        s => s.to_string(),
+    };
+
     // Try to load persistent Agent data
     let old_ak = match &config.agent.agent_data_path {
         Some(path) => {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -602,12 +602,15 @@ async fn main() -> Result<()> {
         }
     }
 
-    let uuid = match config.agent.uuid.as_str() {
+    // Replace the uuid with the actual EK hash if the option was set.
+    // We cannot do that when the configuration is loaded initially,
+    // because only have later access to the the TPM.
+    config.agent.uuid = match config.agent.uuid.as_str() {
```
aplanas, line 608: Makes a lot of sense, but I would move it a bit early, around line 509 where `ek_hash` is already calculated just after `ek_result` is created. This will also remove some duplication.

THS-on: Good catch, I moved the change closer to line 509 and use the hash_ek value.

aplanas (approved):

kkaarreell:
I have tried to implement reproducer for this issue in keylime-tests but my test scenario is failing even with this fix.
https://github.com/RedHat-SP-Security/keylime-tests/pull/272

THS-on:
The issue was with the reproducer and not the fix, if I see that correctly.

kkaarreell:
> I have tried to implement reproducer for this issue in keylime-tests but my test scenario is failing even with this fix. [RedHat-SP-Security/keylime-tests#272](https://github.com/RedHat-SP-Security/keylime-tests/pull/272)

Yes, false alarm. I am sorry for that

ansasaki (approved):

PR #487, opened 2022-12-08
systemd: start agent after network is online

THS-on:


aplanas (approved):
Would be better to sync this and the mount from the Python agent with this version? Or is irrelevant as it is now deprecated?

THS-on:
@aplanas I'll open also a PR for the Python side.

ansasaki (approved):
LGTM!

PR #499, opened 2023-01-13
Refactor async code synchronization

ansasaki:
Use messages for synchronization instead of mutexes and condvar.  Add
separate workers for keys, payload, and revocation execution, as well as
a worker to handle the revocations coming from ZeroMQ channel.

The workers  hold the state locally instead of storing all data in server
context (QuoteData).

The keys worker holds the received U and K keys and is responsible for
combining them to obtain K.  It also provides the K key when requested
by the /keys/verify endpoint to prove that the resulting K is correct.

When the U and V keys combination is successful (the HMAC verification
is successful), the keys worker wakes the payloads worker sending the
obtained K key and the encrypted payload.

The payloads worker is responsible for decrypting and running the
payload sent by the keys worker.

After decrypting and running the payload, the payloads worker will send
messages to the revocation worker and ZMQ worker to start listening for
revocations.

This also implements graceful shutdown on Ctrl + C (SIGINT) signal.
A dedicated task is spawned to receive the signal and stop the other workers.

Unit tests were implemented for the changes.

Fixes: #363, #447

PR #500, opened 2023-01-13
Prevent HMAC mismatch error from disabling rust agent.

galmasi:
This is a trivial fix to prevent situations in which U/V key mismatches disable the keylime agent, putting it into a state where it is no longer possible to successfully activate it.

Because of the simplicity of the fix I didn't bother with a separate Issue to describe the problem, but here is a potted summary of it:

* The bug I'm attempting to fix can be produced by calling two `keylime_tenant -c add` commands in close succession on the same agent, causing the U key of the *second* `add` to mix with the V key from the verifier of the *first* `add`. This causes the HMAC check failure, and the agent now has populated U and V keys, ensuring that later calls to `u_key` or `v_key` will continue to cause this failure. Currently this can only be fixed by a forced re-start and re-registration of the keylime agent.

* The fix simply clears the U and V keys when HMAC failure is registered. No loss of actual information, since the system is broken on the verifier side -- the agent could not be activated. However, once the keys are cleared we are allowing a new tenant command to succeed, since now the HMAC check will be bypassed until both U and V are collected.

*Note* this fix does not deal with the state problems in the keylime verifier. It merely enables the keylime agent to recover from situations that the tenant/verifier should have never put it into. Separate fixes will be required to ensure the tenant and verifier behave well.

Signed-off-by: George Almasi <gheorghe@us.ibm.com>

### DIFF
--- a/keylime-agent/src/keys_handler.rs
+++ b/keylime-agent/src/keys_handler.rs
@@ -82,6 +82,9 @@ pub(crate) fn try_combine_keys(
         }
     }
 
+    // we have failed to decode the U/V keys. Clear keys to enable later attempts.
+    keyset1.clear();
+    keyset2.clear();
     Err(Error::Other(
         "HMAC check failed for all U and V key combinations".to_string(),
     ))


ueno:
Thank you for the PR; I'm a bit confused why it's necessary, as the [loop](https://github.com/keylime/rust-keylime/pull/500/files#diff-12478a01ccbf6df0642466400e6b33263b647b5ecd5df1c4cba2f0bd9af1cf7aL64) above checks all combinations of elements in `keyset1` and `keyset2` every time the function (`try_combine_keys`) is called.

aplanas:
> I'm a bit confused why it's necessary

IIUC is because when the check fails there is no `return` in line `80`, and so there is no process to clear both vectors (lines `78..=79`). So in case of error both `KeySet` will still be populated with data (maybe producing the error that is described)

I wonder if one possibility here is to drop both `.clear()` and transform the `.iter()` in `.into_iter()`, to guarantee the consumption of both KeySet.

galmasi:
Thank you for reviewing @ueno . WRT your question, --> @aplanas has the correct explanation. There is no process for clearing the V key after a failed attempt to activate. And since the V key "exists" any attempt to insert a new U key will result in the HMAC mismatch error, aborting the process. Checkmate with no way out except to kill the agent. 

The fix makes it possible to start a new activation after a failed one.

ansasaki:
I think the real problem are the `?` in lines 66 and 70: they are the reason for the loop to be interrupted without trying all combinations of received `U` and `V` keys. Basically the operations inside the loop should not interrupt the loop in any case other than successful HMAC verification.

In my interpretation, it wouldn't be a problem to have old `U` or `V` keys in the global vectors: once the key combination is successful and the HMAC matches, the vectors are cleaned.

The current problem is to have elements in those vectors that breaks the `xor()` operation (wrong lengths) or if the received HMAC fails to be hex decoded.

The first is probably not the case, unless somehow there are tenants mixing AES-128 and AES-256 keys (note that the `SymmKey` type validates the size). The second case is easier to happen as the HMAC values received are not validated before overwriting the old value.

Which leads to another weird thing in the current code: we keep all received `U` keys, but we overwrite the associated HMAC, meaning that only the last received `U` key is valid.  If we really want to keep and try old `U` keys, we should store them with the associated HMAC.

If a payload is provided, it should also be stored with the associated `U` and HMAC, since the key used to encrypt it will only be derived with the associated `U` key.

Speaking of payloads, I think we simply do the wrong thing: we append the received payload to the previously received payload. This don't make any sense as each payload is encrypted with a different key and this operation would make impossible to correctly decrypt the payload.

The easiest solution is to keep only a single tuple of `U`, HMAC and payload. When we receive a new set, we simply overwrite what we had.

If we want to support multiple payloads being sent by the tenant, we should keep the `(U, HMAC, payload)` in a list.

ansasaki:
@galmasi Please check my proposal to fix this issue in https://github.com/keylime/rust-keylime/pull/499/commits/47cc471b5888666912a986e0e931d7f5534a308b

I made other related changes in https://github.com/keylime/rust-keylime/pull/499 (reviews welcomed)

galmasi:
I looked at 47cc471. I'm not going to claim full understanding, but it seems you are including the auth tag with the ukey so that multiple versions can exist; in this situation tags and keys are going to continue to accumulate and if there is a match the payloads will be decrypted.

* I think this precludes the need for deleting v keys, so I'm going to close my PR.
* I'm looking forward to testing your code and confirming that I *cannot* produce failures anymore. 
* A small nit -- by accumulating keys forever you are exposing the keylime agent to grabbing increasing amounts of memory. I'm not going to claim that a viable resource exhaustion attack can be mounted this way, I just have a small worry about it :)

PR #459, opened 2022-09-26
Cargo.toml: tss-esapi bindings

aplanas:
Generate the tss-esapi bindings during compilation time.  This will extend the ammount of supported architectures.

Signed-off-by: Alberto Planas <aplanas@suse.com>

PR #502, opened 2023-01-17
Cargo.toml: tss-esapi bindings

aplanas:
Generate the tss-esapi bindings during compilation time.  This will extend the ammount of supported architectures.

This is a regression done after the crate split, that was fixed before in #459.

### DIFF
--- a/keylime-ima-emulator/Cargo.toml
+++ b/keylime-ima-emulator/Cargo.toml
@@ -12,4 +12,4 @@ keylime = { path = "../keylime" }
 log = "0.4"
 openssl = "0.10.15"
 thiserror = "1.0"
-tss-esapi = "7.1.0"
+tss-esapi = {version = "7.1.0", features = ["generate-bindings"]}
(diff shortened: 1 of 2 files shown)


ueno (approved):
Thanks for catching this; looks good to me (nit: a typo in commit message: "ammount" → "amount").

aplanas:
I will fix the commit message, but I will drop the changes in the lock file, I think that they should not be there

aplanas:
Should me done, tested here: https://build.opensuse.org/package/show/home:aplanas:branches:security/rust-keylime and we have ppc and s390x binaries : )

ueno:
cargo audit detects an issue that the unmaintained `ansi_term` crate is used through tss-esapi → bindgen 0.59.2 → clap 2:
https://github.com/keylime/rust-keylime/actions/runs/3938762221/jobs/6737828590#step:5:64

Maybe we need a bump of bindgen requirement in rust-tss-esapi to 0.60 at least?

aplanas:
> bump of bindgen requirement in rust-tss-esapi to 0.60 at least?

Done here: https://github.com/parallaxsecond/rust-tss-esapi/pull/386

PR #504, opened 2023-01-19
packit-ci: Add hotfix for tpm2-tss Fedora BZ#2158598

kkaarreell:
Signed-off-by: Karel Srot <ksrot@redhat.com>


aplanas (approved):
Note for me: check if we have the same bug

ansasaki (approved):

kkaarreell:
We can eventually extend the `tpm2-tss` downgrade on Rawhide too. @ansasaki WDYT?

kkaarreell:
There is nothing to downgrade to, we should disable Rawhide completely.

ansasaki:
> We can eventually extend the `tpm2-tss` downgrade on Rawhide too. @ansasaki WDYT?

If there is no workaround, let's disable it for now

PR #511, opened 2023-02-01
crypto, mTLS: allow certificate chain for trusted_client_ca

THS-on:
This allows for setups with intermediate CAs.

### DIFF
--- a/keylime-agent/src/crypto.rs
+++ b/keylime-agent/src/crypto.rs
@@ -132,8 +132,7 @@ pub(crate) fn pkey_pub_from_priv(
             PKey::from_rsa(rsa).map_err(Error::Crypto)
         }
         id => Err(Error::Other(format!(
-            "pkey_pub_from_priv not yet implemented for key type {:?}",
-            id
+            "pkey_pub_from_priv not yet implemented for key type {id:?}"
         ))),
     }
 }
@@ -323,8 +322,7 @@ pub(crate) fn decrypt_aead(key: &[u8], data: &[u8]) -> Result<Vec<u8>> {
         AES_256_KEY_LEN => Cipher::aes_256_gcm(),
         other => {
             return Err(Error::Other(format!(
-                "key length {} does not correspond to valid GCM cipher",
-                other
+                "key length {other} does not correspond to valid GCM cipher"
             )))
         }
     };
@@ -395,9 +393,8 @@ pub mod testing {
             AES_256_KEY_LEN => Cipher::aes_256_gcm(),
             other => {
                 return Err(Error::Other(format!(
-                    "key length {} does not correspond to valid GCM cipher",
-                    other
-                )))
+                "key length {other} does not correspond to valid GCM cipher"
+            )))
             }
         };
         if iv.len() != AES_BLOCK_SIZE {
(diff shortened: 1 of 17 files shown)


THS-on:
cc: @aplanas @ansasaki

keylime-agent/src/crypto.rs
```diff
@@ -44,6 +43,15 @@ pub(crate) fn load_x509(input_cert_path: &Path) -> Result<X509> {
     Ok(cert)
 }
 
+pub(crate) fn load_x509_cert_chain(
+    input_cert_path: &Path,
+) -> Result<Vec<X509>> {
+    let contents = read_to_string(input_cert_path)?;
+    let mut cert_chain = X509::stack_from_pem(contents.as_bytes())?;
```
aplanas, line 50: you are not mutating cert_chain, and if I read the code correcty this can be the last line:

```rust
X509::stack_from_pem(contents.as_bytes())
```

aplanas (changes requested):

keylime-agent/src/crypto.rs
THS-on: Yes fixed. I forgot to change that, thanks for catching it!

aplanas (approved):
AFAICT LGTM

ansasaki:
LGTM! Could you please fix that line clippy is complaining about?

THS-on:
@ansasaki done

THS-on:
It seems like that something has changed how clippy deals with references of strings. My local version of clippy (0.1.67) does not report it.  I'll update the rest of issues later.

aplanas:
It is complaining now about the string interpolation

THS-on:
> It is complaining now about the string interpolation

@aplanas should be fixed now.

keylime/src/algorithms.rs
```diff
@@ -109,8 +108,7 @@ impl TryFrom<&str> for EncryptionAlgorithm {
             "rsa" => Ok(EncryptionAlgorithm::Rsa),
             "ecc" => Ok(EncryptionAlgorithm::Ecc),
             _ => Err(AlgorithmError::Encrypt(format!(
-                "Encryption alogorithm {} not supported by Keylime",
-                value
+                "Encryption alogorithm {value} not supported by Keylime"
```
aplanas, line 111: "algorithm"

keylime-agent/src/main.rs
```diff
@@ -691,23 +690,24 @@ async fn main() -> Result<()> {
...
-            }
-        }?;
+        let keylime_ca_certs =
+            match crypto::load_x509_cert_chain(&ca_cert_path) {
+                Ok(t) => Ok(t),
+                Err(e) => {
+                    error!(
+                        "Failed to load trusted CA certificate {}: {}",
+                        ca_cert_path.display(),
+                        e
```
aplanas, line 700: clippy should complain here too, shouldn't?

THS-on: No because you cannot call functions currently in a fromat string, I think.

aplanas: Right. But I mean "e"

THS-on: Yeah, I don't know. I would keep it this way for now, because mixing the two styles looks wrong.

ansasaki (approved):
LGTM! Thank you for going through clippy complaints and fixing them!

Issue #510, opened 2023-02-01
Environment override does not work if option cointains "_"


THS-on:
For example `KEYLIME_AGENT_CONACT_IP` does nothing.

This is a known problem in the `config-rs` module: https://github.com/mehcode/config-rs/issues/391

The suggestion is to use another separator, e.g double underscores or: `KEYLIME__AGENT__CONACT_IP`

THS-on:
@aplanas @ansasaki any ideas on how to fix this nicely?

ansasaki:
@THS-on Yeah, this is bad and I don't see a clean way to make it not ambiguous for the library. We would need to make the separator something that is prohibited to be used when naming the configuration options. The least ugly options I see are:

* Dots (e.g. `KEYLIME.AGENT.CONTACT_IP`)
* Double underscores (e.g. `KEYLIME__AGENT__CONTACT_IP`)

@maugustosilva FYI, it seems we have a bug related with overriding the configuration with environment variables.

ansasaki:
Another option would be if we modified the internal configuration struct to not have 2 levels and makes the whole `KEYLIME_AGENT_` the prefix. Then we set the separator to be anything other than `_` and it should work.

Currently the struct is something like:

- Keylime
  -  agent
     - contact_ip 
     
We would need to change it to:

- Keylime_agent
  - contact_ip

THS-on:
The question is if we want to support nested levels or not. I don't have a big opinion about it as long as the feature works.

ansasaki:
> The question is if we want to support nested levels or not. I don't have a big opinion about it as long as the feature works.

It is not just a question of nested levels, but a decision to change the configuration file itself. Currently, to make it closer to the python agent configuration, the configuration for the rust side needs all the fields to be in the `[agent]` section. With the change I proposed above it would need a change in the configuration file to drop the `[agent]` section completely (meaning that we need another version for configuration files and respective templates for upgrades)

mpeters:
I like the idea of dropping the `[agent]` section (of course we'd need to support backwards compat, but that backwards compat doesn't need to support env overrides since those don't work anyway). I don't think anything about the agent would need nested sections.

THS-on:
@mpeters this has also the advantage to make it also consistent with the server side, so it is probably the best approach.

ansasaki:
I have a working solution that does not require any change to the configuration file nor to the environment variables format.  The solution is to read the environment configuration separately and then use these values to override what was obtained from the configuration files.

PR #514, opened 2023-02-09
config: Fix overriding options with env vars

ansasaki:
Read the configuration options set via environment variables separately
to initialize an EnvConfig structure and use this structure as a
Source to override the values of the KeylimeConfig obtained from the
configuration files.

For this, change the prefix for the environment variables to
KEYLIME_AGENT with the 'prefix_separator' as '_', and 'separator' as
'.'.

Then the configuration options obtained from the environment variables
are used to initialize an EnvConfig structure, which does not contain
sub-structures.

The EnvConfig structure is a duplicate of AgentConfig structure, but
contains Option<T> for all configuration options, allowing values to not
be set via environment variables.

The remaining optional fields in AgentConfig were replaced to be
required (not Option<T>).

Note that the verification of required/non-required options
was not used since all values are set with the default values and then
successively overriden by values from the configuration files and
environment variables.

The Source trait was implemented for the EnvConfig structure in a way
that it is possible to use it as the source for KeylimeConfig structure.

Fixes: #510

### DIFF
--- a/keylime-agent/src/config.rs
+++ b/keylime-agent/src/config.rs
@@ -59,196 +59,251 @@ pub static DEFAULT_AGENT_DATA_PATH: &str = "agent_data.json";
 pub static DEFAULT_CONFIG: &str = "/etc/keylime/agent.conf";
 pub static DEFAULT_CONFIG_SYS: &str = "/usr/etc/keylime/agent.conf";
 
-impl Source for KeylimeConfig {
-    fn collect(&self) -> Result<Map<String, Value>, ConfigError> {
-        let agent: Map<String, Value> = Map::from([
-            ("version".to_string(), self.agent.version.to_string().into()),
-            ("uuid".to_string(), self.agent.uuid.to_string().into()),
-            ("ip".to_string(), self.agent.ip.to_string().into()),
-            ("port".to_string(), Value::from(self.agent.port)),
-            (
-                "contact_ip".to_string(),
-                if let Some(ref s) = self.agent.contact_ip {
-                    s.to_string().into()
-                } else {
-                    "".into()
-                },
-            ),
-            (
-                "contact_port".to_string(),
-                if let Some(ref s) = self.agent.contact_port {
-                    s.to_string().into()
-                } else {
-                    "".into()
-                },
-            ),
-            (
-                "registrar_ip".to_string(),
-                self.agent.registrar_ip.to_string().into(),
-            ),
-            (
-                "registrar_port".to_string(),
-                self.agent.registrar_port.into(),
-            ),
-            (
-                "enable_agent_mtls".to_string(),
-                self.agent.enable_agent_mtls.into(),
-            ),
-            (
-                "keylime_dir".to_string(),
-                self.agent.keylime_dir.to_string().into(),
-            ),
-            (
-                "server_key".to_string(),
-                if let Some(ref s) = self.agent.server_key {
-                    s.to_string().into()
-                } else {
-                    "".into()
-                },
-            ),
-            (
-                "server_key_password".to_string(),
-                if let Some(ref s) = self.agent.server_key_password {
-                    s.to_string().into()
-                } else {
-                    "".into()
-                },
-            ),
-            (
-                "server_cert".to_string(),
-                if let Some(ref s) = self.agent.server_cert {
-                    s.to_string().into()
-                } else {
-                    "".into()
-                },
-            ),
-            (
-                "trusted_client_ca".to_string(),
-                if let Some(ref s) = self.agent.trusted_client_ca {
-                    s.to_string().into()
-                } else {
-                    "".into()
-                },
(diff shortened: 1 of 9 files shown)


keylime-agent/src/config.rs
```diff
@@ -59,244 +59,228 @@ pub static DEFAULT_AGENT_DATA_PATH: &str = "agent_data.json";
...
+    pub agent_data_path: Option<String>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) struct KeylimeConfig {
+    pub agent: AgentConfig,
+}
+
+impl AgentConfig {
+    pub fn get_map(&self) -> Map<String, Value> {
```
aplanas, line 106: OK for me, but maybe we can drop concept of "configuration as a dictionary", and use the struct directly

keylime-agent/src/config.rs
```diff
@@ -59,244 +59,228 @@ pub static DEFAULT_AGENT_DATA_PATH: &str = "agent_data.json";
...
-                if let Some(ref s) = self.agent.server_key {
-                    s.to_string().into()
-                } else {
-                    "".into()
-                },
-            ),
-            (
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub(crate) struct AgentConfig {
+    pub version: Option<String>,
```
aplanas, line 64: Now everything is optional, can this be a problem?

keylime-agent/src/config.rs
ansasaki: We use the struct directly, the collection to get the map is only used to implement `Source` to allow overriding options with the values obtained from the environment variable.

keylime-agent/src/config.rs
ansasaki: It is not a problem because since the last big configuration change to TOML format and to use the `config` crate, we always base the configuration from the defaults. So basically there is no case when an option is not set.

aplanas (approved):

ansasaki:
> Can we move the default value part into the `KeylimeConfig` struct? I don't really like that we now have checks for the defaults spread all over the code base.

Actually all of those are dead code, as the options are always set. The problem is that we need the internal `AgentConfig` structure to be made of `Option<T>` to allow config to be partially filled from the environment variables, then use it to override the options from the files.

I can replace all of them with `except()` or similar that would panic if not set, but usually this is not the preferred way.

We can alternatively make the `AgentConfig` structure to not contain any `Option<T>` and create a new structure duplicating the `AgentConfig` struct but made of `Option<T>` to read from the environment (e.g. `EnvConfig`). I tried to avoid this when writing this changes, but maybe it is the best way since it makes it clear the purposes of each structure: The `EnvConfig` which may be partially set and used only to read config from the environment and the `AgentConfig` which is the actual configuration where all the fields are required to be set.

THS-on:
Yeah I think the last approach is probably the best. Then we have the duplication only in one place.

ansasaki:
@THS-on I rewrote the code to add `EnvConfig` which is a duplicate of  `AgentConfig` but with all fields optional. The remaining optional fields in `AgentConfig` were also replaced by required fields (i.e. not `Option<T>`).

I removed some checks for options set throughout the code as the options are all required. I had to keep checks for empty strings is some cases, though.

I also added instructions to the configuration file on how to override options using env vars.

keylime-agent/src/main.rs
```diff
@@ -170,27 +170,32 @@ async fn main() -> Result<()> {
...
+    let payload_script = &config.agent.payload_script;
+
     // The agent cannot run when a payload script is defined, but mTLS is disabled and insecure
     // payloads are not explicitly enabled
-    if !&config.agent.enable_agent_mtls
-        && !&config.agent.enable_insecure_payload
-        && !&config.agent.payload_script.is_empty()
+    if !enable_agent_mtls
+        && !enable_insecure_payload
+        && !payload_script.is_empty()
```
THS-on, line 181: Do we still need this change?

THS-on (commented):
@ansasaki thank you for changing it :) 

I think the empty strings checks are fine, because that are not that many.

keylime-agent/src/main.rs
ansasaki: Hum, I think this can be dropped.

ansasaki: @THS-on Do you have any other finding? Otherwise, I'd like to merge this.

THS-on: Let me just build it and test it. But I think it is good in the current form. Thank you for fixing it!

THS-on (approved):

keylime-agent/src/main.rs
THS-on: It works as intended :)

ansasaki: Thank you for the review!

PR #517, opened 2023-02-15
cargo deb: include shim.py in packaging

THS-on:


aplanas (approved):

ashcrow:
Looks like the failure is in a package conflict. Likely not related to the change here.

THS-on:
> Looks like the failure is in a package conflict. Likely not related to the change here.

@ashcrow yes the issue is known, but not yet fixed. I think we can merge this.

aplanas:
So I guess that Debian is deploying the version that support the Python interface (legacy-python-actions)?

In openSUSE we are compiling it without Python support (`--no-default-features --features "with-zmq"`), so there is not requirement of having Python installed in the nodes

THS-on:
Yeah the current cargo deb package is build with all the features enabled.

ansasaki:
On Fedora `legacy-python-actions` is disabled but `with-zmq` is enabled. On RHEL both are disabled.

I propose we deprecate `legacy-python-actions` and `with-zmq` before the next tag/release. Then, on next major release we could remove the support completely.

aplanas:
> I propose we deprecate legacy-python-actions and with-zmq

I agree!

So can be OK to remove "with-zmq"?

ansasaki:
> So can be OK to remove "with-zmq"?

In my opinion, yes. I think that using REST API for revocation notifications support is sufficient.

THS-on:
> I propose we deprecate legacy-python-actions and with-zmq before the next tag/release. Then, on next major release we could remove the support completely.

Yes sounds good to me!

PR #521, opened 2023-02-16
Deprecate `with-zmq` and `legacy-python-actions` features

ansasaki:
The features are disabled by default and will be removed on next major
release.

This changes the default value of the 'enable_revocation_notifications'
to false. 

Note:  The default value used by the python configuration upgrade script needs to be updated accordingly.


keylime-agent/src/config.rs
```diff
@@ -828,6 +828,7 @@ mod tests {
     fn get_revocation_notification_ip_empty() {
         let mut test_config = KeylimeConfig {
             agent: AgentConfig {
+                enable_revocation_notifications: true,
```
aplanas, line 831: I wonder why this is here ... in as `true`

ansasaki: This is to test certain required configuration options when the option is enabled. For that we need to enable the option.

aplanas (approved):

THS-on (approved):

PR #525, opened 2023-02-20
ima-emulator: Implement graceful shutdown

ansasaki:
This adds signal-hook as a dependency for the IMA emulator.  The strategy is similar to what is done on the keylime agent: a dedicated thread is spawned to stop the main task on Ctrl + C (SIGINT).


keylime-ima-emulator/src/main.rs
```diff
@@ -209,25 +215,51 @@ fn main() -> std::result::Result<(), ImaEmulatorError> {
...
+                    ima_hash_alg,
+                    *pcr_hash_alg,
+                    None,
+                ).expect("Error extending position {position} on PCR bank {pcr_hash_alg}");
+            }
+
+            // FIXME: We could poll IMA_ML as in the python implementation, though
+            // the file is not pollable:
+            // https://github.com/torvalds/linux/blob/master/security/integrity/ima/ima_fs.c#L267
+            // Better idea might be to check the "runtime_measurements_count" file.
```
aplanas, line 242: Maybe this can be used? https://crates.io/crates/signal-hook

If I read the code it is not using an active loop like in this solution, and the example seems really similar to the current approach.

ansasaki: I can try, but I'm afraid it will fall into the case mentioned in the comment:
```
        // Do some time-limited stuff here
        // (if this could block forever, then there's no guarantee the signal will have any
        // effect).
```

ansasaki: Well, apparently it works! I'll push the (much simpler) changes.

aplanas: IIRC in POSIX a singal is not a thread, is a callback that needs to be re-registered.

aplanas (approved):

ansasaki:
The issues on Fedora Rawhide can be related with a regression:
https://github.com/tpm2-software/tpm2-tools/issues/3210

The other issue is again the conflict between i686 and x86_64 packages that for some reason are installed together.

PR #526, opened 2023-02-21
Bump version to 0.2.0

ansasaki:
Bump version to 0.2.0 as preparation for tag/release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1109,7 +1109,7 @@ dependencies = [
 
 [[package]]
 name = "keylime"
-version = "0.1.0"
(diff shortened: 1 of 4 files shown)


aplanas (approved):

ansasaki:
Failures on Rawhide are due to a regression on `tpm2-tools` (https://github.com/tpm2-software/tpm2-tools/issues/3210).
Failures on Fedora 36 are caused by old `tpm2-tools` version installed (note that `tpm2-tools 5.4.1` is available on Fedora 36, but the test system is still using `5.2.2`)

PR #528, opened 2023-02-23
config: Skip validation of keylime_dir during tests

ansasaki:
When running tests on machines where `/var/lib/keylime` does not exist (e.g. build machines where keylime is not installed), the tests would fail on the check if keylime_dir exists.

### DIFF
--- a/keylime-agent/src/config.rs
+++ b/keylime-agent/src/config.rs
@@ -608,51 +608,54 @@ fn config_translate_keywords(
         Some(ref dir) => {
             if dir.is_empty() {
                 match &config.agent.keylime_dir {
-                    s => s.clone(),
-                    _ => DEFAULT_KEYLIME_DIR.to_string(),
+                    s => Path::new(s),
+                    _ => Path::new(DEFAULT_KEYLIME_DIR),
                 }
             } else {
-                dir.to_string()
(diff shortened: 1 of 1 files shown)


THS-on (approved):

keylime-agent/src/config.rs
```diff
@@ -608,23 +608,26 @@ fn config_translate_keywords(
...
-            s => s.clone(),
-            _ => DEFAULT_KEYLIME_DIR.to_string(),
+            s => Path::new(s),
+            _ => Path::new(DEFAULT_KEYLIME_DIR),
         },
     };
 
     // Validate that keylime_dir exists
-    let keylime_dir = Path::new(&keylime_dir).canonicalize().map_err(|e| {
+    #[cfg(not(test))]
```
aplanas, line 625: Ok, this should be the fix. Why the type change from String to Path?

aplanas (approved):
LGTM, just a question

keylime-agent/src/config.rs
ansasaki: In the way the code was written before, there was a useless clone of a `String` when we can use the reference `&str` directly to create the `Path`. And the functions below expect a `Path` and not a `String`.

PR #531, opened 2023-03-01
IGNORE: Remove info call due to checking functionality of code coverage

Koncpa:
Just checking if code coverage measurement works properly.


kkaarreell:
@Koncpa try to rebase against the current master to see if the change in required files have any effect.

PR #533, opened 2023-03-01
Disable e2e on Rawhide due to RHBZ#2171376

kkaarreell:


ansasaki (approved):
It is sad that we have to disable tests on Rawhide again, but I agree. We can re-enable when the problem is fixed

PR #536, opened 2023-03-15
Add an example of IMA policy

aplanas:
Add a SELinux extended IMA policy as an example, and some documentation about how to use and extend it.


aplanas:
This IMA policy is using the SELinux types available in openSUSE, but this is the same one that is used in Fedora. Both are instances of the SELinux reference policy, so the types referenced on it are quite generic.

ima-policy/README
```diff
@@ -0,0 +1,55 @@
...
+  security_file_type
+  setfiles_domain
+  spoolfile
+  svirt_file_type
+  systemd_unit_file_type
+  tmpfile
+  tmpfsfile
+
+Special mention to non_auth_file_type and non_security_file_type
+(among other liske logfile or tmpfile), that should cover the most
```
ueno, line 45: "among other liske" → "among others like"?

ima-policy/README
```diff
@@ -0,0 +1,55 @@
...
+To use this policy, we need to copy it in "/etc/ima/ima-policy" and
+systemd will load it after the SELinux policy has been loaded.
+
+For this example, we used the initial set of SELinux attributes, that
+group the file types under categories.  From that list we selected
+some of those attribute to deep more into the types that can be relevant for the IMA policy:
+
+  seinfo -a
+
+The current selection cover full or partially the types under those
```
ueno, line 19: "cover full or partially" → "covers fully or partially"?

ima-policy/ima-policy
```diff
@@ -0,0 +1,1046 @@
...
+dont_measure obj_type=etc_runtime_t
+dont_measure obj_type=system_conf_t
+dont_measure obj_type=system_db_t
+dont_measure obj_type=tmp_t
+dont_measure obj_type=var_lib_t
+dont_measure obj_type=var_lock_t
+dont_measure obj_type=var_run_t
+dont_measure obj_type=var_spool_t
+
+# for i in $(seinfo -alockfile -x | grep _t | tr '[:space:]' " "); do echo "dont_measure obj_type=$i"; done
```
ueno, line 39: nit: it would be a little more robuster to limit the grep pattern as `_t$`? also it could be simply written with a single sed invocation: `seinfo -alockfile -x | sed -n 's/^\t\(.*_t\)$/dont_measure obj_type=\1/p'`.

ueno (approved):

ima-policy/README
aplanas: Fixed!

ima-policy/README
aplanas: yes, fixed.

THS-on:
We also have some IMA policies here: https://github.com/keylime/keylime/tree/master/demo/ima-policies

Maybe we should unify them in one place and then point the documentation to it.

aplanas:
> Maybe we should unify them in one place and then point the documentation to it.

From a brief comparison, seems that the overlayfs fs ID was missing, and I added it now.

PR #546, opened 2023-03-28
Build Fedora packages on Copr using packit

ansasaki:
Build Fedora packages on Copr when new commits are pushed to master branch.

This includes a specfile similar to one used downstream on Fedora, and a patch to make the build to use the dependencies version that are available on Fedora.

EDIT: For CentOS Stream I need to do some extra steps as it needs all dependencies to be vendored.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,3 +1,12 @@
+downstream_package_name: keylime-agent-rust
+upstream_project_url: https://github.com/keylime/rust-keylime
+
+srpm_build_deps:
+  - cargo
+  - rust
+  - git
+  - openssl-devel
+
 jobs:
 - job: tests
   trigger: pull_request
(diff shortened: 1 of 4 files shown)


aplanas (approved):
LGTM.

Maybe it time to archive this:
  https://github.com/keylime/keylime-rpm
  https://github.com/keylime/keylime-deb

THS-on:
@ansasaki it looks like that the Fedora builds are failing because the rust tss bindings cannot be found:
```
Error: 
 Problem 1: nothing provides requested (crate(tss-esapi/default) >= 7.2.0 with crate(tss-esapi/default) < 8.0.0~)
 Problem 2: nothing provides requested (crate(tss-esapi/generate-bindings) >= 7.2.0 with crate(tss-esapi/generate-bindings) < 8.0.0~)
```
Full log: https://download.copr.fedorainfracloud.org/results/packit/keylime-rust-keylime-546/fedora-36-x86_64/05718845-keylime-agent-rust/builder-live.log.gz

@aplanas agree they are way out-of-date and probably not really useful anymore. Can you raise that in the Slack channel?

ansasaki:
> @ansasaki it looks like that the Fedora builds are failing because the rust tss bindings cannot be found:
> 
> ```
> Error: 
>  Problem 1: nothing provides requested (crate(tss-esapi/default) >= 7.2.0 with crate(tss-esapi/default) < 8.0.0~)
>  Problem 2: nothing provides requested (crate(tss-esapi/generate-bindings) >= 7.2.0 with crate(tss-esapi/generate-bindings) < 8.0.0~)
> ```
> 
> Full log: https://download.copr.fedorainfracloud.org/results/packit/keylime-rust-keylime-546/fedora-36-x86_64/05718845-keylime-agent-rust/builder-live.log.gz

Yes, I'm aware, but these packages are not maintained by us. Basically all attempts to update the rust agent on Fedora was blocked by a problem on their side (or on `tpm2-tools` which is also not maintained by us).

A workaround to make it to build would be to use vendored versions of all dependencies on Fedora too (like how it is done in RHEL/CentOS), but we would loose good insight on the state of the dependencies on Fedora.

I think I'll create exceptions for these older Fedora versions to make them work with vendored dependencies. From Fedora 38 onwards I'll keep using the system dependencies.

ansasaki:
The RPM's were built successfully using the vendored dependencies. I'll rebase, drop the temporary commit, and merge.

PR #554, opened 2023-04-04
Allow setting binary measured boot log path on RPM binaries

### LINKED ISSUE #490: Make measured boot event log location configurable
Currently, the measured boot event log location is hard-coded and cannot be changed through configuration options nor environment variables.

Make the measured boot event log location configurable.

ansasaki:
Allow setting the path to the binary bios measurements log file via the `TPM_BINARY_MEASUREMENTS` environment variable when build with the `testing` feature enabled.

The binaries in the RPM are build with the `testing` feature enabled.

Fixes #490

### DIFF
--- a/rpm/fedora/keylime-agent-rust.spec
+++ b/rpm/fedora/keylime-agent-rust.spec
@@ -15,7 +15,7 @@
 %global __brp_mangle_shebangs_exclude_from ^/usr/src/debug/.*$
 
 Name:           keylime-agent-rust
-Version:        v0.2.0
+Version:        0.2.0
 Release:        %{?autorelease}%{!?autorelease:1%{?dist}}
 Summary:        Rust agent for Keylime
 
@@ -38,7 +38,7 @@ Summary:        Rust agent for Keylime
 #
 License:        ASL 2.0 and BSD and MIT
 URL:            https://github.com/keylime/rust-keylime/
-Source0:        rust-keylime-v0.2.0.tar.gz
+Source0:        rust-keylime-v%{version}.tar.gz
 # The vendor tarball is created using cargo-vendor-filterer to remove Windows
 # related files (https://github.com/cgwalters/cargo-vendor-filterer)
 #   tar xf rust-keylime-%%{version}.tar.gz
@@ -102,7 +102,7 @@ EOF
 %endif
 
 %build
-%cargo_build
+%cargo_build -ftesting
 
 %install
 
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,5 +1,9 @@
 downstream_package_name: keylime-agent-rust
 upstream_project_url: https://github.com/keylime/rust-keylime
+specfile_path: rpm/fedora/keylime-agent-rust.spec
+actions:
+  get-current-version:
+  - bash -c "git describe --tags --abbrev=0 | sed 's/v\(.*\)/\1/g'"
 
 srpm_build_deps:
   - cargo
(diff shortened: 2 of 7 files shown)


rpm/fedora/rust-keylime-allow-setting-mb-log-path.patch
```diff
@@ -0,0 +1,41 @@
+diff --git a/keylime-agent/src/main.rs b/keylime-agent/src/main.rs
```
aplanas, line 1: It is the same patch AFAICS, maybe you can make it a link, of put them in `/rpm/patch/` and make both links to it.

aplanas (commented):
I am a bit concern that this rpm/ can confuse later. It is to generate RPMs for the testing farm, and somehow unrelated with the project.

IMHO all this RPM code should be living in the test side and not in the project side, as in any case the testing farm is not really testing the released fedora nor centos package.

If this is not possible, at least we should add a README under rpm/ to clarify that "If you are looking for the reference spec file, do not look here"

rpm/fedora/keylime-agent-rust.spec
```diff
@@ -102,7 +104,7 @@ EOF
 %endif
 
 %build
-%cargo_build
+%cargo_build -ftesting
```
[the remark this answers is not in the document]
ansasaki: AFAIK, the macro from `rust-packaging` in Fedora does not support setting `--features=testing`. The parser expects only `-f`. See https://pagure.io/fedora-rust/rust-packaging/blob/main/f/macros.d/macros.cargo#_104

rpm/fedora/keylime-agent-rust.spec
```diff
@@ -81,7 +83,7 @@ Conflicts:      keylime-agent
 Rust agent for Keylime
 
 %prep
-%autosetup -n rust-keylime-%{version} -N
+%autosetup -n rust-keylime-v0.2.0 -N
```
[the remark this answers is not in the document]
ansasaki: I accidentally added the changes made by packit locally. I'll fix this

rpm/fedora/rust-keylime-allow-setting-mb-log-path.patch
ansasaki: OK, I can do this.

ansasaki:
> I am a bit concern that this rpm/ can confuse later. It is to generate RPMs for the testing farm, and somehow unrelated with the project.
> 
> IMHO all this RPM code should be living in the test side and not in the project side, as in any case the testing farm is not really testing the released fedora nor centos package.
> 
> If this is not possible, at least we should add a README under rpm/ to clarify that "If you are looking for the reference spec file, do not look here"

I agree that it may be confusing at the moment, but the goal in future is to maintain the specfile here and automate downstream maintenance as well, so it will become the "reference specfile".

For now, the rpms will be used to avoid recompiling the agent for testing both in this repo, as well as in `keylime/keylime`. It makes easier for testing farm to have the rpm packages in Copr.

I'll add a commit to explain this in README.

rpm/fedora/keylime-agent-rust.spec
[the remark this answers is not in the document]
ansasaki: Done

rpm/fedora/rust-keylime-allow-setting-mb-log-path.patch
ansasaki: In the end, I think it is safe to make the code part of the project instead of bringing as a separate patch.

aplanas: Yeah ... that is a sensible option

aplanas (approved):
Seems that the Centos package fails for some GPG key. Not sure why, but the changes LGTM

ansasaki:
I added a small change to make the RPMS more usable: packit was picking the version from the tag (v.0.2.0), which made RPM to not consider it an update because `keylime-0.1.0 > keylime-v0.2.0` (note the `v` prepending the version number). I had to manually change the version to result in `keylime-0.2.0` for the RPM.

I'll drop the temporary commit and push the changes.

PR #565, opened 2023-04-14
packit: Build RPM packages on Copr for fedora-all

ansasaki:
Enable building RPM packages on Copr for all active Fedora versions

On Fedora-38 and current Rawhide, it is not possible to build the keylime agent due to the issues:
-  https://github.com/tpm2-software/tpm2-tools/issues/3210
   - `tpm2-tools` is currently broken when build with `-lto -DFORTIFY_SOURCE=3`
- The `rust-base64` package version `0.13`, required to build `rust-tss-esapi` is not available.

This makes the RPM build to use vendored dependencies for all Fedora versions.

PR #568, opened 2023-04-21
rpm/fedora: Fix metadata patch

ansasaki:
With the changes in the Cargo.toml, the patch for the rpm also needs to
be fixed


aplanas (approved):

ansasaki:
The build on Fedora 38 is expected to fail due to the missing `rust-base64` package version `0.13`. Note that without #565 the dependencies used for Fedora >= 38 are the system installed dependencies.

The tests on Fedora 38 are expected to fail. The issue is fixed by https://bodhi.fedoraproject.org/updates/FEDORA-2023-c907a05c30, which should reach stable tomorrow.

I'll drop the temporary commit and merge.

PR #571, opened 2023-04-24
keylime-agent/src/common.rs: remove VTPM and IMA stub variables

hustliyilin:
https://github.com/THS-on/keylime/commit/48eed7436d2eaaef003b53fe37cb326e60d2612a removeed TPM, VTPM and IMA stubbing support. So we need to remove the VTPM and IMA stub variables in the rust-keylime


ansasaki (approved):
LGTM! Thank you!

PR #575, opened 2023-04-25
Cargo: Update base64 to version 0.21

ansasaki:
There were incompatible changes between version `0.13` and `0.21`, which requires some changes in the code.

Supersedes #570

### DIFF
--- a/keylime-agent/src/keys_handler.rs
+++ b/keylime-agent/src/keys_handler.rs
@@ -12,6 +12,7 @@ use crate::{
     Error, QuoteData, Result,
 };
 use actix_web::{web, HttpRequest, HttpResponse, Responder};
+use base64::{engine::general_purpose, Engine as _};
 use log::*;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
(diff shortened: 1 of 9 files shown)


keylime-agent/src/keys_handler.rs
```diff
@@ -865,9 +873,9 @@ mod tests {
             rsa_oaep_encrypt(&quotedata.pub_key, u.as_ref()).unwrap(); //#[allow_ci]
 
         let ukey = KeylimeUKey {
-            encrypted_key: base64::encode(&encrypted_key),
+            encrypted_key: general_purpose::STANDARD.encode(&encrypted_key),
             auth_tag: hex::encode(auth_tag),
-            payload: payload.map(base64::encode),
+            payload: payload.map(general_purpose::STANDARD.encode),
```
aplanas, line 878: There is a complain here, should be something like `payload.map(|p| general_purpose::STANDARD.encode(p.as_ref())),`, as before?

ansasaki: Yes, probably that will be the way to go. Thanks for the suggestion!

aplanas (approved):

THS-on (approved):

PR #582, opened 2023-05-03
build(deps): bump all the dependencies

liushuyu:
This commit bumps all the dependencies to eliminate potential security issues.

If you think this pull request is too large to review adequately, I am happy to split it into smaller ones.


liushuyu:
Note: This also bumps the MSRV (minimum supported Rust version) from 1.60 to 1.65.

If you think this is unacceptable, I am also happy to accommodate the MSRV requirements you currently have.

ansasaki:
Thank you for your contribution, but we follow some rules to update the dependencies, for example, that the dependencies must be available at least in Fedora Rawhide, but preferably also in Fedora stable.

This is to avoid being unable to build the agent in Fedora, as it happened in the past. Thank you for understanding.

I'll close this PR now.

PR #584, opened 2023-05-11
docker/fedora/keylime_rust.Dockerfile: add the logic of cloning and compiling rust-keylime

hustliyilin:
Obviously, the codes of the agent are located in the rust-keylime repository. However, there is no logic for cloning and compiling rust-keylime in the corresponding rust dockerfile. So we need to add it.

Fixes: #585 

### DIFF
--- a/docker/fedora/keylime_rust.Dockerfile
+++ b/docker/fedora/keylime_rust.Dockerfile
@@ -10,7 +10,7 @@ LABEL version="2.0.1" description="Keylime - Bootstrapping and Maintaining Trust
 
 # environment variables
 ARG BRANCH=master
-ENV KEYLIME_HOME ${HOME}/keylime
+ENV RUST_KEYLIME_HOME ${HOME}/rust-keylime
 ENV container docker
 COPY dbus-policy.conf /etc/dbus-1/system.d/
 
@@ -35,6 +35,7 @@ glib2-static \
 gnulib \
 kmod \
 llvm llvm-devel \
+libarchive-devel \
 libselinux-python3 \
 libtool \
 libtpms \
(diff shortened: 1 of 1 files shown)


docker/fedora/keylime_rust.Dockerfile
```diff
@@ -62,3 +64,10 @@ WORKDIR ${KEYLIME_HOME}
 RUN git clone https://github.com/keylime/keylime.git && \
 cd keylime && \
 cp keylime.conf /etc/keylime.conf
+
+# clone and build rust-keylime
+WORKDIR ${RUST_KEYLIME_HOME}
+RUN git clone https://github.com/keylime/rust-keylime.git && \
+cd rust-keylime && \
+make && \
+make install
```
ansasaki, line 73: These changes will result in a huge layer in the resulting container as all intermediary compilation artifacts would be included in it (I mean many GB). Consider removing the artifacts after the installation (e.g. running `cargo clean`)

hustliyilin:
Thanks @ansasaki . After adding the `cargo clean` command, the docker image size reduce about 1.6GB.

docker/fedora/keylime_rust.Dockerfile
```diff
@@ -62,3 +64,11 @@ WORKDIR ${KEYLIME_HOME}
 RUN git clone https://github.com/keylime/keylime.git && \
 cd keylime && \
 cp keylime.conf /etc/keylime.conf
```
ansasaki, line 66: Could you please remove this? It is not necessary anymore as we install the agent configuration in the `make install` step.

ansasaki (changes requested):

hustliyilin:
Thanks @ansasaki . I fixed it and  force-pushed the codes. Could you review it again?

PR #588, opened 2023-05-15
keylime-agent/src: update API version to 2.1 to consistent with https//github.com/keylime/keylime/blob/master/docs/rest_apis.rst

### LINKED ISSUE #589: Don't support the RESTful API v2.1 in https://github.com/keylime/keylime/blob/master/docs/rest_apis.rst
Currently, the rust-keylime only supports RESTful API v2.0. When using the https://github.com/keylime/keylime/blob/master/docs/rest_apis.rst, it will be an error.

image

hustliyilin:
Fixes #589 

### DIFF
--- a/keylime-agent/src/common.rs
+++ b/keylime-agent/src/common.rs
@@ -35,7 +35,7 @@ use tss_esapi::{
 /*
  * Constants and static variables
  */
-pub const API_VERSION: &str = "v2.0";
+pub const API_VERSION: &str = "v2.1";
 pub const TPM_DATA_PCR: usize = 16;
 pub const IMA_PCR: usize = 10;
 pub static RSA_PUBLICKEY_EXPORTABLE: &str = "rsa placeholder";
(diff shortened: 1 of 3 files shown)


aplanas (changes requested):
In the 2.1 changelog I can see that they removed the vmask parameter, but I still can see some references in the quotes_handler.rs test code (and some comments).

The other change seems to affect the verifier, so I am not sure that has an impact here.

@ansasaki do you know if beside the quotes_handlers change, more changes are required?

hustliyilin:
Thanks @aplanas.   Currently, the rust-keylime only supports RESTful API v2.0. When using the https://github.com/keylime/keylime/blob/master/docs/rest_apis.rst, it will be an error like #588 .

image

So we need to update the API version in the rust-keylime agent or add corresponding instructions to the corresponding REST API documentation.  I'm not sure which way is better.

THS-on:
@aplanas yeah the only change for the agent should be that we no longer send the vmask parameter.

hustliyilin:
> @aplanas yeah the only change for the agent should be that we no longer send the vmask parameter.

Thanks @THS-on . @aplanas Could you review this PR again?

aplanas (approved):

ansasaki (approved):
LGTM!

PR #595, opened 2023-05-25
Remove MOUNT_SECURE bool

Koncpa:
Remove MOUNT_SECURE bool and conditional
related with bool.Update secure mount
unit test to mount temp dir.

### DIFF
--- a/keylime-agent/src/secure_mount.rs
+++ b/keylime-agent/src/secure_mount.rs
@@ -81,23 +81,6 @@ fn check_mount(secure_dir: &Path) -> Result<bool> {
  * functions are unsafe function in Rust to use.
  */
 pub(crate) fn mount(work_dir: &Path, secure_size: &str) -> Result<PathBuf> {
-    // Use /tmpfs-dev directory if MOUNT_SECURE flag is not set. This
-    // is for development environment and does not mount to the system.
-    if !MOUNT_SECURE {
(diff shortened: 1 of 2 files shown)


keylime-agent/src/secure_mount.rs
```diff
@@ -163,11 +146,9 @@ mod tests {
 
     #[test]
     fn test_secure_mount() {
-        let path = "/var/lib/keylime";
-        let work_dir = Path::new(&path);
-        let secure_dir_path = Path::new(work_dir).join("secure");
+        let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
```
ansasaki, line 149: I suggest you keep the `"secure"` to force `mount()` to create the directory. It will increase the coverage. Something like:

```suggestion
        let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
        let secure_dir_path = Path::new(temp_workdir).join("secure");
```

ansasaki (commented):
Could you please also remove the MOUNT_SECURE from `common.rs`?

keylime-agent/src/secure_mount.rs
Koncpa: Actually it seems that this decreased code coverage a lot from point of secure mount.

ansasaki (approved):
LGTM!

PR #601, opened 2023-06-07
Adding release Dockerfiles

mheese:
This PR aims to add a Dockerfile for a release version of the rust keylime agent.

**This is a work-in-progress at this point.** It is being discussed in the keylime-operator slack channel.

It currently contains a local build script like in the main keylime repository. Also, it contains different versions which have their advantages/disadvantages that we should discuss:
- `Dockerfile.distroless`:  an absolute bare minimum container which does not even have a shell. It is based on the "distroless" container effort from Google (which for example forms the base images for a couple of bigger projects like Kubernetes)
- `Dockerfile.wolfi`: also a bare minimum container which does not even have a shell. It is based on the wolfi container distribution from Chainguard which in turn was inspired by Google's distroless and ko projects. Its focus is a dedicated container distribution which heavily focuses on security, frequent updates and updates faster/more often than distroless
- `Dockerfile.fedora`: is based on Fedora 38. This is a "fat" image, but both the build stage as well as the assembly stage are based upon Fedora 38. This is probably closer to what we know works.

Here a comparison of the two versions:

## Dockerfile.distroless

- builder stage based on official Rust Docker images
- official Rust image (which is chosen here) is based on Debian 11
- has an additional custom compilation/installation stage for libarchive in the builder stage with minimal options for libarchive (which are enough for our purposes)
- has an additional compilation/installation stage for tpm2-tss with minimal options which _should_ be enough for our purposes (fapi and fapi policy libs are disabled)
- the assembly stage uses `gcr.io/distroless/cc` as a base image which is a minimal no-shell base image including C runtime dependencies (libc, libgcc, etc.) as well as OpenSSL and root CA certificates. This image is intended for use by languages like Rust or D according to its [documentation](https://github.com/GoogleContainerTools/distroless/blob/main/cc/README.md)
- the distroless images are all based on Debian 11 at this point in time, so this makes the build and assembly compatible
- dependencies and compiled agent are copied from the builder stage

Pros of this approach are IMHO:
- maximized for security as it truly does not contain what is not needed
- image size is quite good (still bigger than I expected though)
- good control over used versions of dependencies and rust version etc.

Cons of this approach are IMHO:
- based on Debian 11 which uses oldish versions of things - probably not an issue at all though, and it does have all security patches after all
- OpenSSL version is 1.1 here - again, not sure if this is an issue for us
- harder to debug (because there is no shell)

image size:

```
REPOSITORY                                                         TAG       IMAGE ID       CREATED          SIZE
keylime_agent                                                      latest    3ce6c74b8a73   22 minutes ago   57.6MB
```

## Dockerfile.wolfi

- builder stage based on wolfi-base docker images
- rust is being installed through wolfi which is the latest version - unfortunately Chainguard's rust image does not work because it does not allow to add dependencies
- has an additional compilation/installation stage for tpm2-tss with minimal options which _should_ be enough for our purposes (fapi and fapi policy libs are disabled)
- the assembly stage uses `cgr.dev/chainguard/cc-dynamic` as a base image which is a minimal no-shell base image including C runtime dependencies (libc, libgcc, etc.). This image is intended for use by languages like Rust even according to Chainguard's rust image [documentation](https://edu.chainguard.dev/chainguard/chainguard-images/reference/rust/overview/#application-setup-for-end-users)
- both builder and assembly stage images use the same base, so there is no potential incompatibilities between binaries/libraries
- dependencies and compiled agent are copied from the builder stage

Pros of this approach are IMHO:
- maximized for security as it truly does not contain what is not needed
- even better than distroless in regards to security IMHO
- image size is quite good (still bigger than I expected though)
- good control over used versions of dependencies and rust version etc.

Cons of this approach are IMHO:
- Chainguard is a relatively new player in this game (not sure that matters, just potentially for long-term maintainability)
- harder to debug (because there is no shell)

image size:

```
REPOSITORY                                                         TAG                        IMAGE ID       CREATED          SIZE
keylime_agent                                                      latest                     367bada26999   6 minutes ago    42.4MB
```

## Dockerfile.fedora

- builder stage uses Fedora 38 like the existing docker image for testing
- all build dependencies are installed through Fedora packages
- runtime dependencies are installed in the assembly stage (as compared to copied)

Pros of this approach are IMHO:
- probably the closest to what we know works
- easier to debug (has shell and tools, etc.)

Cons of this approach are IMHO:
- image size is "fat" for an agent which should be very small
- still contains a lot of tools and things (including package manager) which aren't necessary and are just an increase in attack surface
- even the keylime agent has a lot of library dependencies because of its C library dependencies that are for sure not required at all and are solely there to satisfy linker requirements

image size:

```
REPOSITORY                                                         TAG       IMAGE ID       CREATED             SIZE
keylime_agent                                                      latest    cbf68e330487   17 seconds ago      203MB
```

### DIFF
--- a/docker/release/Dockerfile.distroless
+++ b/docker/release/Dockerfile.distroless
@@ -1,0 +1,102 @@
+# SPDX-License-Identifier: Apache-2.0
+# the builder stage which uses the official rust base image (based on Debian 11) to build the keylime agent
+FROM rust:1.70-bullseye AS builder
+
+# we are using the "generate-bindings" feature for the tss-esapi crate which requires clang/llvm
+RUN apt-get update && apt-get install -y --no-install-recommends clang llvm
+
+# Install tpm2-tss (dependency for the tss-esapi crate)
+WORKDIR /src
+RUN wget https://github.com/tpm2-software/tpm2-tss/releases/download/4.0.1/tpm2-tss-4.0.1.tar.gz
+RUN tar xf tpm2-tss-4.0.1.tar.gz
+WORKDIR /src/tpm2-tss-4.0.1
+RUN ./configure \
+    --prefix=/usr \
+    --disable-static \
+    --disable-fapi \
+    --disable-policy \
+    --disable-doxygen-doc \
+    --disable-defaultflags
+RUN make
+RUN make install
+
+# Install libarchive (dependency for the compress-tools crate) - we need only a minimum feature set here
+WORKDIR /src
+RUN wget https://github.com/libarchive/libarchive/releases/download/v3.6.2/libarchive-3.6.2.tar.gz
+RUN tar xf libarchive-3.6.2.tar.gz
+WORKDIR /src/libarchive-3.6.2
+RUN ./configure \
+    --prefix=/usr \
+    --with-openssl \
+    --without-mbedtls \
+    --without-nettle \
+    --without-xml2 \
+    --without-expat \
+    --disable-static
+RUN make
+RUN make install
+# there is a bug in the libarchive.pc file which wrongly adds iconv
+RUN sed -i "s/iconv //" /usr/lib/pkgconfig/libarchive.pc
+
+# build rust-keylime
+COPY . /src/rust-keylime/
+WORKDIR /src/rust-keylime
+RUN make RELEASE=1 TARGETDIR=target target/release/keylime_agent
+
+# truly just for debugging purposes for the assembly stage
+RUN readelf -W \
+    --file-header --program-headers --sections --dynamic --notes --version-info --arch-specific --unwind --section-groups --histogram \
+    /src/rust-keylime/target/release/keylime_agent
+RUN ldd /src/rust-keylime/target/release/keylime_agent
+
+# now assemble a release docker image using a minimal docker image
+FROM gcr.io/distroless/cc-debian11:latest
+ARG VERSION=latest
+LABEL org.opencontainers.image.authors="Keylime Team <main@keylime.groups.io>"
+LABEL org.opencontainers.image.version="$VERSION"
+LABEL org.opencontainers.image.title="Keylime Agent"
+LABEL org.opencontainers.image.description="Keylime Agent - Bootstrapping and Maintaining Trust in the Cloud"
+LABEL org.opencontainers.image.url="https://keylime.dev/"
+LABEL org.opencontainers.image.source="https://github.com/keylime/rust-keylime/"
+LABEL org.opencontainers.image.licenses="Apache-2.0"
+LABEL org.opencontainers.image.vendor="The Keylime Authors"
+
+# Copy all agent dependencies from the builder image
+# NOTE: the cc base image comes with all C runtime dependencies (libc, libm, libgcc, etc.), so no need to copy those
+# TODO: Unfortunately the COPY directive is following links and not preserving the link file. This slightly bloats the image.
+
+# libarchive is a direct dependency for the compress-tools crate, so we must copy itself and all its dependencies
+COPY --from=builder \
+    /usr/lib/libarchive.so* \
+    /lib/x86_64-linux-gnu/liblzma.so* \
+    /lib/x86_64-linux-gnu/libbz2.so* \
+    /lib/x86_64-linux-gnu/libz.so* \
+    /usr/lib/x86_64-linux-gnu/
+# tpm2-tss libraries are a dependency (probably not all of them, but we just copy all)
+# because we are using the tss-esapi crate which is essentially just a wrapper around those (unfortunately)
+COPY --from=builder \
+    /usr/lib/libtss2*.so* \
+    /usr/lib/x86_64-linux-gnu/
+
+# now copy the agent from the builder
+COPY --from=builder /src/rust-keylime/target/release/keylime_agent /bin/keylime_agent
+COPY --from=builder /src/rust-keylime/keylime-agent.conf /etc/keylime/keylime-agent.conf
+ENTRYPOINT ["/bin/keylime_agent"]
+
+# we default the log level to info if not overwritten
+ENV RUST_LOG=keylime_agent=info
+
+# the agent currently listens on this port by default
+# it's good practice to declare this in the Dockerfile
+EXPOSE 9002/tcp
+
+# define the stopsignal to be SIGINT like the systemd service file does
+STOPSIGNAL SIGINT
+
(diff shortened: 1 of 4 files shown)


aplanas:
Very nice. Some time ago I did one for openSUSE:

https://build.opensuse.org/package/show/devel:microos:containers/rust-keylime-image

Maybe we can find some common rules

mheese:
> Very nice. Some time ago I did one for openSUSE:
> 
> https://build.opensuse.org/package/show/devel:microos:containers/rust-keylime-image
> 
> Maybe we can find some common rules

Nice! I like the podman labels and volumes, healthcheck and stopsignal definitions. As I was writing this with Kubernetes in my mind, I did not consider any of these. As they are not mutually exclusive though, they should be included imho. I'll try to incorporate those.

ansasaki (commented):
I suggested some changes to reduce the fedora-based image, basically using `fedora-minimal` as the base image.
With this change, I get:
```
REPOSITORY                                                  TAG               IMAGE ID      CREATED         SIZE
localhost/keylime_agent                                     latest            9c9c97702d63  17 minutes ago  124 MB
```

docker/release/Dockerfile.fedora
```diff
@@ -0,0 +1,69 @@
+# the builder stage which uses the latest Fedora image to build the keylime agent - we know this works well
+FROM fedora:38 AS builder
```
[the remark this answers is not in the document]
mheese: why is that important for the builder? For the resulting image I see why that makes sense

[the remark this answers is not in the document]
ansasaki: It is not important, It is just to speed-up installation a bit and reduce clock ticks wasting for the sake of the planet :)

mheese:
> This is great stuff! When I put together [my Dockerfiles](https://github.com/hse-aurora/keylime-ci/tree/main/docker), I had a slightly different use case in mind as my goal was primarily to support our internal KL development efforts at HPE. Distroless release images were at the back of my mind as a stretch goal that I just never got around to. So I am really happy to see this PR!

thanks! :+1: 

> 
> I've had a look through the files and had some notes/queries:
> 
>     * I noticed that swtpm is getting installed as a dependency in `Dockerfile.fedora` but does not seem to get copied to the release image in the assembly stage. Is this needed to build the agent? (I don't see this dependency in the distroless `Dockerfile`.)
> 
>     * Similarly, I am not sure if tpm2-tools and tpm2-abrmd are necessary.
> 

I'm pretty sure they can be removed (actually very certain). The base for the Fedora builder stage was copy&pasted from the existing Dockerfile. I have to admit that I did not carefully review everything that gets installed there. Good catch!

>     * It might be good if the files could copy the [default config file](https://github.com/keylime/rust-keylime/blob/master/keylime-agent.conf) over to `/etc/keylime`. Doing so would not preclude the user from supplying their own config file by mounting a volume, or overwriting the defaults with environment variables.
> 

Yes, I like that. Good point!

>     * You mention that the Google distroless images are based on the previous Debian release. Might you consider using Chainguard images instead? They use a custom distro which is supposed to get security updates faster than other images. They've published a blog post discussing their differences with Google's distroless effort [here](https://dev.to/chainguard/container-images-for-the-cloud-native-era-2mk1). And their Rust image is [here](https://edu.chainguard.dev/chainguard/chainguard-images/reference/rust/overview/), if you are curious.
> 

Tbh, I never even heard of Chainguard/Wolfi before. I will definitely check it out. The idea here was to propose several files and discuss it with the community. Also, it's not uncommon for some projects to provide multiple tags with different base images.

What I like about the distroless images is that they are "as open as it gets" from an Open Source perspective. And they are well respected as base images in the Kubernetes community.

From a quick look at the links you provided it is definitely worth looking into Chainguard/Wolfi and also have an option for that. I like the SBOMs and cosign/sigstore signatures for sure. It's very much aligned with what we are doing at Hedgehog ourselves.

>     * I like that your build script passes the version through as build arg instead of the template approach taken by the main repo.

Yeah, no need to template it just because of that.

mheese:
The images now work like this tested with the Helm charts in: https://github.com/keylime/attestation-operator/pull/13

The Rust Static code check currently fails with a security vulnerability check problem: openssl needs to be upgraded!

Should I do this just in this PR, or create a separate one for that?

ansasaki:
> The images now work like this tested with the Helm charts in: [keylime/attestation-operator#13](https://github.com/keylime/attestation-operator/pull/13)
> 
> The Rust Static code check currently fails with a security vulnerability check problem: openssl needs to be upgraded!
> 
> Should I do this just in this PR, or create a separate one for that?

There is one open already by dependabot: https://github.com/keylime/rust-keylime/pull/605

docker/release/Dockerfile.fedora
```diff
@@ -0,0 +1,71 @@
...
+
+# we default the log level to info if not overwritten
+ENV RUST_LOG=keylime_agent=info
+
+# the agent currently listens on this port by default
+# it's good practice to declare this in the Dockerfile
+EXPOSE 9002
+
+# define the stopsignal to be SIGINT like the systemd service file does
+STOPSIGNAL SIGINT
```
ansasaki, line 63: With #613, this should not be necessary anymore

docker/release/Dockerfile.wolfi
```diff
@@ -0,0 +1,113 @@
...
+
+# we default the log level to info if not overwritten
+ENV RUST_LOG=keylime_agent=info
+
+# the agent currently listens on this port by default
+# it's good practice to declare this in the Dockerfile
+EXPOSE 9002/tcp
+
+# define the stopsignal to be SIGINT like the systemd service file does
+STOPSIGNAL SIGINT
```
ansasaki, line 105: With #613, this should not be necessary anymore

docker/release/Dockerfile.distroless
```diff
@@ -0,0 +1,102 @@
...
+
+# we default the log level to info if not overwritten
+ENV RUST_LOG=keylime_agent=info
+
+# the agent currently listens on this port by default
+# it's good practice to declare this in the Dockerfile
+EXPOSE 9002/tcp
+
+# define the stopsignal to be SIGINT like the systemd service file does
+STOPSIGNAL SIGINT
```
ansasaki, line 94: With #613, this should not be necessary anymore

mheese:
@ansasaki rebased and force-pushed :+1: 

with regards to your comment about the `STOPSIGNAL` not being necessary anymore because of #613 , I'll make another PR where I'm going to remove this as well as the systemd service line, and I'll cross-reference it

ansasaki (approved):

Issue #489, opened 2022-12-09
config: revocation_actions option should be a list


ansasaki:
Following the python agent configuration, the `revocation_actions` option should be a list and not a single string.

Currently, the rust agent expects a comma separated list of binaries in a single string. It should expect a list.

Related: #455

Issue #455, opened 2022-09-21
Rust agent `trusted_client_ca` configuration option allows setting a single file


ansasaki:
The rust agent does not support setting more than one file path for the trusted CA certificate.

The `trusted_client_ca` should be modified to accept a list of paths.

The downside would be to drop the support of the `default` keyword, as the TOML format does not allow a string to be set to an option expecting a list.

PR #604, opened 2023-06-13
Parse lists from configuration file

ansasaki:
Support lists for `trusted_client_ca` and `revocation_actions` in more flexible formats, including the  format described in the configuration specification. 

The lists in `trusted_client_ca` and `revocation_actions` can be:
* comma separated
* space separated
* new line separated
* within square brackets or not
* with quoted strings or not

Fixes: #440, #455, #489

NOTE: this adds [pest](https://crates.io/crates/pest) as a new dependency


aplanas (approved):

keylime-agent/src/config.rs
```diff
@@ -597,6 +597,12 @@ fn config_get_setting() -> Result<ConfigBuilder<DefaultState>, Error> {
     config_get_file_setting()
 }
 
+pub(crate) fn parse_list(list: &str) -> Vec<&str> {
+    list.split(&['"', ' ', ',', '[', ']', '\n', '\r'][..])
+        .filter(|x| !x.is_empty())
+        .collect()
+}
```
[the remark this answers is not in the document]
ansasaki: TBH, that is a restriction I'm adding here, but I'm not sure is acceptable.

Maybe we need a more sophisticated parser that can deal with quoted strings properly so that we do not require the files names to be restricted.

I'll try to write such parser, and see if it gets too complicated.

kkaarreell:
@ansasaki Would it be reasonable to update our test so it doesn't use bundled CA certificates?
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/functional/basic-attestation-with-custom-certificates/test.sh#L94

ansasaki:
> @ansasaki Would it be reasonable to update our test so it doesn't use bundled CA certificates? https://github.com/RedHat-SP-Security/keylime-tests/blob/main/functional/basic-attestation-with-custom-certificates/test.sh#L94

Yes, it makes sense, I'll update the test and use a temporary commit to use a separate branch for tests

keylime-agent/src/config.rs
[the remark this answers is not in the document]
ansasaki: @ueno I implemented a simple parser for the lists, could you please check?

keylime-agent/src/config.rs
```diff
@@ -652,12 +658,19 @@ fn config_translate_keywords(
...
+            .iter()
+            .map(|t| {
+                config_get_file_path(
+                    "trusted_client_ca",
+                    t,
+                    keylime_dir,
+                    DEFAULT_TRUSTED_CLIENT_CA,
+                )
+            })
+            .collect::<Vec<String>>()
```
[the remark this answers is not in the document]
ansasaki: Done, it worked without specifying the type :)

keylime/src/list_parser.rs
```diff
@@ -0,0 +1,297 @@
...
+        }
+        if s.is_empty() {
+            Err(Error::EmptyString(self.consumed))
+        } else {
+            Ok(s)
+        }
+    }
+}
+
+pub fn parse_list(list: &str) -> Result<Vec<String>, Error> {
```
ueno, line 160: Is my understanding correct that the type of vector elements can be either a quoted string (`"foo"`) or a symbolic name (`default`)? If so I would define an enum like:
```rust
pub enum ListItem {
  QuotedString(String),
  Name(String),
}
```

keylime/src/list_parser.rs
```diff
@@ -0,0 +1,297 @@
...
+    consumed: usize,
+}
+
+impl<'a> ListParser<'a> {
+    fn parse(&mut self, bracketed: bool) -> Result<Vec<String>, Error> {
+        let mut r = Vec::new();
+
+        while let Some(c) = self.c.peek() {
+            match *c {
+                ',' | ' ' | '\n' | '\t' | '\r' => {
```
ueno, line 31: Just wondering: how bad it would be to use a parsing library like [pest|https://pest.rs/], as some of them are already appear in the dependencies?

keylime/src/list_parser.rs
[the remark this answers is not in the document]
ansasaki: Will do.

keylime/src/list_parser.rs
ansasaki: This is a possibility I considered, but it felt like an overkill. If preferred, I can do it (the addition of the dependency as you pointed out is not very impactful as we have a transitive dependency).

keylime/src/list_parser.rs
ueno: Sorry, I thought that a typical  input is something like `["/path/to/somewhere", default, ...]`, where the first element is a path while the second element is a symbol with special meaning. But if it treats everything as a string, distinct types wouldn't be needed.

keylime/src/list_parser.rs
ansasaki: In the end, I like the result obtained by generating the parser with `pest`. It is easier to understand the intended grammar. Maybe there is a simpler grammar than what I used, but it covers the test cases I created for the manually written parser. The part in which the parsed elements are obtained from the resulting structure did not require sophisticated  processing as we simply want to extract the strings, not understand the structure of the provided list.

keylime/src/list_parser.rs
```diff
@@ -0,0 +1,191 @@
...
+/// # Examples
+///
+/// Valid input lists:
+///
+/// * `a b c`
+/// * `'a' "b" c`
+/// * `'a', 'b', 'c'`
+/// * `[a b c]`
+/// * `['a', "b", c]`
+///
```
ueno, line 77: This interface looks a bit too generic to be exposed from the `keylime` lib crate. Not necessarily in this PR, but we might eventually want to move the entire configuration handling in the lib crate with access methods to retrieve list of certs, actions, etc.

ueno (approved):

keylime/src/list_parser.rs
ansasaki: I noticed a bug in the way I parsed the strings: I cannot remove the quotes from quoted strings. I made the changes to keep the quotes.

ansasaki:
I removed the temporary commit that changed the source of tests. There is an ongoing outage on packit/testing farm which made the tests to not be executed after that, but I'll merge based on the previous test results (which included the test with the workaround removed).

PR #608, opened 2023-06-29
Initial PR to add support for IDevID and IAK

Isaac-Matthews:
This is the first of a series of PRs that will enable the use of IDevIDs and IAKs as proposed in enhancement 81.
The modifications to the Keylime repositories are:

**Keylime**
Columns for the IDevID and IAK are added to the database
The IAK and IDevID sent by the agent are processed
The AK is verified by the IAK

**Rust-Keylime**
Config options to enable IDevID and IAK generation and use are added, with choices on how to generate the IDevID and IAK
The IAK and IDevID are generated according to config options
The IAK is used to certify the AK in the registration process

The purpose of this PR is to begin enabling the use of IAK and IDevID in keylime, starting with their generation. Options for the server config are not yet added so the inclusion of IAK and IDevID are left as optional for the agent.
The changes are all backwards compatible. The use of the IAK to certify and verify the AK is to act as a simple early use case in the process. Down the road this will be changed to encompass all the options listed in the enhancement, and those options will be included in the Keylime config.

N.B.
The certify in Rust-keylime currently just uses the UUID as qualifying data. This should perhaps be changed to include some temporal data at some point.

### DIFF
--- a/keylime-agent.conf
+++ b/keylime-agent.conf
@@ -227,6 +227,21 @@ tpm_signing_alg = "rsassa"
 # To override ek_handle, set KEYLIME_AGENT_EK_HANDLE environment variable.
 ek_handle = "generate"
 
+# Enable IDevID and IAK usage and set their algorithms.
+# Choosing a template will override the name and asymmetric algorithm choices.
+# Templates are specified in the TCG document found here, section 7.3.4: 
+# https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf
+#
+# Accepted values:
+# iak_idevid_asymmetric_alg:   rsa, ecc
+# iak_idevid_name_alg:        sha256, sm3_256, sha384, sha512
+# iak_idevid_template:        H-1, H-2, H-3, H-4, H-5
+# Leave template as "" in order to use asymmetric and name algorithm options
+enable_iak_idevid = false
+iak_idevid_asymmetric_alg = "rsa"
+iak_idevid_name_alg = "sha256"
+iak_idevid_template = ""
+
 # Use this option to state the existing TPM ownerpassword.
 # This option should be set only when a password is set for the Endorsement
 # Hierarchy (e.g. via "tpm2_changeauth -c e").
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -78,9 +78,9 @@ use tokio::{
 };
 use tss_esapi::{
     handles::KeyHandle,
-    interface_types::algorithm::AsymmetricAlgorithm,
+    interface_types::algorithm::{AsymmetricAlgorithm, HashingAlgorithm},
     interface_types::resource_handles::Hierarchy,
-    structures::{Auth, PublicBuffer},
+    structures::{Auth, Data, Digest, MaxBuffer, PublicBuffer},
     traits::Marshall,
     Context,
 };
@@ -278,6 +278,24 @@ async fn main() -> Result<()> {
         config.agent.tpm_signing_alg.as_ref(),
     )?;
 
+    let (asym_alg, name_alg) = tpm::get_idevid_template(
+        config.agent.iak_idevid_template.as_str(),
+        config.agent.iak_idevid_asymmetric_alg.as_str(),
+        config.agent.iak_idevid_name_alg.as_str(),
+    )?;
+
+    let (iak, idevid) = if config.agent.enable_iak_idevid {
+        let idevid = ctx.create_idevid(asym_alg, name_alg)?;
+        info!("IDevID created.");
+        // Flush after creating to make room for AK and EK and IAK
+        ctx.as_mut().flush_context(idevid.handle.into())?;
+        let iak = ctx.create_iak(asym_alg, name_alg)?;
+        info!("IAK created.");
+        (Some(iak), Some(idevid))
+    } else {
+        (None, None)
+    };
+
     // Gather EK values and certs
     let ek_result = match config.agent.ek_handle.as_ref() {
         "" => ctx.create_ek(tpm_encryption_alg, None)?,
@@ -384,6 +402,26 @@ async fn main() -> Result<()> {
     }
 
     info!("Agent UUID: {}", agent_uuid);
+
+    let (attest, signature) = if config.agent.enable_iak_idevid {
+        let qualifying_data = config.agent.uuid.as_bytes();
+        let (attest, signature) = ctx.certify_credential_with_iak(
+            Data::try_from(qualifying_data).unwrap(), //#[allow_ci]
+            ak_handle,
+            iak.as_ref().unwrap().handle, //#[allow_ci]
+        )?;
+        info!("AK certified with IAK.");
+
+        // // For debugging certify(), the following checks the generated signature
+        // let max_b = MaxBuffer::try_from(attest.clone().marshall()?)?;
+        // let (hashed_attest, _) = ctx.inner.hash(max_b, HashingAlgorithm::Sha256, Hierarchy::Endorsement,)?;
+        // println!("{:?}", hashed_attest);
+        // println!("{:?}", signature);
+        // println!("{:?}", ctx.inner.verify_signature(iak.as_ref().unwrap().handle, hashed_attest, signature.clone())?); //#[allow_ci]
+        (Some(attest), Some(signature))
+    } else {
+        (None, None)
+    };
 
     // Generate key pair for secure transmission of u, v keys. The u, v
     // keys are two halves of the key used to decrypt the workload after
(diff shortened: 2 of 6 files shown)


keylime/src/tpm.rs
```diff
@@ -225,6 +305,345 @@ impl Context {
...
+                .with_auth_policy(Digest::try_from(auth_policy)?)
+                .with_rsa_parameters(
+                    PublicRsaParametersBuilder::new()
+                        .with_symmetric(SymmetricDefinitionObject::Null)
+                        .with_scheme(RsaScheme::Null)
+                        .with_key_bits(key_bits)
+                        .with_exponent(RsaExponent::default())
+                        .with_is_signing_key(obj_attrs.sign_encrypt())
+                        .with_is_decryption_key(obj_attrs.decrypt())
+                        .with_restricted(obj_attrs.restricted())
```
stefanberger, line 419: I don't know what effect these last 3 have but I don't see them in the spec 7.3.4.2

keylime/src/tpm.rs
```diff
@@ -225,6 +305,345 @@ impl Context {
...
+                HashingAlgorithm::Sha256,
+                &[
+                    PcrSlot::Slot0,
+                    PcrSlot::Slot1,
+                    PcrSlot::Slot2,
+                    PcrSlot::Slot3,
+                    PcrSlot::Slot4,
+                    PcrSlot::Slot5,
+                    PcrSlot::Slot6,
+                    PcrSlot::Slot7,
```
stefanberger, line 498: Why do you pass this specific set of PCRs here?

keylime-agent.conf
```diff
@@ -227,6 +227,21 @@ tpm_signing_alg = "rsassa"
...
+# https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf
+#
+# Accepted values:
+# iak_idevid_asymmetric_alg:   rsa, ecc
+# iak_idevid_name_alg:        sha256, sm3_256, sha384, sha512
+# iak_idevid_template:        H-1, H-2, H-3, H-4, H-5
+# Leave template as "" in order to use asymmetric and name algorithm options
+enable_iak_idevid = false
+iak_idevid_asymmetric_alg = "rsa"
+iak_idevid_name_alg = "sha256"
```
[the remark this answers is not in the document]
Isaac-Matthews: If the user uses a template these options are not used. So selecting H-1 will always use RSA and sha256. The user can choose any combination if they do not use a template.

keylime/src/tpm.rs
```diff
@@ -225,6 +305,345 @@ impl Context {
...
+        let obj_attrs_builder = ObjectAttributesBuilder::new()
+            .with_fixed_tpm(true)
+            .with_st_clear(false)
+            .with_fixed_parent(true)
+            .with_sensitive_data_origin(true)
+            .with_user_with_auth(true)
+            .with_admin_with_policy(true)
+            .with_no_da(false)
+            .with_encrypted_duplication(false)
+            .with_sign_encrypt(true)
```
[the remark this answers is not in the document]
Isaac-Matthews: Yes I believe this is just a naming difference to differentiate from the new x509_sign attribute

keylime/src/tpm.rs
```diff
@@ -225,6 +305,345 @@ impl Context {
...
+                .with_public_algorithm(PublicAlgorithm::Rsa)
+                .with_name_hashing_algorithm(name_alg)
+                .with_object_attributes(obj_attrs)
+                .with_auth_policy(Digest::try_from(auth_policy)?)
+                .with_rsa_parameters(
+                    PublicRsaParametersBuilder::new()
+                        .with_symmetric(SymmetricDefinitionObject::Null)
+                        .with_scheme(RsaScheme::Null)
+                        .with_key_bits(key_bits)
+                        .with_exponent(RsaExponent::default())
```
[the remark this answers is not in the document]
Isaac-Matthews: Yes this is correct. TPMs are only required to accept a 0 exponent that they interpret as 0x10001

keylime/src/tpm.rs
Isaac-Matthews: These come from 7.3.4.1, they are flags for the builder to specify if the key will be used to sign, decrypt etc

keylime/src/tpm.rs
```diff
@@ -225,6 +305,345 @@ impl Context {
...
+                EccCurve::Sm2P256,
+            ),
+            HashingAlgorithm::Sha384 => (
+                IDEVID_AUTH_POLICY_SHA384[0..48].to_vec(),
+                RsaKeyBits::Rsa3072,
+                EccCurve::NistP384,
+            ),
+            HashingAlgorithm::Sha512 => (
+                IDEVID_AUTH_POLICY_SHA512[0..64].to_vec(),
+                RsaKeyBits::Rsa4096,
```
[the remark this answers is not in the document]
Isaac-Matthews: There is no template, but this can only be used if specifically selected by the user instead of a template

keylime/src/tpm.rs
stefanberger: I suppose the state of these PCRs indicates the state of the machine when the key was created? Are you going to check anything with these PCRs? I am wondering how their state is going to be used.

keylime-agent.conf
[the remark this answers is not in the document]
stefanberger: Doesn't the user have to be in sync with one of the templates that the manufacturer chose?

stefanberger:
Another question: Does the agent need to know (at this point in the series of changes or later?) that the new keys are supported on the server side and so they do not go into the void?

ansasaki:
I think this feature is useful, I'm looking forward for it to be finished.

I have some concerns around the addition of new configuration options. We have treated this lightly, but for me the correct would be to bump the minor version of the configuration and add the required mappings and templates for the configuration upgrades to make sure it will work in an upgrade scenario.

The new options defaults should be reflected on the mappings, similar to what was created for version 2.0 in https://github.com/keylime/keylime/tree/master/templates/2.0

 I can help on this if the required changes are not clear. To start I'll write the documentation that is currently missing to explain how the configuration upgrades are supposed to be extended and maintained.

Also, it would be great if tests were added for the feature. I think the [tests repository](https://github.com/RedHat-SP-Security/keylime-tests) evolved quite well, and new tests can be quite easily added based on the available examples. Also the used tools are handy (although the usage for keylime could be better explained in the repo README).

keylime-agent.conf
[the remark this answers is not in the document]
Isaac-Matthews: Yes the user will need to select the template that the manufacturer used to create the DevID originally. Perhaps adding a template (presumably H-1) to be used as default makes sense here.

Isaac-Matthews:
> Another question: Does the agent need to know (at this point in the series of changes or later?) that the new keys are supported on the server side and so they do not go into the void?

My thought was that the requirements for registration and enforcement would all be set on the server side and so this doesn't need to be communicated to the agent.

Isaac-Matthews:
> I have some concerns around the addition of new configuration options. We have treated this lightly, but for me the correct would be to bump the minor version of the configuration and add the required mappings and templates for the configuration upgrades to make sure it will work in an upgrade scenario.
> 
> The new options defaults should be reflected on the mappings, similar to what was created for version 2.0 in https://github.com/keylime/keylime/tree/master/templates/2.0

Makes sense. Could you please send me the docs on that and I will make the changes (here or on slack). Thanks.

> Also, it would be great if tests were added for the feature. I think the [tests repository](https://github.com/RedHat-SP-Security/keylime-tests) evolved quite well, and new tests can be quite easily added based on the available examples. Also the used tools are handy (although the usage for keylime could be better explained in the repo README).

Definitely intend to include tests; I planned on adding them alongside the next patch that will enable proper use of this feature as they should test the full process. Let me know if that isn't OK.

keylime-agent/src/main.rs
```diff
@@ -278,6 +278,56 @@ async fn main() -> Result<()> {
...
+                "ecc" => {tss_esapi::interface_types::algorithm::AsymmetricAlgorithm::Ecc},
+                _ => {tss_esapi::interface_types::algorithm::AsymmetricAlgorithm::Null},
+            },
+            match config.agent.iak_idevid_name_alg.as_str() {
+                "sha256" => {tss_esapi::interface_types::algorithm::HashingAlgorithm::Sha256},
+                "sm3_256" => {tss_esapi::interface_types::algorithm::HashingAlgorithm::Sm3_256},
+                "sha384" => {tss_esapi::interface_types::algorithm::HashingAlgorithm::Sha384},
+                "sha512" => {tss_esapi::interface_types::algorithm::HashingAlgorithm::Sha512},
+                _ => {tss_esapi::interface_types::algorithm::HashingAlgorithm::Null},
+            })
```
ansasaki, line 314: The same comment above applies here. Maybe also reuse the conversion already available in `algorithms.rs`.

keylime-agent/src/main.rs
```diff
@@ -278,6 +278,56 @@ async fn main() -> Result<()> {
...
+        let idevid = ctx.create_idevid(asym_alg, name_alg)?;
+        info!("IDevID created.");
+        // Flush after creating to make room for AK and EK and IAK
+        ctx.as_mut().flush_context(idevid.handle.into())?;
+
+        let iak = ctx.create_iak(asym_alg, name_alg)?;
+        info!("IAK created.");
+        (Some(iak), Some(idevid))
+    } else {
+        (None, None)
```
ansasaki, line 328: Again, wouldn't be better to move this to a dedicated lib?

The API could return the Tuple `(iak, idevid)` or `(None, None)`

keylime-agent/src/main.rs
```diff
@@ -497,18 +567,49 @@ async fn main() -> Result<()> {
...
+                Some(
+                    &PublicBuffer::try_from(iak.unwrap().public.clone())? //#[allow_ci]
+                        .marshall()?,
+                ),
+                Some(
+                    &PublicBuffer::try_from(idevid.unwrap().public.clone())? //#[allow_ci]
+                        .marshall()?,
+                ),
+                Some(attest.unwrap().marshall()?), //#[allow_ci]
+                Some(signature.unwrap().marshall()?), //#[allow_ci]
```
ansasaki, line 588: Please, avoid using `unwrap()` here.  You could use matching to "extract" the contents of the Options before calling `do_register_agent`.

A suggestion would be to use something like:

```
if let (Some(iak), Some(idevid), Some(attest), Some(signature)) = (iak, idevid, attest, signature)  {
    registrar_agent::do_register_agent(...)
} else {
    error!(...)
}
```

I did not test this, but I think it would avoid using unwrap here

ansasaki (changes requested):

keylime-agent/src/main.rs
Isaac-Matthews: I have removed the unwraps as suggested

keylime-agent/src/main.rs
```diff
@@ -278,6 +278,56 @@ async fn main() -> Result<()> {
...
+            tss_esapi::interface_types::algorithm::HashingAlgorithm::Sha384)
+        },
+        "H-4" => {
+            (tss_esapi::interface_types::algorithm::AsymmetricAlgorithm::Ecc,
+            tss_esapi::interface_types::algorithm::HashingAlgorithm::Sha512)
+        },
+        "H-5" => {
+            (tss_esapi::interface_types::algorithm::AsymmetricAlgorithm::Ecc,
+            tss_esapi::interface_types::algorithm::HashingAlgorithm::Sm3_256)
+        },
```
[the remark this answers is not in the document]
Isaac-Matthews: I have moved this to tpm.rs as suggested

keylime-agent/src/main.rs
Isaac-Matthews: likewise for this one, and I have used the `algorithm.rs` conversions

keylime-agent/src/main.rs
Isaac-Matthews: This I think makes sense here along with the similar `create_ak` and `create_ek` functions

keylime-agent/src/main.rs
```diff
@@ -568,6 +536,11 @@ async fn main() -> Result<()> {
     {
         // Request keyblob material
         let keyblob = if config.agent.enable_iak_idevid {
+            let (Some(iak), Some(idevid), Some(attest), Some(signature)) = (iak, idevid, attest, signature)
```
ansasaki, line 539: Is an `if` missing here?

```
    if let (Some(iak), Some(idevid), Some(attest), Some(signature)) = (iak, idevid, attest, signature)
```

Isaac-Matthews: No I used a let-else

ansasaki: Oh, I see, didn't know about this syntax. Do you know since when is this available?

My concern is if this requires too new version of the rust compiler, which might be disruptive for people stuck with older compiler version.

Isaac-Matthews: Looks like it was added end of 2021. Is that too recent?

ansasaki: Ok, found the answer for my own question: this is available since rust 1.65: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html#let-else

@galmasi I remember you asked some time ago which was the oldest compiler version we supported. I think you asked if it could be version 1.67. Is this correct? If yes, I think using let-else is ok (and we should use everywhere as this looks much better then if-let-else statements)

keylime-agent/src/main.rs
ansasaki: Ok, we can leave this in `main.rs`

ansasaki (approved):

PR #611, opened 2023-07-06
hash: add more configurable hash algorithm for public key digest

dongx1x:
The current public key digest is using SHA1 and SHA256, and the digest will extend to TPM PCR 16. But from TPM2, SHA1 is not mandatory anymore, so this patch makes the hash algorithm configurable.

Reference:
TCG PC Client Platform TPM Profile Specification for TPM 2.0, Section 4.6 https://trustedcomputinggroup.org/wp-content/uploads/PC-Client-Specific-Platform-TPM-Profile-for-TPM-2p0-v1p05p_r14_pub.pdf

### DIFF
--- a/keylime/src/tpm.rs
+++ b/keylime/src/tpm.rs
@@ -284,8 +284,6 @@ impl Context {
 
     // This function extends Pcr16 with the digest, then creates a PcrList
     // from the given mask and pcr16.
-    // Note: Currently, this will build the list for both SHA256 and SHA1 as
-    // necessary for the Python components of Keylime.
     fn build_pcr_list(
         &mut self,
         digest: DigestValues,
@@ -329,7 +327,7 @@ impl Context {
         hash_alg: HashAlgorithm,
         sign_alg: SignAlgorithm,
     ) -> Result<String> {
-        let nk_digest = pubkey_to_tpm_digest(pubkey)?;
+        let nk_digest = pubkey_to_tpm_digest(pubkey, hash_alg)?;
 
         let pcrlist =
             self.build_pcr_list(nk_digest, mask, hash_alg.into())?;
@@ -452,10 +450,9 @@ fn parse_cred_and_secret(
 }
 
 // Takes a public PKey and returns a DigestValue of it.
-// Note: Currently, this creates a DigestValue including both SHA256 and
-// SHA1 because these banks are checked by Keylime on the Python side.
 fn pubkey_to_tpm_digest<T: HasPublic>(
     pubkey: &PKeyRef<T>,
+    hash_algo: HashAlgorithm,
 ) -> Result<DigestValues> {
     let mut keydigest = DigestValues::new();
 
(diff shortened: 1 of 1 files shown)


keylime/src/tpm.rs
```diff
@@ -468,16 +465,11 @@ fn pubkey_to_tpm_digest<T: HasPublic>(
...
-    // SHA1
-    let mut hasher = openssl::sha::Sha1::new();
-    hasher.update(&keybytes);
-    let hashvec: Vec<u8> = hasher.finish().into();
-    keydigest.set(HashingAlgorithm::Sha1, Digest::try_from(hashvec)?);
+    let hash_algo = HashingAlgorithm::from(algo);
+    let mut hasher = Hasher::new(hash_alg_to_message_digest(hash_algo)?)?;
+    hasher.update(&keybytes)?;
+    let hashvec = hasher.finish()?;
+    keydigest.set(hash_algo, Digest::try_from(hashvec.as_ref())?);
```
ueno, line 472: Previously, it extended PCR16 with both SHA-1 and SHA-256. With this change, the slot seems to be only extended with the given single algorithm. Wouldn't that be a problem?

dongx1x: Hi @ueno, thanks for your comment! 
I think it's safe here, let me explain more, please correct me if I'm wrong.
The working flow of Keylime attestation should be: `extend PCR16` --> `build PCR list` --> `get a quote with PCR list` --> `check the quote` --> `check PCR 16`.

1. The PCR list is built with [the specific hash algorithm](https://github.com/keylime/rust-keylime/blob/master/keylime/src/tpm.rs#L311).

2.  When the verifier checks the PCR 16, the expected value is generated with [the specific hash algorithm](https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_main.py#L203).

So, although it extended PCR16 with both SHA-1 and SHA-256 but looks like only the specific one will be measured.

THS-on: Yes, we check only the PCR 16 in the configured algorithm, so this change should not break anything.

THS-on (commented):
@dongx1x can you fix the formatting (`cargo fmt`) and also add SM3_256 as an hash algorithm (that one is also supported by the server side: https://github.com/keylime/keylime/blob/master/keylime/common/algorithms.py)?

Otherwise the change LGTM.

keylime/src/tpm.rs
```diff
@@ -452,10 +450,9 @@ fn parse_cred_and_secret(
 }
 
 // Takes a public PKey and returns a DigestValue of it.
-// Note: Currently, this creates a DigestValue including both SHA256 and
-// SHA1 because these banks are checked by Keylime on the Python side.
 fn pubkey_to_tpm_digest<T: HasPublic>(
     pubkey: &PKeyRef<T>,
+    algo: HashAlgorithm
```
aplanas, line 455: nit: this variable is named `hash_alg` in other places

dongx1x: OK, updated.

dongx1x:
> @dongx1x can you fix the formatting (`cargo fmt`) and also add SM3_256 as an hash algorithm (that one is also supported by the server side: https://github.com/keylime/keylime/blob/master/keylime/common/algorithms.py)?
> 
> Otherwise the change LGTM.

Sure, updated.

aplanas (approved):
LGTM ... but I would wait the imput of @ansasaki or @ueno

ueno (approved):

ansasaki (approved):

Issue #323, opened 2022-02-07
Add "currently" to swtpm warning


lkatalin:
https://github.com/keylime/keylime/pull/881

sarroutbi:
@lkatalin: Should this issue be closed? Does [keylime/keylime#881](https://github.com/keylime/keylime/pull/881) handle this?

sarroutbi:
OK. Thanks for clarifying. I will try and open a PR for that. Thanks !!!

PR #632, opened 2023-08-03
Use "currently" for swtpm emulator warning

sarroutbi:
Resolves: #323

### DIFF
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -231,7 +231,7 @@ async fn main() -> Result<()> {
     //  Retrieve the TPM Vendor, this allows us to warn if someone is using a
     // Software TPM ("SW")
     if tss_esapi::utils::get_tpm_vendor(ctx.as_mut())?.contains("SW") {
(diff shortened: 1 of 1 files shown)


keylime-agent/src/main.rs
```diff
@@ -231,7 +231,7 @@ async fn main() -> Result<()> {
     //  Retrieve the TPM Vendor, this allows us to warn if someone is using a
     // Software TPM ("SW")
     if tss_esapi::utils::get_tpm_vendor(ctx.as_mut())?.contains("SW") {
-        warn!("INSECURE: Keylime is using a software TPM emulator rather than a real hardware TPM.");
+        warn!("INSECURE: Keylime is using currently a software TPM emulator rather than a real hardware TPM.");
```
ansasaki, line 234: For me, it sounds better if "currently" comes before "using":
```suggestion
        warn!("INSECURE: Keylime is currently using a software TPM emulator rather than a real hardware TPM.");
```
What do you think?

aplanas (approved):

ansasaki:
@sarroutbi Could you please add a sign-off to the commit to fix the DCO check?

PR #634, opened 2023-08-07
Update container build workflow actions

ansasaki:
Update actions versions to avoid warnings.


THS-on (approved):

THS-on:
Good catch, should we also update them on the keylime python repo?

ansasaki:
> Good catch, should we also update them on the keylime python repo?

I think we should

sergio-correia (approved):

Issue #283, opened 2021-12-10
Don't hardcode endianness


lkatalin:
Use the host endianness instead of `to_le_bytes()` or `to_be_bytes()`. I think there is an easy way to do this from `byteorder` but I'd have to look again.

ansasaki:
Check if this is actually valid: the encoding is kept the same independently of the arch to avoid encoding errors due to endianess. A good approach to check is to test the changes proposed in #637 mixing big endian and little endian arches

PR #637, opened 2023-08-11
Use cpu_endian to check local endianness

sarroutbi:
Resolves: #283

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -539,6 +539,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6245d59a3e82a7fc217c5828a6692dbc6dfb63a0c8c90495621f7b9d79704a0e"
 
 [[package]]
+name = "cpu-endian"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512746971686a86c7b8aee38244eaa7e04696e612790d593f1d73059536beee8"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "cpufeatures"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1097,6 +1106,7 @@ name = "keylime"
 version = "0.2.2"
 dependencies = [
  "base64 0.21.0",
+ "cpu-endian",
  "hex",
  "log",
  "openssl",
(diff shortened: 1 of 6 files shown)


aplanas:
> If anybody can help with this, I am grateful. Otherwise, I will keep the PR as it is right now

You need to add a bound to "T". IIUC "T" will always be "u32", "u16" or "u8". I that case maybe you can use something like this: https://users.rust-lang.org/t/generic-function-for-from-be-bytes/59629/5

aplanas (approved):

ansasaki:
@sarroutbi Have you tested the interoperability of this between a big endian and a little endian machine?

I'm afraid this can break interoperability. The reason of making the endianess fixed is for all machines to encode data in the same way.

Another downside is the addition of a new dependency. It is not available in Fedora for example, which will cause problems on packaging.

sarroutbi:
> @sarroutbi Have you tested the interoperability of this between a big endian and a little endian machine?
> 
> I'm afraid this can break interoperability. The reason of making the endianess fixed is for all machines to encode data in the same way.
> 
> Another downside is the addition of a new dependency. It is not available in Fedora for example, which will cause problems on packaging.

You are right. I did not figure out that scenario. If codification is hard coded intentionally, I guess #283 should be closed with no resolution. I will close this PR. Sorry for confusion

ansasaki:
@sarroutbi Sorry, I didn't mean you should close this, but to test if this would actually work in an interoperability scenario.

We can test it and reopen this if it works. But then we would have to work to introduce the new dependency in Fedora.

If it doesn't work, then we should do what you proposed and close #283 with the reasoning.

sarroutbi:
> @sarroutbi Sorry, I didn't mean you should close this, but to test if this would actually work in an interoperability scenario.
> 
> We can test it and reopen this if it works. But then we would have to work to introduce the new dependency in Fedora.
> 
> If it doesn't work, then we should do what you proposed and close #283 with the reasoning.

No problem. I will reopen it and once tested, will let you know.

aplanas (commented):
I think @ansasaki is right and my review was wrong. Now the data sent from a x86-64 and aarch64 will have different de-serialization order

sarroutbi:
I will close this PR, as it has been confirmed that current development status is intentionally done. My suggestion is closing #283, as it can be confusing

PR #651, opened 2023-09-18
Do not use too specific version on cargo audit workflow

ansasaki:
use actions-rs/audit-check@v1 instead of actions-rs/audit-check@v1.4.1


aplanas (approved):

ueno (approved):

ansasaki:
I create an issue to track the warning from the Fedora 37 run: #652

Issue #657, opened 2023-09-20
Proposal: remove "testing only" limitation from agent: allow configurable locations for boot and IMA logs.


galmasi:
Description of problem
===

The ongoing effort in [attestation operator](https://github.com/keylime/attestation-operator) is to make the keylime agent deploy on kubernetes clusters in unprivileged pods. This requires two problems to be solved; (a) mapping the TPM device and (b) mapping the measured boot and IMA logs into said unprivileged containers.

We propose to use @mheese 's TPM [device plugin](https://github.com/githedgehog/k8s-tpm-device-plugin) container for both these roles. With minor changes, it can be expanded to map the logs in addition to the TPM device.

However, at this moment we cannot find any way to map the boot/IMA logs into their respective places in the security FS (e.g. `/sys/kernel/security/tpm0/binary_boot_measurements`). We can map them to other arbitrary places in the keylime agent container, however -- even in `/` -- and have them be accessible (readonly) to the target pod.

The gist of the proposed solution
===

I would like to propose to add the following functionality to the Rust Agent: let us configure the location of the boot/IMA logs in the configuration file, with the obvious reasonable defaults in place.

Security considerations
===

I cannot think of any serious security implications. The IMA and boot logs are never, ever trusted in any shape or fashion by the agent, by the verifier or any other component of Keylime. Substituting fakes can only lead to keylime verification failing. Therefore I do not see a viable path to subvert keylime functionality with the proposed change.

Code circumstances
===

The current implementation of the rust agent already has the "hidden" ability to change the [location](https://github.com/keylime/rust-keylime/blob/626a1ccdaa6711fcbeb8b6992a63d6899d48f4e6/keylime-agent/src/main.rs#L157) of the binary boot log, but this is a "testing only" feature. I propose to replace "testing only" with a "read configuration" option. This implies that we add at least two new configuration options. The specific naming of these options should be left to whoever proposes the PR to implement this feature.

Affected personnel
===

[...]

PR #665, opened 2023-10-02
Adding "mutable log paths" feature 

galmasi:
As discussed in Issue #657  -- initial commit for implementing configurable logging paths

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -25,6 +25,10 @@ all: $(programs)
 $(programs):
 	cargo build --target-dir="${TARGETDIR}" ${CARGO_ARGS}
 
+.PHONY: clean
+clean::
+	${RM} -rf target
+
 .PHONY: install
 install: all
 	mkdir -p /etc/keylime/
--- a/keylime-agent/src/config.rs
+++ b/keylime-agent/src/config.rs
@@ -63,6 +63,10 @@ pub static DEFAULT_IAK_IDEVID_NAME_ALG: &str = "sha256";
 pub static DEFAULT_IAK_IDEVID_TEMPLATE: &str = "H-1";
 pub static DEFAULT_RUN_AS: &str = "keylime:tss";
 pub static DEFAULT_AGENT_DATA_PATH: &str = "agent_data.json";
+pub static DEFAULT_IMA_ML_PATH: &str =
+    "/sys/kernel/security/ima/ascii_runtime_measurements";
+pub static DEFAULT_MEASUREDBOOT_ML_PATH: &str =
+    "/sys/kernel/security/tpm0/binary_boot_measurements";
 pub static DEFAULT_CONFIG: &str = "/etc/keylime/agent.conf";
 pub static DEFAULT_CONFIG_SYS: &str = "/usr/etc/keylime/agent.conf";
 
@@ -108,6 +112,8 @@ pub(crate) struct EnvConfig {
     pub iak_idevid_template: Option<String>,
     pub run_as: Option<String>,
     pub agent_data_path: Option<String>,
+    pub ima_ml_path: Option<String>,
+    pub measuredboot_ml_path: Option<String>,
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
@@ -152,6 +158,8 @@ pub(crate) struct AgentConfig {
     pub iak_idevid_template: String,
     pub run_as: String,
     pub agent_data_path: String,
+    pub ima_ml_path: String,
+    pub measuredboot_ml_path: String,
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
@@ -325,6 +333,15 @@ impl EnvConfig {
         if let Some(ref v) = self.agent_data_path {
             _ = agent
                 .insert("agent_data_path".to_string(), v.to_string().into());
+        }
+        if let Some(ref v) = self.ima_ml_path {
+            _ = agent.insert("ima_ml_path".to_string(), v.to_string().into());
+        }
+        if let Some(ref v) = self.measuredboot_ml_path {
+            _ = agent.insert(
+                "measuredboot_ml_path".to_string(),
+                v.to_string().into(),
+            );
         }
         agent
     }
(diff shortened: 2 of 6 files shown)


galmasi:
As mentioned above, please do not merge yet -- this is a trial balloon to see how many testing features I run afoul in one try. Also, the implementation is incomplete: the feature discussed in Issue #657 , info logging nonstandard log paths, is not yet implemented.

galmasi:
@ansasaki @kkaarreell I do not believe I can improve the rust agent code in this PR anymore. I have been laboring under a misunderstanding of how rust agent testing is performed. I have created this one-line [PR] (https://github.com/RedHat-SP-Security/keylime-tests/pull/498) in `keylime-tests` to help pass this PR.

Note that the `testing` feature in the keylime agent is not actually used for measured boot in the test framework; instead, the test framework directly `sed`s the source code. This raises the issue that either 

(a) the testing harness should be updated to compile the rust agent correctly by using cargo with the `--feature testing` parameters; -- or -- 
(b) let the SED command stay in the test code and remove the unused `testing` feature of the keylime agent.

ansasaki:
> @ansasaki @kkaarreell I do not believe I can improve the rust agent code in this PR anymore. I have been laboring under a misunderstanding of how rust agent testing is performed. I have created this one-line [PR] ([RedHat-SP-Security/keylime-tests#498](https://github.com/RedHat-SP-Security/keylime-tests/pull/498)) in `keylime-tests` to help pass this PR.
> 
> Note that the `testing` feature in the keylime agent is not actually used for measured boot in the test framework; instead, the test framework directly `sed`s the source code. This raises the issue that either
> 
> (a) the testing harness should be updated to compile the rust agent correctly by using cargo with the `--feature testing` parameters; -- or -- (b) let the SED command stay in the test code and remove the unused `testing` feature of the keylime agent.

This looks like a leftover. It should use the approach `a` and avoid changing the code. I'll try to get rid of that `sed`.

galmasi:
I closed https://github.com/RedHat-SP-Security/keylime-tests/pull/498 in the expectation that https://github.com/RedHat-SP-Security/keylime-tests/pull/499 will solve this problem. Will force a re-test as soon as 499 is merged.

keylime-agent/src/config.rs
```diff
@@ -696,6 +723,20 @@ fn config_translate_keywords(
...
+        DEFAULT_IMA_ML_PATH,
+    );
+
+    let mut measuredboot_ml_path = config_get_file_path(
+        "measuredboot_ml_path",
+        &config.agent.measuredboot_ml_path,
+        keylime_dir,
+        DEFAULT_MEASUREDBOOT_ML_PATH,
+    );
+
```
ansasaki, line 739: I think using `config_get_file_path()` is not correct here. Note that it will use the provided default path relative from the `keylime_dir` (by default `/var/lib/keylime`).

Maybe a way to fix this would be to change the `DEFAULT_AGENT_DATA_PATH`, `DEFAULT_SERVER_KEY`, `DEFAULT_SERVER_CERT`, etc. to be relative paths (i.e. not starting with `/`) and make `config_get_file_path`  to prepend the `keylime_dir` only when the default value is a relative path.

keylime-agent/src/main.rs
```diff
@@ -149,16 +176,27 @@ async fn main() -> Result<()> {
...
-    // Allow setting the binary bios measurements log path when testing
+    // load path for MBA logfile
+    let mut measuredboot_ml_path =
+        Path::new(&config.agent.measuredboot_ml_path);
     let env_mb_path: String;
     #[cfg(feature = "testing")]
     if let Ok(v) = std::env::var("TPM_BINARY_MEASUREMENTS") {
         env_mb_path = v;
         measuredboot_ml_path = Path::new(&env_mb_path);
     }
```
ansasaki, line 187: This should be dropped, and the tests modified to use the new added way to set the measurements logs.

Note that the tests only passed because this is still in place, as the CI is setting this env var.  Otherwise it would had failed as the resulting path after `config_get_file_path()` would not be the expected one.

ansasaki (changes requested):

GNUmakefile
```diff
@@ -25,6 +25,9 @@ all: $(programs)
 $(programs):
 	cargo build --target-dir="${TARGETDIR}" ${CARGO_ARGS}
 
+clean::
```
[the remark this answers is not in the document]
galmasi: Double quote means that multiple clean targets can be attached. It's good `Makefile` practice, esp. for `clean` because there might be multiple aspects to clean up.

+1 I will add the `.PHONY` notation.

keylime-agent/src/main.rs
galmasi: I'd be very happy to drop this, but not while the CI is using it -- correct?
Unless you want to hold this PR until CI is fixed, we need to leave it in place.

ansasaki: This can be fixed in a separate PR.

keylime-agent/src/config.rs
ansasaki: Alternatively, I think it is acceptable that instead of `keylime_dir`, the `/` directory is used in this case.

Also, please add the new options to  `keylime-agent.conf` with a description and default value set. The configuration minor version number should be bumped.

galmasi:
I'll take another look at all the changes I have made and try to reconcile. I have rebased to the current master, but there are a number of thinking errors that I need to address: chief among them, `config_file_get_path()`. I need to build a test case that shows it used wrong, then fix it.

keylime-agent/src/config.rs
galmasi: "/" is fine. I was thinking of "/sys/kernel/security", just in case anyone would ever want to specify the IMA or MB log file as a relative path. But that's not worth parametrizing, so probably "/" is safer and less involved. OK, let's do that.

ansasaki (approved):

keylime-agent/src/config.rs
```diff
@@ -515,6 +532,14 @@ impl Source for KeylimeConfig {
             "agent_data_path".to_string(),
             self.agent.agent_data_path.to_string().into(),
         );
+        _ = m.insert(
```
aplanas, line 535: So many `_` seems a bit ugly. I wonder if this can help: https://github.com/rust-lang/rfcs/issues/3092

aplanas (approved):

ansasaki:
Hi @galmasi, I took the liberty to make the formatting changes and rebase. I'd like to ask if I can also add the configuration change required (in the case, include the new configuration options and bump the version)

keylime-agent/src/config.rs
ansasaki: Thanks for the pointer, but unfortunately we cannot use this (yet). It is still available only on nightly:
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.try_insert

ansasaki:
I'll squash some commits and rebase. Note that I added a commit to add the new configuration options.

PR #669, opened 2023-10-03
Add certificates and certificate checking for IDevID and IAK keys

Isaac-Matthews:
This is the next set of PRs to enable the use of IDevIDs and IAKs as proposed in [enhancement 81.](https://github.com/keylime/enhancements/blob/master/81-IDevID_and_IAK_support.md)
The previous PRs added IDevID and IAK creation, sending, storing and using to verify: [keylime PR](https://github.com/keylime/keylime/pull/1416), [rust-keylime PR](https://github.com/keylime/rust-keylime/pull/608)
The modifications to the Keylime repositories are:

**Keylime**
Columns for IDevID and IAK certificates
Config to require IDevID and IAK
IDevID and IAK certificate checks
Tests for cert checks

**Rust-Keylime**
IDevID and IAK certificate config options
Loading and sending IDevID and IAK certificates from paths

In this PR the certificates must be loaded from path. In future, the ability to load from TPM nvram will need to be added.


stefanberger:
I have the registrar in the mode that it requires iak + idevid. On the registrar I now see this here when starting the agent with the first line from my debugging added:

```
2023-11-02 17:52:14.047 - keylime.registrar - ERROR - idevid_cert: None iak_cert: None
2023-11-02 17:52:14.047 - keylime.registrar - WARNING - POST for d432fbb3-d2f1-4a97-9ef7-75bd81c00000 returning 400 response. Error: argument should be a bytes-like object or ASCII string, not 'NoneType'
2023-11-02 17:52:14.048 - keylime.registrar - ERROR - argument should be a bytes-like object or ASCII string, not 'NoneType'
  File "/usr/local/lib/python3.12/site-packages/keylime/registrar_common.py", line 339, in do_POST
    base64.b64decode(idevid_cert), base64.b64decode(iak_cert), config.get("tenant", "tpm_cert_store")
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/base64.py", line 83, in b64decode
    s = _bytes_from_decode_data(s)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/base64.py", line 45, in _bytes_from_decode_data
    raise TypeError("argument should be a bytes-like object or ASCII "
TypeError: argument should be a bytes-like object or ASCII string, not 'NoneType'

```

The agent then terminates:
```
# keylime_agent
Error: Registrar { addr: "http://127.0.0.1:8890/v2.0/agents/d432fbb3-d2f1-4a97-9ef7-75bd81c00000", code: 400 }
```

The relevant part from the config file is:
```
# Enable IDevID and IAK usage and set their algorithms.
# Choosing a template will override the name and asymmetric algorithm choices.
# Templates are specified in the TCG document found here, section 7.3.4:
# https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_
#
# Accepted values:
# iak_idevid_asymmetric_alg:   rsa, ecc
# iak_idevid_name_alg:        sha256, sm3_256, sha384, sha512
# iak_idevid_template:        H-1, H-2, H-3, H-4, H-5
# Leave template as "" in order to use asymmetric and name algorithm options
enable_iak_idevid = true
iak_idevid_asymmetric_alg = "ecc"
iak_idevid_name_alg = "sha384"
iak_idevid_template = "H-3"

# The name of the file containing the X509 IAK certificate.
# If set as "default", the "iak-cert.crt" value is used
# If a relative path is set, it will be considered relative from the keylime_dir.
# If an absolute path is set, it is used without change.
#
# To override iak_cert, set KEYLIME_AGENT_IAK_CERT environment variable.
iak_cert = "/root/iak.crt"

# The name of the file containing the X509 IDevID certificate.
# If set as "default", the "idevid-cert.crt" value is used
# If a relative path is set, it will be considered relative from the keylime_dir.
# If an absolute path is set, it is used without change.
#
# To override idevid_cert, set KEYLIME_AGENT_IDEVID_CERT environment variable.
idevid_cert = "/root/idevid.crt"
```

The two files are here:

```
# ls -l *.crt
-rw-r--r--. 1 root root 843 Nov  2 17:50 iak.crt
-rw-r--r--. 1 root root 843 Nov  2 17:50 idevid.crt
# openssl verify -CAfile /var/lib/keylime/tpm_cert_store/bundle.pem iak.crt
iak.crt: OK
# openssl verify -CAfile /var/lib/keylime/tpm_cert_store/bundle.pem idevid.crt
idevid.crt: OK
```

It doesn't seem to read the files, otherwise iak_cert and idevid_cert should not be None on the registrar side. 
I have to set `enable_iak_idevid = true` in the agent config file. I suppose the other parameters do not matter in case of files, though I think I have them correct.

Regarding the other parameters: The better identification for the certs to use would IMO be something like `rsa-2048`, `nist-p256`, `nist-p384`, etc. where each one implies all of RSA or ECC, the name hash that's associated with it, and the template identifier. What would happen if someone was using a wrong combination of RSA + name_alg and template? I also don't think people should have to look at the specs to figure out H-1 vs H-3 etc.

ansasaki:
> The two files are here:
> 
> ```
> # ls -l *.crt
> -rw-r--r--. 1 root root 843 Nov  2 17:50 iak.crt
> -rw-r--r--. 1 root root 843 Nov  2 17:50 idevid.crt
> # openssl verify -CAfile /var/lib/keylime/tpm_cert_store/bundle.pem iak.crt
> iak.crt: OK
> # openssl verify -CAfile /var/lib/keylime/tpm_cert_store/bundle.pem idevid.crt
> idevid.crt: OK
> ```

Shouldn't these certificates be respectively in `/root/iak.crt` and `/root/idevid.crt`? These are the paths you set in the configuration file.

> 
> It doesn't seem to read the files, otherwise iak_cert and idevid_cert should not be None on the registrar side. I have to set `enable_iak_idevid = true` in the agent config file. I suppose the other parameters do not matter in case of files, though I think I have them correct.
> 
> Regarding the other parameters: The better identification for the certs to use would IMO be something like `rsa-2048`, `nist-p256`, `nist-p384`, etc. where each one implies all of RSA or ECC, the name hash that's associated with it, and the template identifier. What would happen if someone was using a wrong combination of RSA + name_alg and template? I also don't think people should have to look at the specs to figure out H-1 vs H-3 etc.

About this, I agree that we shouldn't expect the users to know that `H-1` translates to by heart. At least a table with the meaning of this should be added as comments in the configuration file (and if added, also add to the comment to the templates in `keylime/keylime` repo)

stefanberger:
> > The two files are here:
> > ```
> > # ls -l *.crt
> > -rw-r--r--. 1 root root 843 Nov  2 17:50 iak.crt
> > -rw-r--r--. 1 root root 843 Nov  2 17:50 idevid.crt
> > # openssl verify -CAfile /var/lib/keylime/tpm_cert_store/bundle.pem iak.crt
> > iak.crt: OK
> > # openssl verify -CAfile /var/lib/keylime/tpm_cert_store/bundle.pem idevid.crt
> > idevid.crt: OK
> > ```
> 
> Shouldn't these certificates be respectively in `/root/iak.crt` and `/root/idevid.crt`? These are the paths you set in the configuration file.

They are at these locations. Sorry, didn't show full path:

```
# ls -l /root/*.crt
-rw-r--r--. 1 root root 843 Nov  2 17:50 /root/iak.crt
-rw-r--r--. 1 root root 843 Nov  2 17:50 /root/idevid.crt
```

Isaac-Matthews:
> It doesn't seem to read the files, otherwise iak_cert and idevid_cert should not be None on the registrar side.
I have to set enable_iak_idevid = true in the agent config file. I suppose the other parameters do not matter in case of files, though I think I have them correct.

That seems to be a bug with it not using a custom path to specify the certs, I will take a look at this

> Regarding the other parameters: The better identification for the certs to use would IMO be something like rsa-2048, nist-p256, nist-p384, etc. where each one implies all of RSA or ECC, the name hash that's associated with it, and the template identifier, and also in the future the NVRAM location. What would happen if someone was using a wrong combination of RSA + name_alg and template? I also don't think people should have to look at the specs to figure out H-1 vs H-3 etc.

>About this, I agree that we shouldn't expect the users to know that H-1 translates to by heart. At least a table with the meaning of this should be added as comments in the configuration file (and if added, also add to the comment to the templates in keylime/keylime repo)

I can add a table with more detail in the various configuration files etc. For clarity though, the user does not need to match H-n to any algorithms, if a template is specified (eg. H-3) and the user leaves/incorrectly modifies the name_alg and asymmetric_alg (eg. rsa, sha256) fields they are just ignored anyway as a template is being used, so the name and asym algs specified in the template take precedence. I did mention this in the config file but I can add more info maybe just above the template field for extra clarity.

Isaac-Matthews:
I think currently an IDevID cert would be provided alongside a device on purchase, and would be stored in an inventory system, along with probably somewhere on the device. There is no spec for this as you said so it is up to the user at the moment to know where their certs are and point keylime at them or add them to the keylime directory. As far as I know there would not be multiple sources.

When an NVRAM spec becomes available we will be able to make this much more automatic for users going forward.

Isaac-Matthews:
@stefanberger 
I can't seem to recreate the issue you had. I have tried alternative paths and filenames but still get the registration. Looking back through the rust I don't think it can be a path issue actually but something else going on. Presumably you haven't set your environmental variables for these config options as that could be overwriting?

Isaac-Matthews:
@stefanberger sorry for the spam but I have just had a thought. Could you attach the journalctl logs for the agent? Also are you using DER certs for the IAK and IDevID? PEM will not work (DER is specified by the spec but I suppose we should at least add an informative message even if that isn't the issue if not just enable using PEM IDevIDs and IAKs.)

Isaac-Matthews:
I have added a fix for rsapss, and allowed the use of pem idevid and iak certs. Could you check if you are still getting issues after these changes?

stefanberger:
> I have added a fix for rsapss, and allowed the use of pem idevid and iak certs. Could you check if you are still getting issues after these changes?

I have DER certs that I read from the NVRAM locations of swtpm. The problems are still the same, the registrar only gets `None`.

stefanberger:
It works for me now. Certs need to be put into directories where the keylime user has access to, so /root/iak.crt doesn't work...

stefanberger (approved):
LGTM

keylime-agent/src/crypto.rs
```diff
@@ -75,6 +89,66 @@ pub(crate) fn write_x509(cert: &X509, file_path: &Path) -> Result<()> {
...
+pub(crate) fn check_x509_key(
+    cert: &X509,
+    tpm_key: tss_esapi::structures::Public,
+) -> Result<bool> {
+    // Id:RSA_PSS only added in rust-openssl from v0.10.59; remove this let and use Id::RSA_PSS after update
+    // Id taken from https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/nid.h#4039
+    let id_rsa_pss: Id = Id::from_raw(912);
+    match cert
+        .public_key()
+        .unwrap() //#[allow_ci]
```
ansasaki, line 102: `unwrap()` should never be used as it can lead to a panic which should never happen in any circumstance. We can tolerate it in tests, but not in the actual code.

Sorry for being strict, I know there are still some `unwrap()` lingering around, but let's do our best to not add new `unwrap()` usage.

keylime-agent/src/crypto.rs
```diff
@@ -75,6 +89,66 @@ pub(crate) fn write_x509(cert: &X509, file_path: &Path) -> Result<()> {
...
+        .id()
+    {
+        Id::RSA => {
+            let cert_n =
+                cert.public_key().unwrap().rsa().unwrap().n().to_vec(); //#[allow_ci]
```
ansasaki, line 107: Please, do not use `unwrap()`

keylime-agent/src/crypto.rs
```diff
@@ -75,6 +89,66 @@ pub(crate) fn write_x509(cert: &X509, file_path: &Path) -> Result<()> {
...
+            _ = cert_n_str.remove(0);
+            let key = SubjectPublicKeyInfo::try_from(tpm_key)?;
+            let key_der = picky_asn1_der::to_vec(&key)?;
+            let key_der_str = format!("{:?}", key_der);
+
+            Ok(key_der_str.contains(&cert_n_str))
+        }
+        cert_id if cert_id == id_rsa_pss => {
+            let cert_n =
+                cert.public_key().unwrap().rsa().unwrap().n().to_vec(); //#[allow_ci]
```
ansasaki, line 119: Please, do not use `unwrap()`

keylime-agent/src/crypto.rs
```diff
@@ -75,6 +89,66 @@ pub(crate) fn write_x509(cert: &X509, file_path: &Path) -> Result<()> {
...
+            let key = SubjectPublicKeyInfo::try_from(tpm_key)?;
+            let key_der = picky_asn1_der::to_vec(&key)?;
+            let key_der_str = format!("{:?}", key_der);
+
+            Ok(key_der_str.contains(&cert_n_str))
+        }
+        Id::EC => {
+            let cert_n = cert
+                .public_key()
+                .unwrap() //#[allow_ci]
```
ansasaki, line 132: Please, do not use `unwrap()`

keylime-agent/src/crypto.rs
```diff
@@ -75,6 +89,66 @@ pub(crate) fn write_x509(cert: &X509, file_path: &Path) -> Result<()> {
...
+                .ec_key()
+                .unwrap() //#[allow_ci]
```
ansasaki, line 134: Please, do not use `unwrap()`

keylime-agent/src/crypto.rs
```diff
@@ -75,6 +89,66 @@ pub(crate) fn write_x509(cert: &X509, file_path: &Path) -> Result<()> {
...
+                .public_key_to_der()
+                .unwrap(); //#[allow_ci]
```
ansasaki, line 136: Please, do not use `unwrap()`

keylime-agent/src/main.rs
```diff
@@ -295,6 +295,83 @@ async fn main() -> Result<()> {
...
+                        "Loading IAK certificate from {}",
+                        iak_path.display()
+                    );
+                    let iakcert = match crypto::load_x509_der(iak_path) {
+                        Ok(cert) => cert,
+                        Err(error) => crypto::load_x509_pem(iak_path)?,
+                    };
+                    if crypto::check_x509_key(
+                        &iakcert,
+                        iak.clone().unwrap().public, //#[allow_ci]
```
ansasaki, line 322: Please, do not use `unwrap()`

keylime-agent/src/main.rs
```diff
@@ -295,6 +295,83 @@ async fn main() -> Result<()> {
...
+                    )
+                    .unwrap(/*//#[allow_ci]*/)
```
ansasaki, line 324: Please, do not use `unwrap()`

keylime-agent/src/main.rs
```diff
@@ -295,6 +295,83 @@ async fn main() -> Result<()> {
...
+                        "Loading IDevID certificate from {}",
+                        idevid_path.display()
+                    );
+                    let idevcert = match crypto::load_x509_der(idevid_path) {
+                        Ok(cert) => cert,
+                        Err(error) => crypto::load_x509_pem(idevid_path)?,
+                    };
+                    if crypto::check_x509_key(
+                        &idevcert,
+                        idevid.clone().unwrap().public, //#[allow_ci]
```
ansasaki, line 355: Please, do not use `unwrap()`

keylime-agent/src/main.rs
```diff
@@ -295,6 +295,83 @@ async fn main() -> Result<()> {
...
+                    )
+                    .unwrap(/*//#[allow_ci]*/)
```
ansasaki, line 357: Please, do not use `unwrap()`

ansasaki (changes requested):

ansasaki (approved):

PR #692, opened 2023-12-16
Make use of workspace dependencies

ueno:
This allows us to define several crate dependencies in a single place: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

### DIFF
--- a/keylime-agent/Cargo.toml
+++ b/keylime-agent/Cargo.toml
@@ -1,39 +1,39 @@
 [package]
-authors = ["Keylime Authors"]
-edition = "2021"
 name = "keylime_agent"
-version = "0.2.3"
-license = "Apache-2.0"
 description = "Rust agent for Keylime"
-repository = "https://github.com/keylime/rust-keylime"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+version.workspace = true
 
 [dependencies]
-actix-web =  { version = "4", default-features = false, features = ["macros", "openssl"] }
-base64 = "0.21"
-cfg-if = "1"
-clap = { version = "4.3", features = ["derive"] }
-compress-tools = "0.12"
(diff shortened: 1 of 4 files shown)


keylime-agent/Cargo.toml
```diff
@@ -1,47 +1,47 @@
...
+tempfile.workspace = true
+tokio.workspace = true
+tss-esapi.workspace = true
+thiserror.workspace = true
+uuid.workspace = true
 zmq = {version = "0.9.2", optional = true}
 # wiremock was moved to be a regular dependency because optional
 # dev-dependencies are not supported
 # see: https://github.com/rust-lang/cargo/issues/1596
 wiremock = {version = "0.5", optional = true}
```
aplanas, line 41: Any reason why zmq and wiremock are out of the inheritance?

ansasaki: I think because they are used only in `keylime-agent`, not shared with the other parts (`keylime` and `keylime-ima-emulator`)

ansasaki (approved):

keylime-agent/Cargo.toml
ueno: The reason behind this is that the top-level workspace.dependencies can't include optional dependencies, as described at https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

aplanas: @ueno I can be confused here, so bare with me. But IIUC this is valid:

`zmq = {workspace = ture, optional = true}`

See for example https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

You are right that we cannot do this in the top level, in `[workspace.dependencies]`, where all the dependencies are listed, but I am not sure how relevant is that.  For example, if a consequence of doing this is that during vendoring the optional crates will always be included.

Issue #330, opened 2022-02-15
Better error handling on agent startup


lkatalin:
This is related to #287 

In addition to the help message, we should have more helpful output when the agent is started up but gets a TPM error.

greyspectrum:
I'd like to take this one.

THS-on:
We should also switch the default log level to `info` instead of `error` at least for the agent itself.

Issue #639, opened 2023-08-14
Hide NV index errors from tss binindings when reading the EK certificate


THS-on:
Currently we output all the warnings and errors from the tss-esapi bindings, but when we read the EK certificate we get a bunch of errors, that are not really understandable from users.

Example of the errors can be found here: https://files.slack.com/files-pri/T08PSQ7BQ-F05N5ALHBAL/tpm_question_2.png

PR #711, opened 2024-01-22
tpm: Overhaul errors

ansasaki:
Add granular error types for every possible error raised by the `tpm` module. This improves diagnosing the cause of the errors without showing complex errors coming from underlying libraries.
    
This also drops all `expect()` and `unwrap()` uses to avoid panicking.

Fixes: #330, #639


aplanas (approved):
Very interesting use of map_err with an enum.

PR #736, opened 2024-02-06
crypto: Improve error handling and move to library

ansasaki:
Wrap every possible error in `crypto.rs` in a single error type `CryptoError`.
Moved `AES_128_KEY_LEN`, `AES_256_KEY_LEN`, and `AES_BLOCK_SIZE` definitions from `common.rs` to `crypto.rs`
Marked the source of the errors in `tpm.rs`
Move the `crypto` module from `keylime-agent` to the `keylime` library

Also add some auxiliary functions and tests for them:

- `x509_get_pubkey()`
- `x509_to_pem()`
- `x509_to_der()`
- `tss_pubkey_to_pem()`
- `hash()`

Rename:

- `kdf()` to `pbkdf2()`
- `load_x509()` to `load_x509_pem()`

Bump `ahash` to version `0.8.7` to avoid compilation error.

Fixes #648

PR #530, opened 2023-02-27
agent: move from compress tools to zip

THS-on:
compress tools uses libarchive which pulls a lot of compression algorithms as shared library. 
This changes it to use the zip create which uses less:

New:
```
	linux-vdso.so.1 (0x00007ffd877a5000)
	libbz2.so.1.0 => /usr/lib/libbz2.so.1.0 (0x00007fa0e2411000)
	libtss2-esys.so.0 => /usr/lib/libtss2-esys.so.0 (0x00007fa0e237a000)
	libtss2-tctildr.so.0 => /usr/lib/libtss2-tctildr.so.0 (0x00007fa0e2370000)
	libtss2-mu.so.0 => /usr/lib/libtss2-mu.so.0 (0x00007fa0e0bb1000)
	libssl.so.3 => /usr/lib/libssl.so.3 (0x00007fa0e0b11000)
	libcrypto.so.3 => /usr/lib/libcrypto.so.3 (0x00007fa0e0600000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007fa0e234e000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007fa0e0518000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fa0e0331000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007fa0e246a000)
	libtss2-sys.so.1 => /usr/lib/libtss2-sys.so.1 (0x00007fa0e0ae8000)
```

Old:
```
	linux-vdso.so.1 (0x00007ffe6ef78000)
	libarchive.so.13 => /usr/lib/libarchive.so.13 (0x00007f5543936000)
	libtss2-esys.so.0 => /usr/lib/libtss2-esys.so.0 (0x00007f554389f000)
	libtss2-tctildr.so.0 => /usr/lib/libtss2-tctildr.so.0 (0x00007f5545160000)
	libtss2-mu.so.0 => /usr/lib/libtss2-mu.so.0 (0x00007f5543850000)
	libssl.so.3 => /usr/lib/libssl.so.3 (0x00007f55437b0000)
	libcrypto.so.3 => /usr/lib/libcrypto.so.3 (0x00007f5543200000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f5543790000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007f55436a8000)
[...]

PR #739, opened 2024-02-08
cargo: Replace compress-tools with zip crate

### LINKED ISSUE #671: Bug: "distroless" local docker build fails
Bug description
---

The keylime agent is capable of building its own docker container, using a distroless image, or fedora, or wolfi. The distroless build option does not work (fails while building)

Root cause
---

The problem seems to be the last command, creating a `keylime` user with `tss` group membership. My interpretation of this bug is that the command fails because there is no `/bin/sh` to run the command; to not even mention the `useradd` command which is also not part of distroless.

```
------
Dockerfile.distroless:99
--------------------
  97 |     
  98 |     # Create a system user 'keylime' to allow dropping privileges
  99 | >>> RUN useradd -s /sbin/nologin -r -G tss keylime
 100 |     
 101 |     # run as root by default
--------------------
ERROR: failed to solve: process "/bin/sh -c useradd -s /sbin/nologin -r -G tss keylime" did not complete successfully: exit code: 1
Error response from daemon: No such image: keylime_agent:latest-distroless
```

Possible fixes
---

1) copy /bin/bash and other commands from a busybox type container image, and use them. Not ideal because we don't actually need any shell in the final container; it's bloat and a security violation.
2) prepare `/etc/passwd` and `/etc/group` files ahead of time and copy them into place in the container. Better than the previous option but leaves open the question of what user and group IDs to use.
3) forego creation of keylime and tss user and group, since container-local group names and IDs don't really matter.

### LINKED ISSUE #733: update outdated compress-tools dependency (v0.12 -> v0.14) or port to the "zip" crate?
I see that there was a previous effort to drop the compress-tools / libarchive dependency: https://github.com/keylime/rust-keylime/pull/530

If you don't need support for reading/writing AES encrypted zip files (or any other of the "fancy" features that are part of the default feature set), you could just depend on the `zip` crate with `default-features = false` (and add whatever features you actually need on top). That would drop all the crypto related dependencies from Cargo.lock / your dependency tree compared to this PR.

ansasaki:
Replace `compress-tools` with `zip` crate
Remove mentions of `libarchive` dependency from `README`, as well as from rpm spec files and dockerfiles
Remove keylime user creation from non-fedora based dockerfiles.
Add wget installation to `wolfi` based image.

Fixes #733 
Fixes #671

### DIFF
--- a/docker/release/Dockerfile.wolfi
+++ b/docker/release/Dockerfile.wolfi
@@ -13,7 +13,7 @@ RUN apk update
 RUN apk add --no-cache --update-cache \
     rust \
     make pkgconf gcc glibc glibc-dev openssl openssl-dev posix-libc-utils \
-    clang-15 llvm15 wget
+    clang-17 clang-17-dev wget
 
 # Install tpm2-tss (dependency for the tss-esapi crate)
 WORKDIR /src
--- a/docker/release/Dockerfile.wolfi
+++ b/docker/release/Dockerfile.wolfi
@@ -10,15 +10,10 @@ RUN apk update
 # - install rust
 # - install gcc and others to compile tpm2-tss
 # - we are using the "generate-bindings" feature for the tss-esapi crate which requires clang/llvm
-# - Install libarchive (dependency for the compress-tools crate)
 RUN apk add --no-cache --update-cache \
     rust \
     make pkgconf gcc glibc glibc-dev openssl openssl-dev posix-libc-utils \
-    clang-15 llvm15 \
-    libarchive libarchive-dev
-
-# there is a bug in the libarchive.pc file which wrongly adds iconv
-RUN sed -i "s/iconv //" /usr/lib/pkgconfig/libarchive.pc
+    clang-15 llvm15 wget
 
 # Install tpm2-tss (dependency for the tss-esapi crate)
 WORKDIR /src
@@ -69,19 +64,10 @@ COPY --from=builder \
     /usr/lib/libssl.so* \
     /usr/lib/
 
-# libarchive is a direct dependency for the compress-tools crate, so we must copy itself and all its dependencies
+# libz is a direct dependency for the zip crate
 COPY --from=builder \
-    /lib/libacl.so* \
-    /lib/libattr.so* \
     /lib/libz.so* \
     /lib/
-COPY --from=builder \
-    /usr/lib/libarchive.so* \
-    /usr/lib/libexpat.so* \
-    /usr/lib/liblzma.so* \
-    /usr/lib/libzstd.so* \
-    /usr/lib/libbz2.so* \
-    /usr/lib/
 
 # tpm2-tss libraries are a dependency (probably not all of them, but we just copy all)
 # because we are using the tss-esapi crate which is essentially just a wrapper around those (unfortunately)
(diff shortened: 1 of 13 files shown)


ansasaki:
This is very similar to previous attempt from @THS-on on #530

docker/release/Dockerfile.wolfi
```diff
@@ -10,15 +10,10 @@ RUN apk update
...
-# - Install libarchive (dependency for the compress-tools crate)
 RUN apk add --no-cache --update-cache \
     rust \
     make pkgconf gcc glibc glibc-dev openssl openssl-dev posix-libc-utils \
-    clang-15 llvm15 \
-    libarchive libarchive-dev
-
-# there is a bug in the libarchive.pc file which wrongly adds iconv
-RUN sed -i "s/iconv //" /usr/lib/pkgconfig/libarchive.pc
+    clang-15 llvm15 wget
```
aplanas, line 16: Just a note. The last stable version raised llvm to 16: https://releases.rs/docs/1.76.0/#compatibility-notes

rpm/fedora/rust-keylime-metadata.patch
```diff
@@ -1,28 +1,6 @@
...
-+picky-asn1-x509 = "0.7"
- pretty_env_logger = "0.4"
- reqwest = {version = "0.11", default-features = false, features = ["json"]}
- serde = "1.0.80"
-@@ -31,7 +31,7 @@ serde_json = { version = "1.0", features = ["raw_value"] }
- static_assertions = "1"
- tempfile = "3.4.0"
- tokio = {version = "1.24", features = ["rt", "sync"]}
--tss-esapi = {version = "7.2.0", features = ["generate-bindings"]}
-+tss-esapi = {version = "7", features = ["generate-bindings"]}
```
aplanas, line 21: so fedora has now newer crates?

aplanas (commented):
LGTM, the only question is the test fails in Fedora

THS-on (approved):
Rust code changes LGTM. 

At some point I should look again at building a fully static agent, but this will also require some changes to tss library to allow static linking.

rpm/fedora/rust-keylime-metadata.patch
ansasaki: Yes, Fedora now has the `tss-esapi` version `7.4.0`

docker/release/Dockerfile.wolfi
ansasaki: Thanks, I'll update it

ansasaki:
> LGTM, the only question is the test fails in Fedora

Yes, it is currently failing due to the `ahash` crate, which is failing to build with nightly `rustc`. I proposed a workaround in the tests repository via https://github.com/RedHat-SP-Security/keylime-tests/pull/550, but we also need to bump the `ahash` crate.

I'll extract the commit where I bump the `ahash` crate on my other open PR (https://github.com/keylime/rust-keylime/pull/736/commits/d210efdbb7f0f166eded5a1a6b2a0bbdc8385392) and create a dedicated PR

aplanas (approved):

ansasaki:
The proposed workaround in tests repo PR (https://github.com/RedHat-SP-Security/keylime-tests/pull/550) was merged and the tests re-run passed, making this ready to merge

PR #747, opened 2024-02-22
crypto: Add support for ECC key pair generation

ansasaki:
Also, implement some unit tests in `tpm.rs`


keylime/src/crypto.rs
```diff
@@ -577,11 +594,31 @@ fn rsa_generate(key_size: u32) -> Result<PKey<Private>, CryptoError> {
     .map_err(CryptoError::PKeyFromRSAError)
 }
 
+/// Generate RSA key pair with the given size
+///
+/// Returns the a tuple containing the PKey<Public> and PKey<Private>
```
ueno, line 599: typo: "the a"

keylime/src/crypto.rs
```diff
@@ -577,11 +594,31 @@ fn rsa_generate(key_size: u32) -> Result<PKey<Private>, CryptoError> {
...
 ) -> Result<(PKey<Public>, PKey<Private>), CryptoError> {
     let private = rsa_generate(key_size)?;
     let public = pkey_pub_from_priv(private.clone())?;
+
+    Ok((public, private))
+}
+
+/// Generate a ECC key pair on given group
+///
+/// Returns the a tuple containing the PKey<Public> and PKey<Private>
```
ueno, line 611: typo: "the a"

keylime/src/crypto.rs
```diff
@@ -607,6 +644,18 @@ fn pkey_pub_from_priv(
...
+        Id::EC => {
+            let ec_key = privkey
+                .ec_key()
+                .map_err(CryptoError::ECGetPrivateKeyError)?;
+
+            let pubkey =
+                EcKey::from_public_key(ec_key.group(), ec_key.public_key())
+                    .map_err(CryptoError::ECKeyFromPublicPointError)?;
+
+            Ok(PKey::from_ec_key(pubkey)
```
ueno, line 656: I guess you don't need to wrap this with `Ok(...?)`.

keylime/src/crypto.rs
```diff
@@ -1466,4 +1512,50 @@ mod tests {
         let r = generate_tls_context(&loaded_a, &privkey, loaded_list);
         assert!(r.is_ok());
     }
+
+    #[test]
+    fn test_x509_rsa() {
+        for length in [2048, 3072, 4096] {
```
ueno, line 1518: I guess this would increase CI running time rather significantly; do we really want to check 3072+ bits key generation? If so maybe you might want to add [`#[ignore]`](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested).

keylime/src/tpm.rs
```diff
@@ -1929,18 +1993,251 @@ pub mod testing {
...
+    }
+
+    #[test]
+    fn test_activate_credential() {
+        let mut ctx = Context::new().unwrap(); //#[allow_ci]
+        let r = ctx.create_ek(EncryptionAlgorithm::Rsa, None);
+        assert!(r.is_ok());
+
+        // Create EK
+        let ek_result = r.unwrap(); //#[allow_ci]
```
ueno, line 2107: This is a matter of taste but I would say it's easier to read if you write:
```suggestion
        let ek_result = ctx.create_ek(EncryptionAlgorithm::Rsa, None)
            .expect("unable to create EK");
```

ueno (approved):
Looks good to me in general; you might want to adjust `.gitleaks.toml` to suppress secret scanning.

keylime/src/crypto.rs
ansasaki: Thanks, I'll fix these

keylime/src/crypto.rs
ansasaki: Thanks!

keylime/src/crypto.rs
ansasaki: I think I'll create a duplicate for the long keys and mark it with `#[ignore]`.

keylime/src/tpm.rs
ansasaki: Thanks, will follow the suggestion!

PR #758, opened 2024-03-05
Adding /agent/info API to agent

mpeters:
This is a new read-only API on the agent to allow for 3rd party integrations (starting with SPIRE) that will need certain information about the agent.

### DIFF
--- a/keylime-agent/src/agent_handler.rs
+++ b/keylime-agent/src/agent_handler.rs
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 Keylime Authors
+
+use crate::common::JsonWrapper;
+use crate::{tpm, Error as KeylimeError, QuoteData};
+use actix_web::{web, HttpRequest, HttpResponse, Responder};
+use base64::{engine::general_purpose, Engine as _};
+use log::*;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub(crate) struct AgentInfo {
+    pub agent_uuid: String,
+    pub tpm_hash_alg: String,
+    pub tpm_enc_alg: String,
+    pub tpm_sign_alg: String,
+    pub ak_handle: u32,
+}
+
+// This is an Info request which gets some information about this keylime agent
+// It should return a AgentInfo object as JSON
+pub async fn info(
+    req: HttpRequest,
+    data: web::Data<QuoteData>,
+) -> impl Responder {
+    debug!("Returning agent information");
+
+    let mut info = AgentInfo {
+        agent_uuid: data.agent_uuid.clone(),
+        tpm_hash_alg: data.hash_alg.to_string(),
+        tpm_enc_alg: data.enc_alg.to_string(),
+        tpm_sign_alg: data.sign_alg.to_string(),
+        ak_handle: data.ak_handle.value(),
+    };
+
+    let response = JsonWrapper::success(info);
+    info!("GET info returning 200 response");
+    HttpResponse::Ok().json(response)
+}
+
+#[cfg(test)]
+#[cfg(feature = "testing")]
+mod tests {
+    use super::*;
+    use crate::common::API_VERSION;
+    use actix_web::{test, web, App};
+
+    #[actix_rt::test]
+    async fn test_agent_info() {
+        let mut quotedata = QuoteData::fixture().unwrap(); //#[allow_ci]
+        quotedata.hash_alg = keylime::algorithms::HashAlgorithm::Sha256;
+        quotedata.enc_alg = keylime::algorithms::EncryptionAlgorithm::Rsa;
+        quotedata.sign_alg = keylime::algorithms::SignAlgorithm::RsaSsa;
+        quotedata.agent_uuid = "DEADBEEF".to_string();
+        let data = web::Data::new(quotedata);
+        let mut app =
+            test::init_service(App::new().app_data(data.clone()).route(
+                &format!("/{API_VERSION}/agent/info"),
+                web::get().to(info),
+            ))
+            .await;
+
+        let req = test::TestRequest::get()
+            .uri(&format!("/{API_VERSION}/agent/info"))
+            .to_request();
+
+        let resp = test::call_service(&app, req).await;
+        assert!(resp.status().is_success());
+
+        let result: JsonWrapper<AgentInfo> = test::read_body_json(resp).await;
+        assert_eq!(result.results.agent_uuid.as_str(), "DEADBEEF");
+        assert_eq!(result.results.tpm_hash_alg.as_str(), "sha256");
+        assert_eq!(result.results.tpm_enc_alg.as_str(), "rsa");
+        assert_eq!(result.results.tpm_sign_alg.as_str(), "rsassa");
+    }
+}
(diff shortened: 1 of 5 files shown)


keylime-agent/src/errors_handler.rs
```diff
@@ -166,6 +166,39 @@ pub(crate) async fn quotes_default(req: HttpRequest) -> impl Responder {
...
+    req: HttpRequest,
+) -> impl Responder {
+    let error;
+    let response;
+    let message;
+
+    match req.head().method {
+        http::Method::GET => {
+            error = 400;
+            message = "URI not supported, only /info is supported for GET in /agent/ interface";
```
aplanas, line 179: you removed the trailing "/", maybe you want to remove it here too

keylime-agent/src/errors_handler.rs
```diff
@@ -166,6 +166,39 @@ pub(crate) async fn quotes_default(req: HttpRequest) -> impl Responder {
...
+            response = HttpResponse::BadRequest()
+                .json(JsonWrapper::error(error, message));
+        }
+        _ => {
+            error = 405;
+            message = "Method is not supported in /agent/ interface";
```
aplanas, line 185: same

ansasaki:
The feature seems straightforward.

About the test issues, it seems the nightly rust compiler changed and is now failing to compile. Let's see if that gets fixed "automagically" or if we will have to make changes to the compiler, probably pinning the version with the coverage measurement support instead of using the nightly version.

The formatting issue normally can be fixed automatically by running `cargo fmt`.

mpeters:
> The formatting issue normally can be fixed automatically by running `cargo fmt`.

Thanks, done.

ansasaki:
I hope this fixes the CI issues: https://github.com/RedHat-SP-Security/keylime-tests/pull/560

keylime-agent/src/agent_handler.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Serialize, Deserialize, Debug, Default)]
+pub(crate) struct AgentInfo {
+    pub agent_uuid: String,
+    pub tpm_hash_alg: String,
+    pub tpm_enc_alg: String,
+    pub tpm_sign_alg: String,
+    pub ak_handle: u32,
+}
+
+// This is a Info request which getting some information about this keylime agent
```
aplanas, line 24: I am not English native, but "which getting" sounds weird. Maybe "which is getting" or "which gets"?

aplanas (approved):

keylime-agent/src/agent_handler.rs
mpeters: You're absolutely right, fixing now.

aplanas (approved):

keylime-agent/src/agent_handler.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::common::JsonWrapper;
+use crate::{tpm, Error as KeylimeError, QuoteData};
+use actix_web::{web, HttpRequest, HttpResponse, Responder};
+use base64::{engine::general_purpose, Engine as _};
+use log::*;
+use serde::{Deserialize, Serialize};
+use std::{
+    fs::{read, read_to_string},
+    io::{Read, Seek},
+};
```
ueno, line 13: None of these under `std::fs` and `std::io` seems to be used in this file.

ueno (approved):
Looks good to me!

ansasaki:
Should the API version be bumped due to this addition?

mpeters:
> Should the API version be bumped due to this addition?

Yes. I don't know what process we want for that. Do we want to bump the version when we make a single change? Or before we do a release? I'm open to either one.

ansasaki:
> > Should the API version be bumped due to this addition?
> 
> Yes. I don't know what process we want for that. Do we want to bump the version when we make a single change? Or before we do a release? I'm open to either one.

I think the best would be to bump the version with the first API change made in the release, but do a single bump including all the changes made to the API per release.

This would better test the bump itself during the release development. It would be risky and error prone to bump the version only when creating the new release.

mpeters:
@ansasaki It's an interesting problem with the version numbers: If I bump them in the code in the agent not only will the agent's URLs be /v2.2 but it will be trying to use the /v2.2 APIs of the registrar and verifier which don't yet exist. Now in my case I do have a companion PR (https://github.com/keylime/keylime/pull/1532) on the server side so I'll go ahead and bump up the versions there too... but what do we do if it's just a change in the agent's APIs? Do we have to have a "dummy" PR in the server side to bump up the APIs first there before we can do it in the agent?

ansasaki:
> t what do we do if it's just a change in the agent's APIs? Do we have to have a "dummy" PR in the serve

For now, I think the easiest way would be to add the new supported version there and then bump the version here. We should discuss with the community what will be the process, and also think if it is worth to increase the complexity of the agent to add more than one supported API version.

We agreed in the past that the agent would provide only the latest API version, while the python components would handle the various possible supported agent API's.

I also hit this problem when updating some tests and created an issue to track this: #790 

We should consider dissociating the API versions, but that will come with the need to manage on the agent side which versions of the registrar API are supported. Not an easy decision.

keylime-agent/src/agent_handler.rs
```diff
@@ -0,0 +1,80 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Default)]
```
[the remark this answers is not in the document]
mpeters: good catch, removing

keylime-agent/src/agent_handler.rs
mpeters: good catch, removing

ansasaki:
@kkaarreell It seems the new version of the coverage tool `cargo-tarpaulin` cannot be compiled with the `rustc` version we pinned in the test (`1.74.1`). The (maybe) good news is that it can be compiled with the current `rustc` version in Fedora (`1.78.0`). I think we will have to adjust the test `install_upstream_rust_keylime` test again :disappointed: 

@mpeters There are warnings that are generated by the new rust compiler, I proposed a fix for them in #793. Meanwhile, the tests in the CI will probably continue to fail. I tested your code manually on top of #793 and the test you added passes.

kkaarreell:
I think that the build on F39 is failing due to some repodata (cache) issue. It is installing old i686 version of tpm2-tss-devel from fedora repo instead of installing it from updates repo.
```
tpm2-tss-4.0.1-6.fc39.x86_64
tpm2-tss-devel                    i686    4.0.1-4.fc39                     fedora
```

kkaarreell:
You could try editing `packit-ci.fmf` and adding explicit call of
`dnf makecache` to https://github.com/mpeters/rust-keylime/blob/spire-apis/packit-ci.fmf#L19
It could fix this problem, though prolong test execution a bit.

mpeters:
> You could try editing `packit-ci.fmf` and adding explicit call of `dnf makecache` to https://github.com/mpeters/rust-keylime/blob/spire-apis/packit-ci.fmf#L19 It could fix this problem, though prolong test execution a bit.

@kkaarreell Is this before or after the calls to disable `dnf-makecache.service` and `dnf-makecache.timer`?

kkaarreell:
Hm, it seems it didnt help, I'm sorry. Weird.

kkaarreell:
I think I know whats going on. F39 image Is old and already contains old tpm2-tss package. Therefore the -devel subpackage gets installed in the same old version. Either we would have to explicitly update or ask Testing Farm to create newer image. I will do the later tomorrow for sure.

kkaarreell:
Hi Michael, try rescheduling it once more. In https://github.com/keylime/rust-keylime/pull/795 CI has passed for me even without an explicit makecache call.

PR #761, opened 2024-03-15
crypto: Enable TLS 1.3

ansasaki:
The agent server used the `SslAcceptor::mozilla_intermediate()` which disables TLS 1.3 and restricts the TLS 1.2 ciphers to:

```
TLSv1.2:
 ciphers:
   TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519)
   TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (ecdh_x25519)
   TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519)
   TLS_DHE_RSA_WITH_AES_128_GCM_SHA256 (dh 2048)
   TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 (dh 2048)
   TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 (ecdh_x25519)
   TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384 (ecdh_x25519)
   TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA (ecdh_x25519)
   TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA (ecdh_x25519)
   TLS_DHE_RSA_WITH_AES_128_CBC_SHA256 (dh 2048)
   TLS_DHE_RSA_WITH_AES_128_CBC_SHA (dh 2048)
   TLS_DHE_RSA_WITH_AES_256_CBC_SHA256 (dh 2048)
   TLS_DHE_RSA_WITH_AES_256_CBC_SHA (dh 2048)
   TLS_RSA_WITH_AES_128_GCM_SHA256 (rsa 4096)
   TLS_RSA_WITH_AES_256_GCM_SHA384 (rsa 4096)
   TLS_RSA_WITH_AES_128_CBC_SHA256 (rsa 4096)
   TLS_RSA_WITH_AES_256_CBC_SHA256 (rsa 4096)
   TLS_RSA_WITH_AES_128_CBC_SHA (rsa 4096)
   TLS_RSA_WITH_AES_256_CBC_SHA (rsa 4096)
```

With this change TLS 1.3 is enabled. The following ciphers are accepted:
```
TLSv1.2:
 ciphers:
   TLS_DHE_RSA_WITH_AES_128_GCM_SHA256 (dh 2048)
   TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 (dh 2048)
   TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (secp256r1)
   TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (secp256r1)
   TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (secp256r1)
TLSv1.3:
 ciphers:
   TLS_AKE_WITH_AES_128_GCM_SHA256 (ecdh_x25519)
   TLS_AKE_WITH_AES_256_GCM_SHA384 (ecdh_x25519)
   TLS_AKE_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519)
```

### DIFF
--- a/keylime/src/crypto.rs
+++ b/keylime/src/crypto.rs
@@ -655,7 +655,7 @@ pub fn generate_tls_context(
     key: &PKey<Private>,
     ca_certs: Vec<X509>,
 ) -> Result<SslAcceptorBuilder, CryptoError> {
-    let mut ssl_context_builder = SslAcceptor::mozilla_intermediate(
+    let mut ssl_context_builder = SslAcceptor::mozilla_intermediate_v5(
         SslMethod::tls(),
     )
     .map_err(|source| CryptoError::SSLContextBuilderError {


keylime/src/crypto.rs
```diff
@@ -785,7 +785,7 @@ pub fn generate_tls_context(
     key: &PKey<Private>,
     ca_certs: Vec<X509>,
 ) -> Result<SslAcceptorBuilder, CryptoError> {
-    let mut ssl_context_builder = SslAcceptor::mozilla_intermediate(
+    let mut ssl_context_builder = SslAcceptor::mozilla_intermediate_v5(
```
ueno, line 788: I wonder if it could take into account of system-wide configuration instead of hard-coding the selected ciphersuites. In Fedora, for example, if the ciphersuites are not explicitly set (or set to `PROFILE=SYSTEM`) it will look at `/etc/crypto-policies/back-ends/opensslcnf.conf`.

ansasaki: I will investigate if such system-wide configuration is possible.

ansasaki: From my initial investigation, the way to go to follow system-wide configuration would be to detect if the configuration is available and then manually parse (maybe using the [`openssl::conf::Conf`](https://docs.rs/openssl/latest/openssl/conf/struct.Conf.html)) and then set manually using [`openssl::ssl::SslContextBuilder`](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html).

@ueno What do you think about this approach? Do you have a better suggestion?

Ideally we would simply use the default `openssl` setting which would do all this work by itself.

ueno: Thanks for looking into it. This might be a downstream issue as the `PROFILE=SYSTEM` is not merged in the upstream yet, so if there is a way to switch to the default `openssl` setting (maybe through a configuration option?) that might be sufficient.

ansasaki: There is no easy way to use the default `openssl` configuration.

The best would be if there was the possibility to obtain the default configuration from the system using something similar to [SslAcceptor::mozilla_intermediate_v5](https://docs.rs/openssl/latest/openssl/ssl/struct.SslAcceptor.html#method.mozilla_intermediate_v5), but for the default configuration.

From the comments in the code, these `SslAcceptorBuilders` were created to avoid using the openssl default configuration, which is considered unsafe. I believe it will be complicated to convince `rust-openssl` upstream to accept a PR adding the possibility of using the default options there.

Parsing the openssl configuration manually and setting the found options using [openssl::ssl::SslContextBuilder](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html) one by one seems to be the only option, but parsing the configuration manually is also not facilitated. I don't like the idea of doing this in keylime.

And on top of this, the `actix_web::HttpServer::bind_openssl()` expects an `SslAcceptorBuilder`.

@ueno I propose we merge this PR as it is (which I see as an improvement over the current configuration), and create an issue to track using the system-wide configuration while we investigate what is the best way to do that.

ueno (approved):

Issue #755, opened 2024-02-29
Make the certificate generation to follow builder pattern


ansasaki:
This is a follow up to the changes proposed on #744. The goal is to make the certificate generation to follow the builder pattern to add information/extensions to the certificate.

Issue #753, opened 2024-02-27
Adapt Github actions to avoid Node1[2,6] based ones and upgrade to Node20 based ones


sarroutbi:
Github actions being used are somehow deprecated, taking into account that Spring 2024 is the recommended date to migrate Github actions to versions using Node 20. More information here:
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

For this reason, next Github actions should be migrated:
https://github.com/keylime/rust-keylime/actions/runs/8018008405
https://github.com/keylime/rust-keylime/actions/runs/7914403416
https://github.com/keylime/rust-keylime/actions/runs/8018008411
https://github.com/keylime/rust-keylime/actions/runs/8008246302
https://github.com/keylime/rust-keylime/actions/runs/8008224564

Issue #583, opened 2023-05-04
RFE: Format of setting up IPv6 in agent conf.


Koncpa:
If you want use IPv6 stack for all keylime component you need to set up IPv6 in agent.conf with bracket format like ([::1]), but for other components can paste just IPv6 address.  It could be more user friendly to set up IPv6 in similar format for all keylime components. 

Is it possible to implement for rust keylime in same format as for python components? There is [thread](https://github.com/keylime/keylime/issues/1328), where we disscuss it on python project and there is actual prepared [test](https://github.com/RedHat-SP-Security/keylime-tests/pull/380) for IPv6. WDYT?

PR #765, opened 2024-03-22
config: Support IPv6 with or without brackets

ansasaki:
This adds support for IPv6 addresses in the configuration with or without brackets.
The brackets are removed when the addresses are parsed and then added back when necessary.

This fixes the addition of IPv6 addresses in the mTLS certificate.

This also implements the builder pattern for the certificate generation.

Fixes: #583
Fixes: #753
Fixes: #755

### DIFF
--- a/keylime/src/crypto.rs
+++ b/keylime/src/crypto.rs
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2021 Keylime Authors
 
+pub mod x509;
+
 use base64::{engine::general_purpose, Engine as _};
 use log::*;
 use openssl::{
-    asn1::Asn1Time,
     ec::{EcGroupRef, EcKey},
     encrypt::Decrypter,
     hash::MessageDigest,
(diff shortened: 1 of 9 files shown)


keylime/src/crypto.rs
```diff
@@ -1471,22 +1342,27 @@ mod tests {
...
 
-        let r = generate_x509(
-            &privkey,
-            "uuidB",
-            Some(vec!["1.2.3.4".to_string(), "1.2.3.5".to_string()]),
-        );
+        let r = CertificateBuilder::new()
+            .private_key(privkey)
+            .common_name("uuidB")
+            .add_ips(vec!["1.2.3.4".to_string()])
```
sarroutbi, line 1363: Why has 1.2.3.5 address been removed in this call? The idea is to test with more than one IP in the vector. Apart from that, I recommend adding also a test with a list of IPv6 tests, if possible

ansasaki: Thanks for the review! I moved the unit tests to `keylime/crypto/x509.rs`. I also added a test using IPv6 there.

keylime/src/crypto/x509.rs
```diff
@@ -0,0 +1,592 @@
...
+            }
+        }
+        self
+    }
+
+    /// Set additional IPs to add to the Subject Alternative Name
+    ///
+    /// # Arguments:
+    ///
+    /// * ips: (Vec<String>): A Vec<String> containing IPs to add to the certificate Subject
```
ueno, line 175: Does this need to be owned? If this is `&[&str]`, the callers don't need to create owned values by themselves: `.add_ips(&vec!["192.168.0.1"])` instead of `.add_ips(vec!["192.168.0.1".to_string()])`.

ueno (approved):
Looks good to me!

Issue #760, opened 2024-03-08
Non-string auth values are not compatible with the agent.


Isaac-Matthews:
If a TPM has been set up with a non-string auth value, for eaxmple by running:
```tpm2_changeauth -c E "hex:0000DF0237BE41C0E0124CAAD0A619B29B8E3EE4173521656729C6F339DBC449"```
the keylime agent is not compatible with the TPM as there is no way provide this value.

The current method for processing auth values set by the config file or environment variables [can be found here](https://github.com/keylime/rust-keylime/blob/2ca4b081f6d2b0c8349e0fe637d112d3b4e19f84/keylime-agent/src/main.rs#L289-L301), this only works for simple string auth values.

A possible fix could be if "hex:" is at the start of a provided auth value it should be interpreted as a hex auth value, this is how tpm2 tools currently works.

PR #769, opened 2024-03-27
Enable hex values to be used for tpm_ownerpassword

Isaac-Matthews:
This fixes [#760](https://github.com/keylime/rust-keylime/issues/760).

If a TPM has been set up with a non-string auth value, for example by running:
```tpm2_changeauth -c E "hex:0000DF0237BE41C0E0124CAAD0A619B29B8E3EE4173521656729C6F339DBC449"```
the keylime agent is not currently compatible with the TPM as there is no way provide this value.

This fixes the issue, by checking for "hex:" at the start of the tpm password and using a hex password if it is found.

### DIFF
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -295,7 +295,15 @@ async fn main() -> Result<()> {
     // ownership of TPM access, which will not be implemented here.
     let tpm_ownerpassword = &config.agent.tpm_ownerpassword;
     if !tpm_ownerpassword.is_empty() {
-        let auth = Auth::try_from(tpm_ownerpassword.as_bytes())?;
+        let auth = if let Some(hex_ownerpassword) =
+            tpm_ownerpassword.strip_prefix("hex:")
+        {
+            let decoded_ownerpassword =
+                hex::decode(hex_ownerpassword).map_err(Error::from)?;
+            Auth::try_from(decoded_ownerpassword)?
+        } else {
+            Auth::try_from(tpm_ownerpassword.as_bytes())?
+        };
         ctx.as_mut().tr_set_auth(Hierarchy::Endorsement.into(), auth)
             .map_err(|e| {
                 Error::Configuration(format!(
--- a/keylime-agent.conf
+++ b/keylime-agent.conf
@@ -263,6 +263,9 @@ idevid_cert = "default"
 # Use this option to state the existing TPM ownerpassword.
 # This option should be set only when a password is set for the Endorsement
 # Hierarchy (e.g. via "tpm2_changeauth -c e").
+# In order to use a hex value for the password, use the prefix "hex:"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by Isaac-Matthews** [comment]
Fixes #760

File: keylime-agent/src/main.rs [hunk 1]
```diff
@@ -295,7 +295,14 @@ async fn main() -> Result<()> {
     // ownership of TPM access, which will not be implemented here.
     let tpm_ownerpassword = &config.agent.tpm_ownerpassword;
     if !tpm_ownerpassword.is_empty() {
-        let auth = Auth::try_from(tpm_ownerpassword.as_bytes())?;
+        let auth = if tpm_ownerpassword.starts_with("hex:") {
+            let (_, hex_ownerpassword) = tpm_ownerpassword.split_at(4);
+            let decoded_ownerpassword = hex::decode(hex_ownerpassword)
+                .expect("Failed to decode tpm_ownerpassword hex");
```
THS-on, line 301: expect works similarly to unwrap. Can you change this to an error that is returned?

THS-on (changes requested):

keylime-agent/src/main.rs
Isaac-Matthews: Yes I have now changed this, thanks!

keylime-agent/src/main.rs
```diff
@@ -295,7 +295,18 @@ async fn main() -> Result<()> {
     // ownership of TPM access, which will not be implemented here.
     let tpm_ownerpassword = &config.agent.tpm_ownerpassword;
     if !tpm_ownerpassword.is_empty() {
-        let auth = Auth::try_from(tpm_ownerpassword.as_bytes())?;
+        let auth = if tpm_ownerpassword.starts_with("hex:") {
+            let (_, hex_ownerpassword) = tpm_ownerpassword.split_at(4);
```
ueno, line 299:
```suggestion
        let auth = if let Some(hex_ownerpassword) = tpm_ownerpassword.strip_prefix("hex:") {
```

keylime-agent/src/main.rs
```diff
@@ -295,7 +295,18 @@ async fn main() -> Result<()> {
...
+            let decoded_ownerpassword = hex::decode(hex_ownerpassword)
+                .map_err(|_| {
```
ueno, line 301: Not sure if this manual conversion is required, as `FromHex` error is already defined.

ueno (approved):
A couple of minor nits, but otherwise it looks good to me.

Isaac-Matthews:
> A couple of minor nits, but otherwise it looks good to me.

Thanks for the feedback, I have made both changes and merged into one commit.

ansasaki:
@THS-on Hi, could you please mark your request as solved? Otherwise I cannot merge this.

THS-on (approved):
@ansasaki done

PR #784, opened 2024-05-06
rpm: fix COPR RPMs build for centos-stream-10

ansasaki:
Fix macro usage on centos-stream-10 build


sergio-correia (approved):
LGTM, thanks

PR #785, opened 2024-05-07
Add the ability to use persistent handles for IDevID and IAK rather than regenerating, and add auth values for them

Isaac-Matthews:
This PR adds into the config the ability to specify handles and auth values for IDevID and IAK keys. The auth values function in the same way as the TPM ownerpassword in that they can be [hex values](https://github.com/keylime/rust-keylime/pull/769).

The new config options are:
```
idevid_password = ""
idevid_handle = ""

iak_password = ""
iak_handle = ""
```

### DIFF
--- a/keylime-agent.conf
+++ b/keylime-agent.conf
@@ -227,7 +227,10 @@ tpm_signing_alg = "rsassa"
 # To override ek_handle, set KEYLIME_AGENT_EK_HANDLE environment variable.
 ek_handle = "generate"
 
-# Enable IDevID and IAK usage and set their algorithms.
+# Enable IDevID and IAK usage 
+enable_iak_idevid = true
+
+# Select IDevID and IAK templates or algorithms for regenerating the keys.
 # By default the template will be detected automatically from the certificates. This will happen if iak_idevid_template is left empty or set as "default" or "detect".
 # Choosing a template will override the name and asymmetric algorithm choices. To use these choices, set iak_idevid_template to "manual"
 # Templates are specified in the TCG document found here, section 7.3.4: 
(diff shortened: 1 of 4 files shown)


keylime/src/tpm.rs
```diff
@@ -613,6 +624,99 @@ impl Context {
...
+    /// Load a key handle from a string of the handle location
+    /// If a password is supplied, authorise the handle
+    /// # Arguments
+    ///
+    /// `handle` : The string of the handle, eg. from config
+    /// `password` ; The string password, to be converted to hex if there is the "hex:" prefix
+    ///
+    /// # Return
+    /// The corresponding KeyHandle, or a TPMError
+    pub fn get_key_handle(
```
ansasaki, line 636: I see you declared this method as public, but it is currently only used to get the handles for IDevID and IAK keys. 

Do you see future usage of this in other use case? If not, could you make it private by removing the `pub` keyword?

keylime-agent.conf
```diff
@@ -237,12 +240,19 @@ ek_handle = "generate"
...
-enable_iak_idevid = false
 iak_idevid_template = "detect"
 # In order for these values to be used, set the iak_idevid_template option to manual
 iak_idevid_asymmetric_alg = "rsa"
 iak_idevid_name_alg = "sha256"
 
+# Alternatively if the keys are persisted, provide the handles for their location below, and optionally their passwords.
+# If handles are provided, they will take priority over templates/algorithms selected above.
+# To use a hex password, use the prefix "hex:" at the start of the password.
+idevid_password = ""
```
ansasaki, line 251: Since you are adding new configuration options, do not forget to make a PR to add the configuration files to the configuration upgrade template.

I'll create an issue when we merge this to remind you.

ansasaki (changes requested):

ansasaki:
@Isaac-Matthews Could you please rebase on top of current master?

keylime/src/tpm.rs
Isaac-Matthews: I thought it could be a useful method but did not have a specific use-case in mind. I'll change to private and we can always change back to public if we want to use it in future.

ansasaki (approved):
Thank you for the contribution!

I created https://github.com/RedHat-SP-Security/keylime-tests/issues/587 to track the end-to-end test creation

PR #861, opened 2024-10-18
Make the API configuration modular

ansasaki:
Move the configuration of the API for each scope into the respective handler file.
Also, move the API configuration into a dedicated file, and make it modular.

This is a preparation to add support for multiple API versions.


THS-on (commented):
Generally LGTM, just would like to have someone else that has more experince with actix to also take a look at it.

aplanas (approved):
LGTM

sergio-correia (approved):
This looks good to me as well, thanks.

PR #862, opened 2024-10-21
Fix showing wrong UUID, remove unused library

UntriexTv:
In case of invalid UUID agent showed:
```
WARN  keylime_agent::config > Misformatted UUID: invalid-uuid-0909
INFO  keylime_agent::config > Using generated UUID: invalid-uuid-0909
```
and warning while compiling:
```
warning: unused import: `std::io::prelude`
    --> keylime/src/tpm.rs:1816:9
     |
1816 |     use std::io::prelude::*;
     |         ^^^^^^^^^^^^^^^^
     |
     = note: `#[warn(unused_imports)]` on by default
```

### DIFF
--- a/keylime-agent/src/config.rs
+++ b/keylime-agent/src/config.rs
@@ -685,7 +685,7 @@ fn get_uuid(agent_uuid_config: &str) -> String {
             Err(_) => {
                 warn!("Misformatted UUID: {}", &uuid_config);
                 let agent_uuid = Uuid::new_v4();
-                info!("Using generated UUID: {}", &uuid_config);
+                info!("Using generated UUID: {}", &agent_uuid);
                 agent_uuid.to_string()
             }
         },


THS-on:
@UntriexTv thanks for the fix. Can you add a signoff line in the commit?  I think you can't drop the import because the tests are no longer building.

THS-on:
@UntriexTv can you also squash your commits together. Once that is done and rest of the test pass this LGTM.

ansasaki:
@UntriexTv thank you for your contribution!

Could you please squash the commits into a single commit? I'd like to avoid adding commits that add a change just to revert it back in the following commit.

UntriexTv:
Sorry for late reply, didn't have much time to check github in past days.

ansasaki (approved):
LGTM!

PR #864, opened 2024-10-24
Use ${DESTDIR} for config

Itxaka:
Prepends ${DESTDIR} for config dir creation and config file copying, so its in line with the other artifacts. Makes no sense to be able to install to dir X but config files being created in the system /etc/ dir

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -31,9 +31,9 @@ clean::
 
 .PHONY: install
 install: all
-	mkdir -p /etc/keylime/
-	mkdir -p /etc/keylime/agent.conf.d
-	cp ${CONFFILE} /etc/keylime/agent.conf
+	mkdir -p ${DESTDIR}/etc/keylime/
+	mkdir -p ${DESTDIR}/etc/keylime/agent.conf.d
+	cp ${CONFFILE} ${DESTDIR}/etc/keylime/agent.conf
 	for f in $(programs); do \
 		install -D -t ${DESTDIR}/usr/bin "$$f"; \
 	done


ansasaki:
> Prepends ${DESTDIR} for config dir creation and config file copying, so its in line with the other artifacts. Makes no sense to be able to install to dir X but config files being created in the system /etc/ dir

Why do you believe it makes no sense to install them at `/etc/`?

It is basically the default location where the Keylime agent binary will search for the configuration file. Changing this location will simply make the configuration file to not be used. Similarly, changing the location where the `systemd` unit files is installed will make `systemd` to not use it and, therefore, not configure the service as expected.

Itxaka:
> > Prepends ${DESTDIR} for config dir creation and config file copying, so its in line with the other artifacts. Makes no sense to be able to install to dir X but config files being created in the system /etc/ dir
> 
> Why do you believe it makes no sense to install them at `/etc/`?
> 
> It is basically the default location where the Keylime agent binary will search for the configuration file. Changing this location will simply make the configuration file to not be used. Similarly, changing the location where the `systemd` unit files is installed will make `systemd` to not use it and, therefore, not configure the service as expected.

I think it makes no sense to allow setting `${DESTDIR}` but then not respect that makes for a very weird choice :D

i.e. If im packaging this as a sysextension, I migth want everything to be installed at /tmp/keylime so then I can build a sysextension from the binaries and a confextension from the files under /etc

Maybe I just want to inspect the config files or where the files are supposed to go because Im building a keylime installer.

So Im not against setting it to /etc but if the user mentions a `${DESTDIR}` and we use that as root for all the artifacts, it makes sense :D Plus the rest of the files also use `${DESTIR}` as root for the locations of the files (binaries and systemd files)

ansasaki:
@Itxaka OK, you convinced me :)

I think it makes sense to allow the user to install all the files to some custom location for inspection.

ansasaki (approved):

PR #870, opened 2024-11-20
Use a single TPM context and avoid race conditions during tests

### LINKED ISSUE #259: Include a straightforward way to use the swtpm
The way we are using it is undocumented and inconsistent with the documentation in `keylime/keylime`. It would be even better if there is a way to set this up the same way each time via an included optional script.

ansasaki:
This implements a singleton to use a single instance of the
`tss_espi::Context`.

This also avoids race conditions by using a thread-safe mutex to access
the context.

The tests were modified to cleanup generated key handles between test
cases, preventing leftover handles to fill the TPM memory.

Introduce a mutex to allow only a single test that create keys in the
TPM to run at once.  It is required that tests that create keys in the
TPM to run the `tpm::testing::lock_tests()` to obtain the mutex and be
able to continue executing without the risk of other tests filling the
TPM memory.

The `QuoteData::fixture` implementation (used only during tests) was
modified to flush the created key contexts when dropped. It also was
modified to lock the global mutex by calling the
`tpm::testing::lock_tests()` and provide the respective guard, preventing
other tests that use the fixture (which necessarily create TPM keys) to
run in parallel.

Note that it is necessary to `drop` the `QuoteData` fixture explicitly to guarantee
that the keys are flushed before releasing the global mutex.

Finally, the following changes were made to the `tests/run.sh`:

* Do not use `tpm2-abrmd` anymore, and use only the `swtpm` socket instead
* Stop the started processes at exit

With all these changes, the `tests/run.sh` can now be executed locally
without any special setting.

This also adds a script to initialize a software TPM locally for testing.

Fixes #259

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -2,27 +2,63 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2021 Keylime Authors
 
+# Store the old TCTI setting
+OLD_TCTI=$TCTI
+OLD_TPM2TOOLS_TCTI=$TPM2TOOLS_TCTI
+
 set -euf -o pipefail
 
-echo "-------- Setting up Virtual TPM"
-mkdir /tmp/tpmdir
+echo "-------- Setting up Software TPM"
+
+# Create temporary directories
+TEMPDIR=$(mktemp -d)
+TPMDIR="${TEMPDIR}/tpmdir"
+mkdir -p ${TPMDIR}
+
+# Manufacture a new Software TPM
 swtpm_setup --tpm2 \
-    --tpmstate /tmp/tpmdir \
+    --tpmstate ${TPMDIR} \
     --createek --decryption --create-ek-cert \
     --create-platform-cert \
+    --lock-nvram \
+    --not-overwrite \
+    --pcr-banks sha256 \
     --display
-swtpm socket --tpm2 \
-    --tpmstate dir=/tmp/tpmdir \
-    --flags startup-clear \
-    --ctrl type=tcp,port=2322 \
-    --server type=tcp,port=2321 \
-    --daemon
-tpm2-abrmd \
-    --logger=stdout \
-    --tcti=swtpm: \
-    --allow-root \
-    --session \
-    --flush-all &
+
+function start_swtpm {
+    # Initialize the swtpm socket
+    swtpm socket --tpm2 \
+        --tpmstate dir=${TPMDIR} \
+        --flags startup-clear \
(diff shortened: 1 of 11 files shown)


tests/run.sh
```diff
@@ -32,13 +59,17 @@ echo "-------- Building"
...
+start_swtpm
 mkdir -p /var/lib/keylime
-TCTI=tabrmd:bus_type=session RUST_BACKTRACE=1 RUST_LOG=info \
+RUST_BACKTRACE=1 RUST_LOG=info \
 KEYLIME_CONFIG=$PWD/keylime-agent.conf \
-cargo test --features testing -- --nocapture
+cargo test --features testing -- --nocapture --test-threads=1
+
+# Flush contexts
+tpm2_flushcontext -t -l -s
```
THS-on, line 69: Why do we need flushing? We should ideally flush everything in the tests. (Most of the things should be flushed by the tss-esapi abstraction anyway)

THS-on (approved):
Had a brief look and LGTM, only a small question

tests/run.sh
ansasaki: Empirically I can affirm that they are not flushed, unless explicitly called in the code. Trust me, I spent hours finding the leaks in our code, and modifying the tests to flush the created keys and sessions.

When running the `swtpm` without a resource manager (`tpm2-abrmd`), then it can only hold 3 keys in memory. This number is easily reached if we run tests in parallel, causing errors due to lack of memory. For that reason I had to add the `--test-threads=1` to run the tests one after the next, allowing the keys to be flushed in between.

I modified the `QuoteData::fixture` implementation to create the AK and EK keys only once. The problem is that I did not find a way to cleanup those keys after the tests are executed. So I had to add the `tpm2_flushcontext`.

BTW, I see I forgot to flush the keys at the end of the script, I'll add that.

I also found a deadlock I added by accident. I'll fix this as well.

ansasaki: I found a way! But there are still some leakages, I'm still searching the sources

ansasaki: In the end I was able to find and cleanup all the points were the tests were leaving leftover key handles and added the cleanup. With this, we don't leave any leftover handle, which makes the `tpm2_flushcontext` there unnecessary. I removed it from the script.

ansasaki:
So I ended up implementing a mutex to let only a single test that creates keys in the TPM to run at once. This allows the tests to run in parallel, in multiple threads. I believe some of the flakyness we saw in the past is now solved.

@THS-on Could you take a look again?

Issue #871, opened 2024-11-21
Also check TPM2TOOLS_TCTI as environment variable to set TCTI


THS-on:
Currently we only allow TCTI environment variable to set the TCTI for the agent. To be compatible with tpm2-tools we should also check  `TPM2TOOLS_TCTI`.

PR #872, opened 2024-11-22
tpm: also parse TPM2TOOLS_TCTI and TEST_TCTI

THS-on:
This aligns envs for TCTI with ruts-tss-esapi and tpm2-tools.

Fixes: #871

### DIFF
--- a/keylime/src/tpm.rs
+++ b/keylime/src/tpm.rs
@@ -472,7 +472,10 @@ impl AsMut<tss_esapi::Context> for Context {
 impl Context {
     /// Creates a connection context.
     pub fn new() -> Result<Self> {
-        let tcti_path = match std::env::var("TCTI") {
+        let tcti_path = match std::env::var("TPM2TOOLS_TCTI")
+            .or_else(|_| std::env::var("TCTI"))
+            .or_else(|_| std::env::var("TEST_TCTI"))
+        {
             Ok(val) => val,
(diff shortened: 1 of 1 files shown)


ansasaki:
In my opinion, this is not correct and even confusing. The agent is not related to and does not use `tpm2-tools` in any way. It would be weird that by changing `tpm2-tools` configuration the agent is also affected.

The `TCTI` configuration should be sufficient, and I believe the confusion comes from proper documentation and better error messages.

THS-on:
Kinda agree, this just aligns it more with what was supported by the old agent. This options is still weird regardless, as its the only option that cannot be set via the config/env combintation. Ideally it would be `KEYLIME_AGENT_TCTI` and the default set as an option in the agent.conf. I can change it to that if you want.

ansasaki:
> Kinda agree, this just aligns it more with what was supported by the old agent. This options is still weird regardless, as its the only option that cannot be set via the config/env combintation. Ideally it would be `KEYLIME_AGENT_TCTI` and the default set as an option in the agent.conf. I can change it to that if you want.

I think that making it a configuration option makes a lot of sense. Adding the `tcti` config option, which would result on `KEYLIME_AGENT_TCTI` env var being supported would be the ideal.

Unfortunately, we would have to still support the plain `TCTI` env var for compatibility. But it is no different from some other options we have.

THS-on:
Ok I'll close this PR then and send a new one to add this as an option + override if TCTI env is set.

PR #883, opened 2024-12-04
tests: Add script to generate IAK and IDevID certificates

ansasaki:
The script generates a test root CA and intermediate CA keys and certificates. The output directory can be set via the `--output` option, and the CA password with the `--pwd` option.  If an output directory is not set, a temporary directory is created and the certificates are placed there.

Then, the script generates the IDevID and IAK keys inside the TPM and respective certificates.

The script requires the `tpm2-openssl` provider to access the keys inside the TPM.

It is recommended to setup the TCTI, TPM2TOOLS_TCTI, and TPM2OPENSSL_TCTI environment variables to select the TPM to use. If not set, the default `/dev/tpmrm0` will be used instead.

This also modifies the `tests/setup_swtpm.sh` and `tests/run.sh` to configure the TPM2OPENSSL_TCTI to use the `swtpm` with the `tpm2-openssl` provider.

This is based on the test contributed by @Isaac-Matthews on https://github.com/RedHat-SP-Security/keylime-tests

### DIFF
--- a/tests/ca.conf
+++ b/tests/ca.conf
@@ -1,0 +1,111 @@
+[ ca ]
+default_ca        = CA_default
+
+[ CA_default ]
+dir               = REPLACE_ROOT_CA_DIR
+certs             = $dir/certs
+crl_dir           = $dir/crl
+new_certs_dir     = $dir/certs
+database          = $dir/index.txt
+serial            = $dir/serial
+RANDFILE          = $dir/.rand
+
+private_key       = $dir/private.pem
+certificate       = $dir/cacert.pem
+
+crlnumber         = $dir/crl/crlnumber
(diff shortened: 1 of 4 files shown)


Isaac-Matthews (approved):
LGTM

sergio-correia (approved):
Thanks, it looks good to me. I would run it through `shellcheck` and work on the "double quote to prevent globbing and word splitting" warnings, as they usually make sense for at least for paths.

ansasaki:
> Thanks, it looks good to me. I would run it through `shellcheck` and work on the "double quote to prevent globbing and word splitting" warnings, as they usually make sense for at least for paths.

Fixed the issues reported by `shellcheck`. Thanks for the review!

PR #886, opened 2024-12-09
Move IAK and IDevID code to a dedicated module

ansasaki:
This moves the code from the `main.rs` file to a new dedicated module in `keylime` library, `keylime::device_id`.

The new module implements the builder pattern in `DeviceIDBuilder` to setup the arguments and then generate an instance of the `DeviceID` structure.

This modifies the `tests/run.sh` script to generate the IAK and IDevID certificates using the TPM state placed in `test-data/tpm-state` (which is generated if not present) and place the resulting certificates in `keylime/test-data/iak-idevid-certs`

### DIFF
--- a/keylime/src/device_id.rs
+++ b/keylime/src/device_id.rs
@@ -9,7 +9,7 @@ use thiserror::Error;
 use tss_esapi::{
     handles::KeyHandle,
     interface_types::algorithm::{AsymmetricAlgorithm, HashingAlgorithm},
-    structures::{Attest, Data, Signature},
+    structures::{Attest, Data, Public as TssPublic, Signature},
 };
 
 #[derive(Error, Debug)]
@@ -22,9 +22,17 @@ pub enum DeviceIDBuilderError {
     #[error("Public key does not match the certificate")]
     CertPubKeyMismatch(#[source] tpm::TpmError),
 
+    /// Failed to flush context
+    #[error("Failed to flush context")]
+    FlushContext(#[source] TpmError),
+
     /// Failed to create IAK
     #[error("Failed to create IAK")]
     IAKCreate(#[source] TpmError),
+
+    /// IAK default Template not set
+    #[error("IAK default template not set")]
+    IAKDefaultTemplateNotSet,
 
     /// Could not get IAK from provided handle
     #[error("Could not get IAK from provided handle")]
@@ -34,10 +42,6 @@ pub enum DeviceIDBuilderError {
     #[error("IAK handle not set in DeviceBuilder. Set the IAK handle with the iak_handle() method from the DeviceIDBuilder object")]
     IAKHandleNotSet,
 
-    /// IAK certificate not set in DeviceIDBuilder
-    #[error("IAK certificate not set in DeviceBuilder. Set the IAK certificate with the iak_cert() method from the DeviceIDBuilder object")]
-    IAKCertNotSet,
-
     /// IAK password not set in DeviceIDBuilder
     #[error("IAK password not set in DeviceBuilder. Set the IAK password with the iak_password() method from the DeviceIDBuilder object")]
     IAKPasswordNotSet,
@@ -46,6 +50,10 @@ pub enum DeviceIDBuilderError {
     #[error("Failed to create IDevID")]
     IDevIDCreate(#[source] TpmError),
 
+    /// IDevID default Template not set
+    #[error("IDevID default template not set")]
+    IDevIDDefaultTemplateNotSet,
+
     /// Could not get IDevID from provided handle
     #[error("Could not get IDevID from provided handle")]
     IDevIDFromHandle(#[source] TpmError),
@@ -53,10 +61,6 @@ pub enum DeviceIDBuilderError {
     /// IDevID handle not set in DeviceIDBuilder
     #[error("IDevID handle not set in DeviceBuilder. Set the IDevID handle with the idevid_handle() method from the DeviceIDBuilder object")]
     IDevIDHandleNotSet,
-
-    /// IDevID certificate not set in DeviceIDBuilder
-    #[error("IDevID certificate not set in DeviceBuilder. Set the IDevID certificate with the idevid_cert() method from the DeviceIDBuilder object")]
-    IDevIDCertNotSet,
 
     /// IDevID password not set in DeviceIDBuilder
     #[error("IDevID password not set in DeviceBuilder. Set the IDevID password with the idevid_password() method from the DeviceIDBuilder object")]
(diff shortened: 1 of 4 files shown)


ansasaki:
Hopefully, the fix in the CI image will be sufficient for the current test failure: https://github.com/keylime/keylime/pull/1699

ansasaki:
The `xxd` tool was missing in the CI image and was added via https://github.com/keylime/keylime/pull/1701

keylime/src/device_id.rs
```diff
@@ -0,0 +1,535 @@
...
+        }
+    }
+
+    /// Get the IDevID from the given handle and password or recreate following the given template
+    ///
+    /// If a handle has been set, try to obtain the IDevID from the handle
+    /// If there is a configured IDevID password, add the password to the handle
+    ///
+    /// If the handle is empty, recreate the IDevId using the provided algorithms
+    fn get_idevid(
```
Isaac-Matthews, line 325: I think if this is recreated here we need to flush it as otherwise we will have 4 keys at once (IAK, IDevID, AK, EK) which can cause an out of memory TPM error.

Isaac-Matthews (changes requested):

keylime/src/device_id.rs
```diff
@@ -0,0 +1,535 @@
...
+        mut self,
+        tpm_ctx: &mut tpm::Context,
+    ) -> Result<DeviceID, DeviceIDBuilderError> {
+        let iak = self.get_iak(tpm_ctx)?;
+        let Some(iak_cert) = self.iak_cert.take() else {
+            unreachable!();
+        };
+
+        // Check that recreated/collected IAK key matches the one in the certificate
+        tpm::check_pubkey_match_cert(&iak.public, &iak_cert, "IAK")
```
Isaac-Matthews, line 420: I realise this is the same as current behaviour, but I think the pubkey to cert key matching check should only be performed if the certificate is not None, and it should be possible to have None for cert. My reasoning is that a user might not want to configure certs on each device for keylime but instead check keys against certs in the inventory server side, in which case we just want to send the keys and perform the certify, leaving out the certificates.

keylime/src/device_id.rs
```diff
@@ -0,0 +1,535 @@
...
+            .map_err(DeviceIDBuilderError::CertPubKeyMismatch)?;
+
+        let idevid = self.get_idevid(tpm_ctx)?;
+        let Some(idevid_cert) = self.idevid_cert.take() else {
+            unreachable!();
+        };
+
+        // Check that recreated/collected IDevID key matches the one in the certificate
+        tpm::check_pubkey_match_cert(&idevid.public, &idevid_cert, "IDevID")
```
Isaac-Matthews, line 429: See comment on IAK pubkey cert key check.

keylime/src/device_id.rs
```diff
@@ -0,0 +1,535 @@
...
+                }
+            }
+            None => Err(DeviceIDBuilderError::IDevIDHandleNotSet),
+        }
+    }
+
+    /// Get the IAK certificate
+    /// If the iak_cert is not set, try to load the certificate from the iak_cert_path and cache
+    /// the loaded certificate
+    fn get_iak_cert(&mut self) -> Result<&X509, DeviceIDBuilderError> {
```
Isaac-Matthews, line 352: I think this could be an option, with it only failing if a path is set but cert can't be found, and on empty string it is None. See pubkey cert key check comment for more detail.

keylime/src/device_id.rs
```diff
@@ -0,0 +1,535 @@
...
+                }
+                None => Err(DeviceIDBuilderError::IAKCertNotSet),
+            },
+        }
+    }
+
+    /// Get the IDevID certificate
+    ///
+    /// If the idevid_cert is not set, try to load the certificate from the idevid_cert_path and cache the loaded certificate
+    fn get_idevid_cert(&mut self) -> Result<&X509, DeviceIDBuilderError> {
```
Isaac-Matthews, line 382: Same as IAK cert getter.

keylime/src/device_id.rs
ansasaki: Since the IDevID key is not needed for the AK certification later, do you think we should flush it immediately after checking that the certificate matches its public key and drop it from the `DeviceID` structure?

Maybe the `DeviceID` structure should contain the IAK handle, IAK pubkey, IAK certificate, IDevID pubkey, and IDevID certificate instead of holding the `IAKResult` and `IDevIDResult`.

Something like:
```
#[derive(Debug)]
pub struct DeviceID {
    pub iak_handle: KeyHandle,
    pub iak_pub: TssPublic,
    pub iak_cert: X509,
    pub idevid_pub: TssPublic,
    pub idevid_cert: X509,
}
```
We could also flush the IAK handle after certifying AK with it as it is not necessary later.

Isaac-Matthews: We don't want to attempt to flush if the handle is from the config rather than from us regenerating the key. So I think we can flush the IDevID immediately after regenerating as long as we have collected the pub value if we are going to check against the cert.

Same for the IAK (we only want to flush after the certify if we regenerated it). I suppose you could check to see if there is a value in the handle value in config before deciding to flush?

That structure looks like all we need at the moment (but I think we maybe want the certs to be options, see other comments).

keylime/src/device_id.rs
ansasaki: To make the certificates optional, we need to decide on the behavior. I think the best is:

- If the configured certificate path is empty then ignore when the certificate is missing and continue without certificates
- If the configured certificate path is `default`, then log a warning when it is missing and continue without certificates
- If the configured certificate path was set, then error if the certificate is missing

Otherwise users may think their configuration is sane when actually the certificates were ignored.

Isaac-Matthews: I had been thinking that even with default it would error, and that when the path is empty it emits an info message to note that the keys are being sent without certs, but either way is fine I think (I would always include an info message to note the lack of certs though).

keylime/src/device_id.rs
ansasaki: Not the most elegant solution, but I added a `is_persistent` flag in the `IAKResult` and `IDevIDResult` to indicate if the handles should be flushed after use. Reading the configuration was not an option since the goal is to move the code to the `keylime` library, detached from the `keylime-agent` application.

keylime/src/device_id.rs
ansasaki: I made the changes to make the certificates optional, could you please review again?

sergio-correia (approved):
LGTM

PR #889, opened 2024-12-18
Support multiple API versions

ansasaki:
This introduces a new configuration option `api_versions` which allows the user to select a subset of the supported API versions to enable.

The agent creates the endpoints under each enable API version, allowing it to receive requests from older versions of the other components (verifier and tenant).

During registration, the agent will try to query the version of the API supported by the registrar making a GET request to the `/version` endpoint. If the registrar supports it and replies with a version, the agent will check if the version is enabled and will use it for the following requests.

In case the registrar does not support the `/version` endpoint, then the agent will try all the enabled versions. If the registration is successful, the agent will keep the successful API version to use in the following requests.

This is part of the implementation for the enhancement proposal 114: https://github.com/keylime/enhancements/pull/115


ansasaki:
The failing test needs an update proposed in: https://github.com/RedHat-SP-Security/keylime-tests/pull/688

ansasaki:
I just realized the registrar already supports the `/version` API in an undocumented API change.
We need to be more diligent updating the documentation. This is also a reminder for myself.

ansasaki:
@keylime/developer Please consider prioritizing reviewing this PR as it is blocking the new agent version release.

keylime-agent.conf
```diff
@@ -11,7 +11,17 @@
 # The configuration file version
 #
 # To override, set KEYLIME_AGENT_VERSION environment variable.
-version = "2.2"
+version = "2.4"
```
sarroutbi, line 14: What happened to 2.3 version? Is this 2.2 -> 2.4 jump expected?

sarroutbi (commented):
Minor comments. Change LGTM

keylime-agent.conf
sergio-correia: Look at the first commit in the series. 2.3 takes into account some idevid options.

ansasaki: as @sergio-correia commented, the first commit in the series is the bump to the `2.3` version to account for the IDevID/IAK related configuration options.

I recommend reviewing the commits one by one instead of the whole diff from master as it makes it easier to follow the changes.

sergio-correia (approved):
Thanks, LGTM

PR #935, opened 2025-02-15
Allow agent to start without root privileges

sgebauer:
Currently, `keylime_agent` must be started as root, and then drops permissions later.
This is not strictly necessary, as long as the agent has permission to talk to the TPM and read the TPM/IMA logs (e.g. through a group and/or ACLs).

This PR changes the UID check on startup to allow starting without root privileges.
If `run_as` is configured but we are already running unprivileged, then the `run_as` option will be ignored with a warning.

This allows, for example, running `keylime_agent` from a systemd service with `DynamicUsers=true`.

### DIFF
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -250,12 +250,10 @@ async fn main() -> Result<()> {
             Some(&config.agent.run_as)
         }
     } else {
-        error!("Cannot drop privileges: not enough permission");
-        return Err(Error::Configuration(
-            config::KeylimeConfigError::Generic(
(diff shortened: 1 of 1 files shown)


ansasaki:
Thank you for your contribution!

I was wondering if this could break deployments that rely on the current behavior, but I don't think it can: current working deployments are starting the process as root, and will continue to drop the privilege normally.

When the system is not prepared for this, meaning the process is not running under a user with access to the UEFI boot eventlog, IMA log, and `/var/lib/keylime`, I expect it to fail due to permission denied.

ansasaki (approved):

PR #945, opened 2025-02-24
Add capabilities_negotiation structures

sarroutbi:
This change aims to include those structures
that will be required to communicate
capabilities negotiation information for
Keylime Push model

Resolves: #933 

### DIFF
--- a/keylime/src/structures/capabilities_negotiation.rs
+++ b/keylime/src/structures/capabilities_negotiation.rs
@@ -1,0 +1,624 @@
+use serde::{Deserialize, Serialize};
+use serde_json::{from_value, to_value, Value as JsonValue};
+use std::convert::TryFrom;
+
+// Define the structure for the AttestationRequest:
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AttestationRequest {
+    #[serde(rename(serialize = "data", deserialize = "data"))]
+    pub data: RequestData,
+}
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RequestData {
+    #[serde(rename(serialize = "type", deserialize = "type"))]
+    pub type_: String,
+    pub attributes: Attributes,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Attributes {
+    pub evidence_supported: Vec<EvidenceSupported>,
+    pub boot_time: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "evidence_class", rename_all = "snake_case")]
+pub enum EvidenceSupported {
+    Certification {
+        evidence_type: String,
+        agent_capabilities: AgentCapabilities,
+    },
+    FullLog {
+        evidence_type: String,
+        version: String,
+    },
+    PartialLog {
+        evidence_type: String,
+    },
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct MbLog {
+    evidence_class: String,
+    version: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ImaEntries {
+    evidence_class: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AgentCapabilities {
+    spec_version: String,
+    hash_algorithms: Vec<String>,
+    signing_schemes: Vec<String>,
+    attestation_keys: Vec<AttestationKey>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct AttestationKey {
+    key_class: String,
+    key_identifier: String,
+    key_algorithm: String,
+    public_hash: String,
+}
+
+// Define the structure for the AttestationResponse:
+#[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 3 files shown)


keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -0,0 +1,353 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct EvidenceSupported {
+    pub evidence_class: String,
+    pub evidence_type: String,
+    pub agent_capabilities: AgentCapabilities,
+    pub version: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AgentCapabilities {
```
ansasaki, line 31: I think this should be renamed: these fields are about the TPM capabilities, and not the agent.

keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -0,0 +1,353 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ChosenParameters {
+    pub nonce: String,
+    pub pcr_selection: Vec<i32>,
+    pub hash_algorithm: String,
+    pub signing_scheme: String,
+    pub attestation_key: AttestationKey,
+    pub starting_offset: Option<i32>,
+}
```
ansasaki, line 81: This structure is mixing parameters from different types of evidence: `starting_offset` is related with the IMA log (in the example it is called a `partial_log`), while the rest of the fields are related with the TPM (and TPM quote).

Maybe it should be an union of the different types of parameters that are related with each type of evidence supported.

keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -0,0 +1,353 @@
...
+    pub boot_time: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct EvidenceSupported {
+    pub evidence_class: String,
+    pub evidence_type: String,
+    pub agent_capabilities: AgentCapabilities,
+    pub version: Option<String>,
+}
```
ansasaki, line 28: `EvidenceSupported` could probably be more generic, having just the type and version fields, and then maybe an `attributes` field that are different for each type of evidence: `tpm_quote`, `ima_log`, and `mb_log`. 

`AgentCapabilities` in this structure, as it is now, is related only to the `tpm_quote`, so probably it should not be part of all the evidence types. Maybe the idea was to have the `AgentCapabilities` be the `attributes` field I suggested above.

ansasaki (commented):
I think this is going to the right direction, it just need some adjustments.

I'll try to create the draft for my proposal of JSON schemas ASAP

keylime/src/structures/capabilities_negotiation.rs
sarroutbi: The problem with unions is how they are "selected" without messing the JSON ... you are completely right that some of the fields only apply to TPM, and starting_offset only applies to IMA but ... why not letting all them as an Option and let serde code them as required?

keylime/src/structures/capabilities_negotiation.rs
sarroutbi: I have no problem in changing name (indeed, it is updated in the PR) but ... are you proposing this change also to the OpenAPI / JSON? My point is ... should we change information detailed [here](https://gist.github.com/stringlytyped/ab76c6c83e4630be9446576fbeb7534e) ?

keylime/src/structures/capabilities_negotiation.rs
sarroutbi: My question is similar to the previous one ... why not leaving Agent/TPMCapabilities and version as optional instead of using unions?

keylime/src/structures/capabilities_negotiation.rs
ansasaki: We should discuss it with HPE guys, nothing is set in stone. Maybe we can keep it as it is now to unblock the rest of the implementation (sorry for making you change it).

keylime/src/structures/capabilities_negotiation.rs
ansasaki: To improve code readability and avoid confusion. Also we want to take advantage of using the structures as a form of validation.

If you disagree, let's discuss in the meeting

keylime/src/structures/capabilities_negotiation.rs
sarroutbi: Initially, I implemented the solution by using Option<>. I was not able to find a solution by using enums.

keylime/src/structures/capabilities_negotiation.rs
sarroutbi: Initially, I implemented the solution by using Option<>. I was not able to find a solution by using enums.

ansasaki:
I created JSON schemas that I believe (sorry, I didn't test) can validate the data format as it is in this PR:
https://gist.github.com/ansasaki/4c4b0159cfc4f8b85900cb86ff333ed9

ansasaki (approved):
LGTM! I think this is a good initial set of structs for the attestation parameters negotiation.

PR #952, opened 2025-03-04
keylime-agent.conf: only mention ecdsa and rsassa for signing

THS-on:
The other algorithms, while supported by the TPM, are not supported on the Keylime verifier side.


sarroutbi (approved):
Thanks for your change. LGTM

ansasaki:
LGTM! @THS-on could you please do a similar change updating the configuration file templates? I mean, modifying the line similar to the following for all template versions:

https://github.com/keylime/keylime/blob/bac42c41058a84efb8e52e43c28f82a5e2e2fbaa/templates/2.4/agent.j2#L173

ansasaki (approved):

PR #961, opened 2025-03-24
Update push model structures to latest values

sarroutbi:


keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -168,51 +185,158 @@ mod tests {
...
-                                        key_class: "private_key".to_string(),
-                                        key_identifier: "att_key_identifier".to_string(),
+                                signature_schemes: vec!["rsassa".to_string()],
+                                available_subjects: ShaValues {
+                                    sha1: vec![0x01, 0x02, 0x03],
+                                    sha256: vec![0x04, 0x05, 0x06],
+                                },
+                                certification_keys: vec![
+                                    CertificationKey {
+                                        key_class: "pair".to_string(),
```
ansasaki, line 198: In my opinion, we should use `asymmetric` here instead of `pair`. It is better aligned with crypto lingo.
And to me this information is redundant if we keep the `key_algorithm` field.

keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -168,51 +185,158 @@ mod tests {
...
+                                        local_identifier: "att_local_identifier".to_string(),
                                         key_algorithm: "rsa".to_string(),
-                                        public_hash: "cd293be6cea034bd45a0352775a219ef5dc7825ce55d1f7dae9762d80ce64411".to_string(),
+                                        key_size: 2048,
+                                        server_identifier: "ak".to_string(),
+                                        public: "cd293be6cea034bd45a0352775a219ef5dc7825ce55d1f7dae9762d80ce64411".to_string(),
```
ansasaki, line 203: Just a detail, but this example is definitely not base64 encoded. The content here should be the public key encoded as base64.

keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -49,19 +52,29 @@ pub struct ImaEntries {
...
+// Do not serialize the struct name, only the fields
+pub struct ShaValues {
+    sha1: Vec<u8>,
+    sha256: Vec<u8>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct AttestationKey {
+pub struct CertificationKey {
     key_class: String,
```
ansasaki, line 72: In my opinion, `key_class` is redundant with `key_algorithm` as the type of the key (symmetric of asymmetric) is implicit in the `key_algorithm`

keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -90,34 +108,35 @@ pub struct EvidenceRequested {
     pub evidence_type: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub chosen_parameters: Option<ChosenParameters>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub version: Option<String>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct TpmParameters {
```
ansasaki, line 114: To make it coherent with the  other structures, maybe this should be called `CertificationParameters`

keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -90,34 +108,35 @@ pub struct EvidenceRequested {
...
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub signing_scheme: Option<String>,
+    pub signature_scheme: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub attestation_key: Option<AttestationKey>,
+    pub certification_key: Option<CertificationKey>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct StartingOffset {
```
ansasaki, line 128: Shouldn't this be renamed to something like `LogParameters`?

ansasaki (commented):
I added some comments, probably we can discuss/clarify them quickly via slack. Otherwise we can bring this topic to the weekly meeting.

keylime/src/structures/capabilities_negotiation.rs
sarroutbi: It is OK for me to change it, but we can wait for next meeting to clarify it

ansasaki (approved):
LGTM! Let's merge as it is to unblock the next steps

keylime/src/structures/capabilities_negotiation.rs
stringlytyped: I am happy to have it as `asymmetric`. I agree that it is more in line with usual terminology

PR #964, opened 2025-03-26
Push model prototype

sarroutbi:
This prototype contains the basic structure
of the Push Model prototype. It allows
sending capabilities negotiation requests
to Verifier to check its current status
    
### DIFF
--- a/keylime-push-model-agent/Cargo.toml
+++ b/keylime-push-model-agent/Cargo.toml
@@ -1,0 +1,40 @@
+[package]
+name = "keylime_push_model_agent"
+description = "Rust agent for Keylime (Push Model)"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+version.workspace = true
+
+[dependencies]
+actix-web.workspace = true
+assert_cmd.workspace = true
+clap.workspace = true
+keylime.workspace = true
+predicates.workspace = true
+reqwest.workspace = true
+serde.workspace = true
+serde_derive.workspace = true
+serde_json.workspace = true
+static_assertions.workspace = true
+tokio.workspace = true
+
+[dev-dependencies]
+actix-rt.workspace = true
+
+[features]
+# The features enabled by default
+default = []
+testing = []
+legacy-python-actions = []
+
+[package.metadata.deb]
+section = "net"
+assets = [
+  ["target/release/keylime_push_model_agent", "usr/bin/", "755"],
+  ["../README.md", "usr/share/doc/keylime-agent/README", "644"],
+  ["../keylime-agent.conf", "/etc/keylime/agent.conf", "640"],
+]
+maintainer-scripts = "../debian/"
+systemd-units = { unit-scripts = "../dist/systemd/system/", enable = true }
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ tpmdata.json
 tests/.tmp*
 tests/actions/__pycache__
 tests/unzipped/__pycache__
+**/tarpaulin-report.*
+**/__pycache__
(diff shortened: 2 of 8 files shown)


sarroutbi:
Sorry for the "Copilot" review. I missclicked on it when attempting to ask for a review

keylime-push-model-agent/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+serde_derive.workspace = true
+serde_json.workspace = true
+static_assertions.workspace = true
+tempfile.workspace = true
+tokio.workspace = true
+tss-esapi.workspace = true
+thiserror.workspace = true
+uuid.workspace = true
+zeroize = "1.8.1"
+getrandom = "0.3.2"
```
ansasaki, line 37: Could you please try to cleanup the dependencies here and remove things that you will not use?
Also please be very defensive and add new dependencies only when strictly needed. If you really need a dependency, please check its availability in Fedora first.

sarroutbi: Sorry, I forgot to groom this. I kept those strictly necessary.

By checking in Fedora ... do you mean the component is available in Fedora? Are dependencies not handled through Cargo? Why are tests passing in the different Fedora versions if packages are not available?

sarroutbi: I checked in Fedora 41 and both `assert_cmd` and `predicates` exist:
```bash
$ dnf search rust assert_cmd 2>/dev/null 1>/dev/null && echo exists
exists
$ dnf search rust predicates 2>/dev/null 1>/dev/null && echo exists
exists
```

sarroutbi: Yes, both exists:
https://src.fedoraproject.org/rpms/rust-predicates
https://src.fedoraproject.org/rpms/rust-assert_cmd

If possible, I would like to add them, as both packages enable testing binaries easily:
https://github.com/keylime/rust-keylime/pull/964/files#diff-629dea229153170f9db2325748def31f811e87e9cc49c68307740a97ca0d57e3R18

keylime-push-model-agent/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+name = "keylime_push_model_agent"
+description = "Rust agent for Keylime (Push Model)"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+version.workspace = true
+
+[dependencies]
+actix-web.workspace = true
```
ansasaki, line 11: If I understand it correctly, the prototype only needs the `actix-web` crate for testing. We need to find a way to make it optional when compiling, only required when testing. It doesn't need to be done as part of this PR.

ansasaki (changes requested):
Sorry, but I'm requesting changes to drop the usage of the `rustls` crate. This is to prevent future work when working on FIPS-140 compliance.

keylime-push-model-agent/Cargo.toml
sarroutbi: Sorry, but I don't understand it very well ... why is it a problem taking into account it is being used in other parts of the project? I would understand if it was a brand new dependency just being included for this ... but in this case ... I don't find the reason (I might be wrong in any case)

ansasaki: The main reason I see is for clarity: setting it as optional only for testing shows that the final release binary do not depend on the crate.

A minor reason would be to speed-up the build in case someone wants to build only this individual package without tests, and not all the packages in the workspace (TBH, I don't know if that is actually achievable)

As I wrote before, this change is a nice to have, you don't need to change it for now.

sarroutbi:
Thanks for your quick response, @ansasaki . Sorry for the `rust-tls` issue, I was not aware of the FIPS-140 issue. May I ask you please to review PR again when possible?

ansasaki (changes requested):

ansasaki (approved):

PR #976, opened 2025-04-11
Move AuthTag and EncryptedData from the common.rs to the Keylime library

ansasaki:
Move the EncryptedData and AuthTag structures from common.rs into the keylime library.

This is to continue the effort to move the code that can be shared from the keylime-agent application to the keylime library

### DIFF
--- a/keylime-agent/src/common.rs
+++ b/keylime-agent/src/common.rs
@@ -33,11 +33,6 @@ use tss_esapi::{
     structures::PcrSlot, traits::UnMarshall, utils::TpmsContext,
 };
 
-/*
- * Constants and static variables
- */
-pub const AUTH_TAG_LEN: usize = 48;
-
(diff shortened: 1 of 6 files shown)


keylime-agent/src/common.rs
```diff
@@ -83,32 +78,6 @@ where
...
-            }
-            other => Err(format!(
-                "auth tag length {other} does not correspond to valid SHA-384 HMAC",
-            )),
-        }
-    }
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct EncryptedData {
```
sarroutbi, line 82: @ansasaki : Should not this EncryptedData be removed from here? Aren't we duplicating it?

ansasaki: Yes, you are right, we left it behind. Thanks for noticing!

PR #984, opened 2025-04-23
Add client certificates to push-attestation prototype

ansasaki:
This replaces the usage of `default-tls` feature from reqwest with `native-tls`, which uses OpenSSL as backend on Linux.

Add 'anyhow' crate as dependency to add context information to user facing errors.

This also disables the hostname verification because the certificates generated by the verifier don't have the Subject Alternative Name (SAN) set properly, making the certificates to fail hostname verification


Cargo.toml
```diff
@@ -35,7 +36,7 @@ picky-asn1-der = "0.4"
 picky-asn1-x509 = "0.12"
 predicates = { version = "3.1.3" }
 pretty_env_logger = "0.5"
-reqwest = {version = "0.12", default-features = false, features = ["json", "default-tls"]}
+reqwest = {version = "0.12", default-features = false, features = ["json", "native-tls"]}
```
sergio-correia, line 39: Good point. I was under the impression we had to use native-tls explictly instead of default-tls, since we have the goal to use openssl on Linux, as default-tls might change its backend in the future.

sergio-correia (approved):
Thanks, looks good to me.

sarroutbi (approved):
In general, changes LGTM.

The only issue I see is `get_https_client` function seems to be too long. I would define some intermediate functions such as:
- load_certificate
- load_client_identity
for the new code.
I left it to you, @ansasaki . You can push directly and we can refactor it once working.

Thanks!

PR #990, opened 2025-05-12
Integrate Evidence Handling/Authentication messages to push attestation prototype

sarroutbi:


ansasaki:
@sarroutbi Please use a more specific title for the PR. I suggest "Integrate Evidence Handling messages to push attestation prototype"

keylime-push-model-agent/src/main.rs
```diff
@@ -187,10 +207,10 @@ async fn send_attestation_request(
 
 #[derive(Parser, Debug)]
 #[command(author, version, about, long_about = None, ignore_errors = true)]
-struct Args {
+pub struct Args {
```
ansasaki, line 210: Is it necessary to make the `Args` structure public? If it is used only inside the crate, maybe we could limit it to be `pub(crate)`

ansasaki (approved):
LGTM!

PR #999, opened 2025-05-14
Refactor code: decrease main function size

sarroutbi:


ansasaki:
Sorry for being picky about the title of the PRs, but again I will suggest you to change it.
The thing is that in future when you are searching through history to understand when certain changes were introduced, the easiest is to find the PR where it landed.
I don't think this is a refactoring to reduce the size of `main.rs`. The reduction of the size of `main.rs` is only a consequence of your refactoring. To me the changes included here are about moving code to specific modules.

So, I suggest something like: "Move structure filling and URL selection related code to specific modules". Feel free to rephrase it if it is too long.

ansasaki (approved):
LGTM!

PR #1017, opened 2025-05-30
Move Agent identity information from the `RegistrarClient` structure to the new structure AgentIdentity

ansasaki:
Introduce a new `AgentIdentity` structure that contains the Agent identity information, removing the corresponding data from the `RegistrarClient` structure.

### DIFF
--- a/keylime/src/agent_identity.rs
+++ b/keylime/src/agent_identity.rs
@@ -1,0 +1,353 @@
+use log::*;
+use openssl::x509::X509;
+use thiserror::Error;
+
+use crate::crypto::{x509_to_der, x509_to_pem, CryptoError};
+
+#[derive(Error, Debug)]
+pub enum AgentIdentityBuilderError {
+    /// Agent public AK not set
+    #[error("Agent public AK not set")]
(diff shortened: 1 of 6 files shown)


keylime/src/agent_identity.rs
```diff
@@ -0,0 +1,353 @@
+use log::*;
+use openssl::x509::X509;
+use thiserror::Error;
+
+use crate::crypto::{x509_to_der, x509_to_pem, CryptoError};
+
+#[derive(Error, Debug)]
+pub enum AgentIdentityBuilderError {
+    /// Agent public AK not set
```
sarroutbi, line 9: Could you please remove these kind of comments? Error is already self-describing

sarroutbi (changes requested):
Minor suggestion for comments removal. Rest of changes LGTM

keylime/src/agent_identity.rs
sarroutbi: In case it is used for code documentation, I am OK with it

sarroutbi (approved):

PR #1018, opened 2025-06-04
Move reusable code from `keylime-agent` to the `keylime` library

ansasaki:
This moves the code that can be used by other applications, such as the push-attestation prototype, to the `keylime` library.


keylime/src/config/env.rs
```diff
@@ -0,0 +1,210 @@
+// SPDX-License-Identifier: Apache-2.0
```
sarroutbi, line 1: Is file header necessary? I have not included it in the latest changes but I doubt if it should be added or not. If so, I will create an additional PR for the ones that don't contain it

sarroutbi (approved):
Minor comments regarding file headers (we should homogeneize them, but it can be done in a different PR). Rest of changes LGTM

ansasaki:
@sarroutbi Hi, could you please take a look again? I added a specific error type  `ContextInfoError` to remove the `expect()` usage in the code. I overlooked it during my review of your code, but please avoid using `unwrap()` or `expect()` in the code as it will generate `panic` which kills the agent. It is acceptable to use them as part of the tests, but not as part of the actual code. I know we still have some of them spread around, but ideally we should remove all of them from the code.

sarroutbi (approved):
LGTM

PR #1028, opened 2025-06-16
Fix race conditions in tests

ansasaki:
Some tests could fail due to race conditions:

- In `keylime-agent/src/payloads.rs`, the `test_run_encrypted_payload()` and `test_payload_worker()` could affect each other because both set the environment variable `KEYLIME_TEST_DIR`
- In `keylime/src/cert.rs`, the `test_cert_no_server_cert()` and `test_server_from_server_key()` could delete each other's files because they used the same path
  - This was fixed by using a temporary directory for each test, which also avoids the need for cleanup before and after the test

### DIFF
--- a/keylime-agent/src/payloads.rs
+++ b/keylime-agent/src/payloads.rs
@@ -343,6 +343,13 @@ mod tests {
     };
     use tokio::sync::mpsc;
 
+    #[cfg(feature = "testing")]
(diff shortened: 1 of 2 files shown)


sarroutbi (approved):
Nice catch @ansasaki . LGTM

PR #1029, opened 2025-06-17
Fix code to send multiple AKs per EK

sarroutbi:


keylime-push-model-agent/src/main.rs
```diff
@@ -203,19 +208,46 @@ async fn send_push_model_request(args: &Args) -> Result<ResponseInformation> {
...
+            if args.avoid_tpm.unwrap_or(false) {
+                return Ok(None);
+            }
+            let config = keylime::config::PushModelConfig::default();
+            debug!("Initializing unique TPM Context...");
+            let context_info = context_info::ContextInfo::new_from_str(
+                context_info::AlgorithmConfigurationString {
+                    tpm_encryption_alg: config.get_tpm_encryption_alg(),
+                    tpm_hash_alg: config.get_tpm_hash_alg(),
+                    tpm_signing_alg: config.get_tpm_signing_alg(),
```
ansasaki, line 223: I was trying to understand why these are fixed in the context. I think you do not need to follow the pull agent so much, in the sense that there we had a single set of algorithms to be configured, while in the push model the agent will negotiate the parameters.

For me, it would make more sense if the context held the supported algorithms obtained from the TPM capabilities.

Then, after negotiating the algorithms with the verifier, you should use this to decide which keys to generate/load and send the requested AK.

ansasaki: OK, sorry for my last message, I think I was probably tired by the end of the day and could not understand the problem here. Now I see the problem.

I think the Agent should be configurable so that the user can decide which keys to offer, with the possibility of offering multiple keys if they want. So, in my mind, we should allow the user to configure a list of AK to offer, which can be:

- by setting the parameters to generate the key
- by setting the key handle, if it was persisted
- by setting a file path to load a previously generated key (maybe this will be part of the first option as usually we want to persist the generated keys somewhere)

We also should have a default option where it will offer a single key using a default set of algorithms. I think this is the easiest way for you to move forward (basically implement the default case, which is to offer a single AK key that is loaded from a default location or generated if the file is not there, like it is today in the pull agent).

I think we have to discuss with HPE guys and agree how this should work in the final form.

We could use JSON strings to specify the keys parameters or create a specific configuration language and write the parser for it.

sarroutbi: @ansasaki : Do you think it would be feasible to have a first version that sends one RSA and one ECC, then create all the possible ones from config once config is groomed? This way, we can have something with multiple AKs and deliver proposed changes (persistence and configuration based AKs creation) in other PRs

ansasaki: Yes, I think we can proceed like this and make the changes on a future PR

PR #1030, opened 2025-06-18
tests: Do not assume `/var/lib/keylime` exists

ansasaki:
The tests would fail when executed in an environment where the `/var/lib/keylime` path does not exist or is not accessible.
This makes the tests to not rely on the existence of  `/var/lib/keylime` and use a temporary directory instead.

For this a `get_testing_config()` method was added to the `config` crate, available only when the `testing` feature  is enabled. This method returns a `KeylimeConfig` structure, but with the `keylime_dir` option set to the given directory.

In tests, the calls to `KeylimeConfig::default()` were replaced by calls to `get_testing_config()`. Alternatively, when the `KeylimeConfig` structure was constructed manually, the `keylime_dir` option was set as an existing temporary directory.

The goal is to make the tests to pass even when executed in an environment where Keylime is not installed yet, for example during packit RPM building which is currently broken.

### DIFF
--- a/keylime-agent/src/keys_handler.rs
+++ b/keylime-agent/src/keys_handler.rs
@@ -607,7 +607,6 @@ mod tests {
         encrypt_aead, pkey_pub_from_pem, rsa_oaep_encrypt,
     };
     use crate::{
-        config::KeylimeConfig,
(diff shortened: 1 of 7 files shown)


ansasaki:
@sarroutbi @sergio-correia The reported coverage is very weird. I suspect something is not working as we expect. Please ignore it for now.

PR #1034, opened 2025-07-01
Add disallowed_signing_algorithms, avoid ecschnorr

sarroutbi:
This change introduces a new configuration option, disabled_signing_algorithms, to prevent the agent from using certain signing schemes, even if they are reported as supported by the TPM.

The primary motivation is to avoid issues with the `ecschnorr` algorithm, which, while technically supported by some TPMs, has been observed to cause rejection from Verifier.

### DIFF
--- a/keylime-push-model-agent/src/main.rs
+++ b/keylime-push-model-agent/src/main.rs
@@ -234,6 +234,9 @@ fn init_context<T: PushModelConfigTrait>(
                     tpm_hash_alg: config.tpm_hash_alg().to_string(),
                     tpm_signing_alg: config.tpm_signing_alg().to_string(),
                     agent_data_path: config.agent_data_path().to_string(),
+                    disabled_signing_algorithms: config
+                        .disabled_signing_algorithms()
+                        .clone(),
                 },
             )?;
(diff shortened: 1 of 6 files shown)


sarroutbi:
@ansasaki : Please, forget about coverity data. If I run it locally with `tarpaulin`, I am getting 100% coverage on push_model, for example:

image

sarroutbi:
> In my opinion, the wording should be changed from `prohibited` to `disallowed` or `disabled`. `Prohibited` sounds like there is something inherently wrong with the algorithm, like if it was broken, when actually we just want to not allow its usage.
> 
Sure. I will change it.

> Another option is to introduce something like `enabled_*_algorithms` to list all the enabled/allowed algorithms instead of having a list of rejected algorithms. In general having an explicit list of enabled algorithms is better than having to construct a list by taking the supported algorithms and removing the rejected algorithms. It makes it clear what is enabled in a single place, helping also auditing the code in future.
At this particular moment, I prefer to keep the "disabled" option, but we can open an issue to track this in the future if you think this is necessary

ansasaki (approved):

PR #1037, opened 2025-07-01
config: Unify configuration for pull and push agents

ansasaki:
Refactor the configuration code, removing the wrapper structure `KeylimeConfig` and keeping only the internal configuration structure `AgentConfig`.

The `AgentConfig` structure holds the configuration options for both pull and push agents.

A new crate `keylime-macros` is introduced, with the implementation of the `define_trait_view` procedural macro.

The define_config_view procedural macro allows annotating a structure
with the attribute `#[define_view_trait(for_struct = "SomeStructure")]`.

The annotated structure serves as a template to create a Trait which
provides a view of the referred structure, restricting the accessible
fields through getter methods. The created Trait is named as the
annotated name followed by "Trait".

For types that are `Copy` primitives (e.g. `u16`, `u32`, etc.), the getter
method will return a copy of the value. For other types, the getter
method will return a reference to the value, unless it is annotated with
the `#[copy]` attribute. Only types that implement the `Copy` trait can be
annotated with the `#[copy]` attribute.

The annoted structure fields can also be annotated with the attribute
`#[transform(using = custom_transform_fn, error = CustomError)]`.
The annotated fields getters are generated using the provided transform
function which may return an error of the given type.

```rust
   struct KeylimeConfig {
        node_id: String,
        port: u16,
        api_versions: String,
        info: CustomCopyType,
    }
    
    \#[define_view_trait(for_struct = KeylimeConfig)]
    struct KeylimeView {
        node_id: String,
        port: u16,
        #[transform(using = parse_string_list, error = ParsingError)]
        api_versions: Vec<String>,
        #[copy]
        info: CustomCopyType,
    }
 ```
    
 Expands into:
    
 ```rust
    pub trait KeylimeViewTrait {
        fn node_id(&self) -> &str;
        fn port(&self) -> u16;
        fn api_versions(&self) -> Result<Vec<String>, ParsingError>;
        fn info(&self) -> CustomCopyType;
    }
    
    impl KeylimeViewTrait for KeylimeConfig {
        fn node_id(&self) -> &str { &self.node_id }
        fn port(&self) -> u16 { self.port }
        fn api_versions(&self) -> Result<Vec<String>, ParsingError> {
            parse_string_list(&self.api_versions)
        }
        fn info(&self) -> CustomCopyType { self.info }
   }
```

The `PushModelConfigTrait` is implemented using the `define_view_trait` macro, replacing the previously existing implementation. The `PushModelConfig` is also removed in favor of using the unified `AgentConfig`, whith the view restricted by the imported trait.

Finally, the `AgentConfig` is used in the push model client prototype, adding support to the whole configuration via the `agent.conf` file, overrides using configuration snippets, and overrides with environment variables, in the same way it is currently supported by the pull model agent.


sergio-correia (approved):
Nice job, LGTM!

packit-ci.fmf
```diff
@@ -22,8 +22,8 @@
 
   discover:
     how: fmf
-    url: https://github.com/RedHat-SP-Security/keylime-tests
-    ref: main
+    url: https://github.com/ansasaki/keylime-tests
+    ref: fix_secure_mount
```
sarroutbi, line 26: Is this really required? If necessary ... why not updating the tests in RedHat-SP-Security? Normally, it is not a good practice to use personal forks ...

ansasaki: Yes, this is necessary. The changes were proposed in: https://github.com/RedHat-SP-Security/keylime-tests/pull/798

We usually to this to be able to test before the PR is merged on `keylime-tests` side. This is the way we've been testing while we don't have the test merged there, and also to confirm that the fixes to the tests are correct.

ansasaki: The test changes were merged in the `keylime-tests` side. I dropped the commit changing the location of the tests.

PR #1038, opened 2025-07-02
Push Model: Perform attestation (quote, uefi and ima logs)

sarroutbi:
This PR enhances the push-model agent by adding the complete logic for IMA and UEFI log collection and submission, making the attestation process asynchronous, and refactoring the agent's main loop for clarity and robustness.

Key Changes:

* Full IMA and UEFI Log Collection:
   The agent now properly handles requests for IMA and UEFI event logs.

* Asynchronous Attestation:
    The entire evidence creation process is now asynchronous to prevent blocking.

* Code Refactoring and Cleanup:
   Added the core perform_attestation function to ContextInfo, centralizing the logic for creating attestation evidence. 
   The main agent binary's logic has been significantly refactored. The main.rs file now orchestrates the high-level flow, while the specifics of sending HTTP requests and handling responses have been moved into a dedicated attestation.rs module.

PR #1039, opened 2025-07-03
Send Content-Type:application/json on request

sarroutbi:


sarroutbi:
@ansasaki , @sergio-correia : right now the best way to send expected headers is to code them manually. I will work to use .json() on reqwest on the final attestation PR #1038

sergio-correia (approved):

PR #1042, opened 2025-07-09
Fix clippy warnings about uninlined format arguments

ansasaki:
Use the arguments inline when constructing strings using `format!()` and for logging.


sarroutbi (approved):

sarroutbi:
LGTM

PR #1044, opened 2025-07-10
context_info_handler: Do not assume /var/lib/keylime exists

ansasaki:
This changes the tests in the context_info_handler.rs to not assume the `/var/lib/keylime` exists.

In tests, whenever a configuration is necessary, the `config::get_testing_config()` should be called, passing the path to a temporary directory as argument so that it overrides the `keylime_dir` option to not use the default `/var/lib/keylime`.

Note that this is only available when the `testing` feature is enabled.

This is necessary to avoid the tests to fail when the `/var/lib/keylime` directory is not present, for example in build systems or when the tests are executed before Keylime is installed.

### DIFF
--- a/keylime-push-model-agent/src/context_info_handler.rs
+++ b/keylime-push-model-agent/src/context_info_handler.rs
@@ -85,6 +85,7 @@ mod tests {
     #[tokio::test]
     async fn test_context_with_avoid_tpm_flag() {
         // Lock mutex to avoid race condition accessing the TPM
+        // Use temporary directory instead of assuming /var/lib/keylime exists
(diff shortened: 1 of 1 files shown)


keylime-push-model-agent/src/context_info_handler.rs
```diff
@@ -101,16 +102,15 @@ mod tests {
     }
 
     #[tokio::test]
-    #[cfg(feature = "testing")]
```
sarroutbi, line 104: Why did you remove this "testing" feature label? If I am not wrong, all the TPM related tests are executed with it

sarroutbi (approved):
LGTM

PR #1049, opened 2025-07-14
keylime/structures: Rename ShaValues to PcrBanks

ansasaki:
Add all the supported hash algorithms, including the missing sha384 and sha512.

Implement the builder pattern to create the structure more ergonomically, which will not set empty banks to the structure.

Skip the serialization of unset banks, avoiding empty banks to be sent to the Verifier.


sarroutbi (approved):
Apart from provided comment, change LGTM

PR #1050, opened 2025-07-14
push-model: refactor attestation logic into a state machine

sergio-correia:
This commit refactors the main attestation logic into a state machine, simplifying the main function and making the agent's state 

The state machine handles the agent's lifecycle, from registration to attestation.


ansasaki (approved):
LGTM!

PR #1053, opened 2025-07-16
tpm: Add method to extract signing scheme and hash algorithm from AK

ansasaki:
The signing scheme and hash algorithm that the AK can use is set and locked during its generation.  This means that an AK cannot be used to generate signatures using a different scheme.

This adds a method to extract the signing scheme and hash algorithm set during the AK generation from the TPM for the given AK.

The push attestation agent prototype is modified to generate a quote using the extracted signing scheme and hash algorithm.


sarroutbi (approved):
LGTM

PR #1066, opened 2025-07-31
push-model: implement continuous attestation with configurable intervals

sergio-correia:
After the first successful attestation, the agent now waits for a
configurable interval before repeating the attestation process by
returning to the Negotiating state, creating a continuous attestation
loop.

The interval between the attestations is currently fixed, but in the
future, the verifier will provide this information in its response to
the attestation, so we can parse it from there and use it instead.

Currently, the interval between sending the measurements is defined
as 60s, but can be configured with the --attestation-interval-seconds
switch.

### DIFF
--- a/keylime-push-model-agent/src/attestation.rs
+++ b/keylime-push-model-agent/src/attestation.rs
@@ -14,6 +14,16 @@ pub struct ResponseInformation {
     pub body: String,
 }
 
+impl Default for ResponseInformation {
+    fn default() -> Self {
+        Self {
+            status_code: StatusCode::OK,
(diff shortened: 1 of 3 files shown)


sarroutbi (commented):
Please, define const values to avoid "magic numbers":

measurement_interval_seconds: 60, -> measurement_interval_seconds: DEFAULT_MEASUREMENT_INTERVAL_SECONDS

ansasaki (approved):
LGTM! I tested the continuous attestation against current state of the Verifier from the https://github.com/keylime/keylime/pull/1693 and it worked well in the happy execution path.

PR #1071, opened 2025-08-11
Remove unnecessary configuration fields

sarroutbi:


sergio-correia (approved):

keylime-push-model-agent/src/struct_filler.rs
```diff
@@ -164,21 +164,21 @@ impl<'a> FillerFromHardware<'a> {
...
                         },
                         structures::EvidenceSupported::EvidenceLog {
                             evidence_type: "ima_log".to_string(),
                             capabilities: structures::LogCapabilities {
                                 evidence_version: None,
                                 entry_count: ima_log_count,
-                                supports_partial_access: config.ima_logs_supports_partial_access(),
-                                appendable: config.ima_logs_appendable(),
+                                supports_partial_access: false,
+                                appendable: false,
```
ansasaki, line 179: For the IMA logs, I expected these two to be true: we do support providing the IMA logs from a given index (which I think was the purpose for the `supports_partial_access`) and also we expect new entries to be added dynamically (which I think was the purpose of the `appendable` option).

Did I misunderstood the options? Or is this setting temporary until we integrate these to the push-attestation agent?

ansasaki (approved):

PR #1073, opened 2025-08-12
Fix IMA log format (it must be 'text/plain')

sarroutbi:
This patch corrects the MIME type format for the IMA log reported by the agent.

The value for formats was changed from the incorrect "application/text" to the standard and correct "text/plain". This ensures the agent's capabilities report aligns with IANA standards and what the verifier expects for a plain text IMA log.


sarroutbi:
> Nice catch, sorry for the oversight during my previous review.

No problem at all. It was my fault as I did not consider previous configuration values

PR #1078, opened 2025-08-14
Fix boot time retrieval

sarroutbi:
This patch correctly implements boot time retrieval for the Rust agent, replacing the temporary use of the current time with the actual system boot time.

* New boot_time.rs Module: A new module was created to contain the logic for getting the system boot time. It works by parsing the btime field from the /proc/stat file, which is the standard method on Linux.

* Updated Agent Payload: The agent no longer sends the current time (chrono::Utc::now()) as the boot_time. Instead, it now calls the new get_boot_time() function to provide the true boot time in its capabilities report.

* Error Handling: If reading the boot time fails, the agent now logs an error and defaults to the Unix epoch (1970-01-01), ensuring the program doesn't crash.

### DIFF
--- a/keylime-push-model-agent/src/struct_filler.rs
+++ b/keylime-push-model-agent/src/struct_filler.rs
@@ -7,7 +7,7 @@ use keylime::context_info::ContextInfo;
 use keylime::ima::ImaLog;
 use keylime::structures;
 use keylime::uefi::uefi_log_handler;
-use log::error;
+use log::{error, warn};
 
 #[async_trait]
 pub trait StructureFiller {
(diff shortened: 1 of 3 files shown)


keylime-push-model-agent/src/struct_filler.rs
```diff
@@ -48,6 +48,17 @@ impl StructureFiller for FillerFromHardware<'_> {
     }
 }
 
+fn boot_time() -> chrono::DateTime<chrono::Utc> {
+    match keylime::boot_time::get_boot_time() {
+        Ok(time) => time,
+        Err(e) => {
+            error!("Failed to get boot time: {e}");
```
ansasaki, line 55: Could we make this message a warning? And also include the information that we are falling back to EPOCH?

PR #1082, opened 2025-08-19
config: Use the singleton pattern to avoid reading the files multiple times

ansasaki:
This makes use of the singleton pattern to avoid re-reading the configuration files multiple times. Also, by providing a global way to obtain a reference to the configuration options (via the `get_config()` method), there is no need to pass the configuration structure through multiple layers, simplifying the API.

The configuration is basically immutable after loaded and processed, which means that it is basically read-only throughout the execution of the agent. Since no changes are expected to the configuration options after it is first loaded, there is no need to protect it against concurrent accesses. Using only references to the configuration options instead of creating copies via `clone()` improves the performance and reduces the memory usage.

For testing, it is necessary to have a mechanism that allows overriding the configuration options. This mechanism is more complex when the tests are executed in parallel under the same process, since the configuration is loaded only once for all tests. To override the options, a special caching mechanism was created to allow the configuration override to be applied throughout the test case execution. The tests using the configuration override **cannot** be executed in parallel because they could affect each other, being the recommended way to use the mutex mechanism used to isolate test cases that use the TPM (from `keylime::tpm::testing`). In addition, it is necessary to cleanup the override configuration if it is set.

### DIFF
--- a/keylime-agent/src/keys_handler.rs
+++ b/keylime-agent/src/keys_handler.rs
@@ -881,7 +881,7 @@ mod tests {
     async fn test_u_or_v_key(key_len: usize, payload: Option<&[u8]>) {
         // Create temporary working directory and secure mount
         let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
-        let test_config = get_testing_config(temp_workdir.path());
+        let test_config = get_testing_config(temp_workdir.path(), None);
 
         let (mut fixture, mutex) = QuoteData::fixture().await.unwrap(); //#[allow_ci]
 
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -956,7 +956,7 @@ mod testing {
             let work_dir =
                 Path::new(env!("CARGO_MANIFEST_DIR")).join("tests");
 
-            let test_config = get_testing_config(&work_dir);
+            let test_config = get_testing_config(&work_dir, None);
             let mut ctx = tpm::Context::new()?;
 
             let tpm_encryption_alg =
--- a/keylime-agent/src/payloads.rs
+++ b/keylime-agent/src/payloads.rs
@@ -415,7 +415,7 @@ echo hello > test-output
     #[test]
     fn test_setup_unzipped() {
         let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
-        let test_config = get_testing_config(temp_workdir.path());
+        let test_config = get_testing_config(temp_workdir.path(), None);
         let secure_mount =
             PathBuf::from(&temp_workdir.path().join("tmpfs-dev"));
         fs::create_dir(&secure_mount).unwrap(); //#[allow_ci]
@@ -448,7 +448,7 @@ echo hello > test-output
     #[test]
     fn test_unzip_payload() {
         let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
-        let test_config = get_testing_config(temp_workdir.path());
+        let test_config = get_testing_config(temp_workdir.path(), None);
         let payload_path = Path::new(env!("CARGO_MANIFEST_DIR"))
             .join("test-data")
             .join("payload.zip");
@@ -481,7 +481,7 @@ echo hello > test-output
             .lock()
             .await;
         let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
-        let test_config = get_testing_config(temp_workdir.path());
+        let test_config = get_testing_config(temp_workdir.path(), None);
         let secure_mount =
             PathBuf::from(&temp_workdir.path().join("tmpfs-dev"));
         fs::create_dir(&secure_mount).unwrap(); //#[allow_ci]
@@ -531,7 +531,7 @@ echo hello > test-output
         use crate::{config::DEFAULT_PAYLOAD_SCRIPT, secure_mount};
 
         let temp_workdir = tempfile::tempdir().unwrap(); //#[allow_ci]
-        let test_config = get_testing_config(temp_workdir.path());
+        let test_config = get_testing_config(temp_workdir.path(), None);
         let secure_mount =
             PathBuf::from(&temp_workdir.path().join("tmpfs-dev"));
         fs::create_dir(&secure_mount).unwrap(); //#[allow_ci]
--- a/keylime-agent/src/revocation.rs
+++ b/keylime-agent/src/revocation.rs
@@ -525,7 +525,7 @@ mod tests {
     fn revocation_scripts_ok() {
         let work_dir = tempfile::tempdir()
             .expect("failed to create temporary directory");
-        let test_config = get_testing_config(work_dir.path());
+        let test_config = get_testing_config(work_dir.path(), None);
         let json_file = concat!(
             env!("CARGO_MANIFEST_DIR"),
             "/tests/unzipped/test_ok.json"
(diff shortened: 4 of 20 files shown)


sarroutbi (changes requested):
It seems there is an issue with test execution that could lead to race conditions (I have to recognize this was detected by Gemini, as I would have not noticed it so). Let me explain it:

* Test-Safe Configuration Overrides:

The current implementation for overriding configurations in tests uses a global `static Mutex`. While functional for serial execution, this pattern introduces significant risks of flaky tests and state leakage when tests are run in parallel.

* The Problem: Test Parallelism and State Leakage:

Parallel Test Interference: cargo test runs tests in parallel by default. With a single global override, tests will race to set and clear the configuration. One test can interfere with another, leading to unpredictable failures that are difficult to debug.

  - State Leakage on Panics: If a test sets an override and then panics, the clear_testing_config_override() function is never called. The leaked configuration will then affect all subsequent tests, breaking test independence.

* The Solution: Thread-Local Storage and RAII:

A more robust and idiomatic Rust pattern for managing test-specific state is to use a combination of a `thread-local static` and an RAII guard.

    - thread_local!: This creates a static variable that is unique to each test thread, completely eliminating race conditions between parallel tests.

    - RAII Guard: This is a struct whose Drop implementation automatically cleans up the override when it goes out of scope. This guarantees that the configuration is reset, even if the test panics.

* Suggested implementation:
Here is how you could refactor keylime/src/config/testing.rs to use this improved pattern.

1. Define the thread_local! static and the RAII Guard

```
// In keylime/src/config/testing.rs

use std::cell::RefCell;

// Use a thread_local static. No Mutex is needed.
thread_local! {
    static TESTING_CONFIG_OVERRIDE: RefCell<Option<AgentConfig>> = RefCell::new(None);
}

/// A guard that clears the configuration override when it goes out of scope.
#[must_use = "The guard must be held for the override to apply"]
pub struct ConfigOverrideGuard;

impl Drop for ConfigOverrideGuard {
    fn drop(&mut self) {
        // Automatically clear the override when the guard is dropped.
        TESTING_CONFIG_OVERRIDE.with(|cell| {
            *cell.borrow_mut() = None;
        });
    }
}

/// Override the global configuration for the duration of a test.
///
/// This function sets a thread-local override and returns a guard object.
/// When the guard is dropped at the end of the scope, the override is automatically cleared.
pub fn override_config_for_test(config: AgentConfig) -> ConfigOverrideGuard {
    TESTING_CONFIG_OVERRIDE.with(|cell| {
        *cell.borrow_mut() = Some(config);
    });
    ConfigOverrideGuard
}

// This function is still needed for get_config() to read the value.
pub fn get_testing_config_override() -> Option<AgentConfig> {
    TESTING_CONFIG_OVERRIDE.with(|cell| cell.borrow().clone())
}

// The old `set_...` and `clear_...` functions can now be removed.
```

* How the Code Improves:

  - Before: Manual and unsafe for parallel execution.
```
#[test]
fn my_test() {
    let test_config = get_testing_config(...);
    set_testing_config_override(test_config);

    // ... test logic ...

    clear_testing_config_override(); // <-- Easy to forget and not panic-safe
}
```
  - After: Ergonomic, panic-safe, and parallel-safe.

```
#[test]
fn my_test() {
    let test_config = get_testing_config(...);

    // Override is active while `_guard` is in scope.
    // Cleanup is automatic and guaranteed.
    let _guard = override_config_for_test(test_config);

    // ... test logic ...
}
```

ansasaki:
> It seems there is an issue with test execution that could lead to race conditions (I have to recognize this was detected by Gemini, as I would have not noticed it so). Let me explain it:
> 
>     * Test-Safe Configuration Overrides:
> 
> 
> The current implementation for overriding configurations in tests uses a global `static Mutex`. While functional for serial execution, this pattern introduces significant risks of flaky tests and state leakage when tests are run in parallel.
> 
>     * The Problem: Test Parallelism and State Leakage:
> 
> 
> Parallel Test Interference: cargo test runs tests in parallel by default. With a single global override, tests will race to set and clear the configuration. One test can interfere with another, leading to unpredictable failures that are difficult to debug.
> 
>     * State Leakage on Panics:
> 
> 
> If a test sets an override and then panics, the clear_testing_config_override() function is never called. The leaked configuration will then affect all subsequent tests, breaking test independence.
> 
>     * The Solution: Thread-Local Storage and RAII:
> 
> 
> A more robust and idiomatic Rust pattern for managing test-specific state is to use a combination of a `thread-local static` and an RAII guard.
> 
>     1. thread_local!: This creates a static variable that is unique to each test thread, completely eliminating race conditions between parallel tests.
> 
>     2. RAII Guard: This is a struct whose Drop implementation automatically cleans up the override when it goes out of scope. This guarantees that the configuration is reset, even if the test panics.
> 
> 
>     * Suggested implementation:
>       Here is how you could refactor keylime/src/config/testing.rs to use this improved pattern. Define the thread_local! static and the RAII Guard:
> 
> 
> ```
> // In keylime/src/config/testing.rs
> 
> use std::cell::RefCell;
> 
> // Use a thread_local static. No Mutex is needed.
> thread_local! {
>     static TESTING_CONFIG_OVERRIDE: RefCell<Option<AgentConfig>> = RefCell::new(None);
> }
> 
> /// A guard that clears the configuration override when it goes out of scope.
> #[must_use = "The guard must be held for the override to apply"]
> pub struct ConfigOverrideGuard;
> 
> impl Drop for ConfigOverrideGuard {
>     fn drop(&mut self) {
>         // Automatically clear the override when the guard is dropped.
>         TESTING_CONFIG_OVERRIDE.with(|cell| {
>             *cell.borrow_mut() = None;
>         });
>     }
> }
> 
> /// Override the global configuration for the duration of a test.
> ///
> /// This function sets a thread-local override and returns a guard object.
> /// When the guard is dropped at the end of the scope, the override is automatically cleared.
> pub fn override_config_for_test(config: AgentConfig) -> ConfigOverrideGuard {
>     TESTING_CONFIG_OVERRIDE.with(|cell| {
>         *cell.borrow_mut() = Some(config);
>     });
>     ConfigOverrideGuard
> }
> 
> // This function is still needed for get_config() to read the value.
> pub fn get_testing_config_override() -> Option<AgentConfig> {
>     TESTING_CONFIG_OVERRIDE.with(|cell| cell.borrow().clone())
> }
> 
> // The old `set_...` and `clear_...` functions can now be removed.
> ```
> 
>     * How the Code Improves:
> 
> 
> Before: Manual and unsafe for parallel execution.
> 
> ```
> #[test]
> fn my_test() {
>     let test_config = get_testing_config(...);
>     set_testing_config_override(test_config);
> 
>     // ... test logic ...
> 
>     clear_testing_config_override(); // <-- Easy to forget and not panic-safe
> }
> ```
> 
> After: Ergonomic, panic-safe, and parallel-safe.
> 
> ```
> #[test]
> fn my_test() {
>     let test_config = get_testing_config(...);
> 
>     // Override is active while `_guard` is in scope.
>     // Cleanup is automatic and guaranteed.
>     let _guard = override_config_for_test(test_config);
> 
>     // ... test logic ...
> }
> ```

I can try again to use this approach (I tried before reaching this final implementation), but it won't allow any parallel execution of most tests anyway because the tests need to access the TPM which is a single shared resource. Even though the configuration could be made thread-local, the TPM cannot. That is why we have the mutex defined in `tpm::testing::lock_tests()` to force the tests to run sequentially.

AFAIK, I used the mutex to lock the execution of tests that access the TPM **and** the tests that access the configuration.

About the singleton implementation, it is intentionally not using any access control for better performance: since the configuration is read-only, there is no risk of accessing only references for reading in parallel.

sarroutbi:
> it is intentionally not using any access control for better performance: since the configuration is read-only, there is no risk of accessing only references for reading in parallel.

OK. Thanks for the explanation.

sarroutbi (approved):
LGTM

sergio-correia (approved):

ansasaki:
@sarroutbi I see that one thing confusing is that the `create_test_state_machine()` calls the `set_testing_config_override()` internally, which requires the test to call `clear_testing_config_override()` at the end with no explicit reason. I'll try to make an improvement so that it is no longer necessary to call it by the end of the test, which is less error prone, and less confusing.

PR #1093, opened 2025-08-28
Add state to avoid exiting on wrong attestation

sarroutbi:
This patch refactors the agent's state machine to make it more resilient
by preventing it from exiting on recoverable attestation or negotiation
errors. Previously, any failure during these stages would transition the
agent to a terminal Failed state.

This change ensures that the agent can recover from transient issues
(like network hiccups or temporary verifier unavailability) without
needing to be restarted. This is the list of the key changes:

* New AttestationFailed State: A new, non-terminal state
AttestationFailed(error, context) has been introduced to specifically
handle recoverable errors that occur during communication with the
verifier.

* Retry Logic: The main run loop now handles the AttestationFailed
state. When this state is reached, the agent will:

* Log the error: Wait for the default attestation interval.

* Transition: go back to the Negotiating state to retry the process.

* State Transitions: The handle_negotiation and attest methods have
been updated. Instead of transitioning to the terminal Failed state on
network errors or non-successful HTTP status codes, they now transition
to the new AttestationFailed state, enabling the retry loop.

### DIFF
--- a/keylime-push-model-agent/src/state_machine.rs
+++ b/keylime-push-model-agent/src/state_machine.rs
@@ -22,6 +22,7 @@ pub enum State {
     Negotiating(ContextInfo),
     Attesting(ContextInfo, ResponseInformation),
     RegistrationFailed(anyhow::Error),
+    AttestationFailed(anyhow::Error, ContextInfo),
     Failed(anyhow::Error),
 }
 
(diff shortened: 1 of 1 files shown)


ansasaki (approved):
LGTM!

PR #1095, opened 2025-08-29
Add Mermaid FSM chart

sarroutbi:
Include Mermaid FSM chart containing a description of both the states and transitions existing on the FSM


sergio-correia:
I wonder if github would render it if the file was `.md` instead of `.txt`?

sergio-correia (approved):
Thanks, it renders the file now.

PR #1097, opened 2025-09-02
Add policy to handle Retry-After response headers

sarroutbi:
This pull request enhances the ResilientClient by adding support for the standard Retry-After HTTP response header.

Currently, our client relies solely on an exponential backoff strategy for all retryable errors. While effective, this approach is less precise when a server explicitly indicates how long we should wait before trying again (e.g., for rate limiting or temporary maintenance).

This change introduces a dedicated middleware that respects the Retry-After header, making our client a better HTTP citizen and improving its ability to recover from transient server-side issues efficiently.

Co-Authored-By: Gemini <noreply@google.com>
Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1409,6 +1409,7 @@ dependencies = [
  "glob",
  "hex",
  "http 1.1.0",
+ "httpdate",
  "keylime-macros",
  "libc",
  "log",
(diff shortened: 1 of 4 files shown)


ansasaki (approved):
LGTM!

PR #1103, opened 2025-09-03
Remove compilation warnings

sarroutbi:


sarroutbi:
> I am working on that auth code now, as I am continuing the implementation of the challenge-response protocol, which will even involve the moving of some things here to the core library, to better integrate with the resilient client. To reduce the conflicts, please delay this kind of change until that integration is done.

OK. I will do that. But please, from now on, try minimizing committing changes that contain warnings, as they keep compilation output dirty, even  more when they are as easy to fix as adding `#[allow(dead_code)]`

PR #1105, opened 2025-09-03
FSM: Fix compilation warnings detected by clippy

sarroutbi:
This PR resolves a clippy::needless_lifetimes warning found in the create_test_config test helper function.

The explicit 'a lifetime annotation was unnecessary, as Rust's lifetime elision rules can correctly infer the lifetime of the input &str and the returned NegotiationConfig.

The function signature has been updated to remove the explicit lifetime, making the code more idiomatic.


sergio-correia:
Looks good, but you could probably indicate in the title that this is about the state machine, so it's clearer at a first sight.

PR #1108, opened 2025-09-03
Integrate authentication middleware into ResilientClient 

sergio-correia:
Here we integrated authentication middleware into ResilientClient with automatic token
management and refresh, establishing  an authentication system that transparently handles
TPM-based challenge-response authentication for HTTP clients.


sarroutbi (approved):
Thanks for the PR. Changes LGTM

PR #1110, opened 2025-09-03
rpm: Do not remove wiremock which is now available in Fedora

ansasaki:
This also fixes the latest introduced build errors during the RPM package build on Copr.


sarroutbi (approved):
LGTM

PR #1111, opened 2025-09-03
workflows: Add test to check if RPM patches still apply

ansasaki:
The tests simply try to patch the current code with the patches for both Fedora and CentOS. This is a sanity check to prevent changes changes that would make the patches incompatible to be introduced without properly updating the patches.

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -1,0 +1,25 @@
+---
+name: rpm-test
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+  workflow_dispatch:
+
+jobs:
+  fedora:
+    name: Check that Fedora RPM patches still apply
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - name: Check that Fedora patches apply
+        run: for p in rpm/fedora/*.patch; do patch -p1 < "$p"; done
+  centos:
+    runs-on: ubuntu-latest
+    name: Check that CentOS RPM patches still apply
+    steps:
+      - uses: actions/checkout@v5
+      - name: Check that CentOS patches apply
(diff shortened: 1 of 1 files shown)


sergio-correia (approved):

sarroutbi (commented):
yamllint reports some issues:
```bash
  1:1       warning  missing document start "---"  (document-start)
  3:1       warning  truthy value should be one of [false, true]  (truthy)
  5:16      error    too many spaces inside brackets  (brackets)
  5:23      error    too many spaces inside brackets  (brackets)
  7:16      error    too many spaces inside brackets  (brackets)
  7:23      error    too many spaces inside brackets  (brackets)
  15:5      error    wrong indentation: expected 6 but found 4  (indentation)
  22:5      error    wrong indentation: expected 6 but found 4  (indentation)
```
Please, fix them if possible

sarroutbi:
> > yamllint reports some issues:
> > ```shell
> >   1:1       warning  missing document start "---"  (document-start)
> >   3:1       warning  truthy value should be one of [false, true]  (truthy)
> >   5:16      error    too many spaces inside brackets  (brackets)
> >   5:23      error    too many spaces inside brackets  (brackets)
> >   7:16      error    too many spaces inside brackets  (brackets)
> >   7:23      error    too many spaces inside brackets  (brackets)
> >   15:5      error    wrong indentation: expected 6 but found 4  (indentation)
> >   22:5      error    wrong indentation: expected 6 but found 4  (indentation)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please, fix them if possible. It you are in a hurry, you can keep this as is and we can open an issue for them as they are easily fixable
> 
> Fixed most of them. I think the warning about truthy value does not make much sense.

Thanks. That would be enough. Normally, this means changing on to "on" but feel free to keep this as is now as it is only a warning

sarroutbi (approved):
LGTM

PR #1112, opened 2025-09-03
config: Fix singleton tests

ansasaki:
Remove unnecessary integration tests and fix unit tests to not assume `/var/lib/keylime` will always exist.


sarroutbi (approved):
LGTM

PR #1132, opened 2025-09-25
Fix ECC algorithm selection and reporting for keylime agent

sergio-correia:
- Use AsymmetricAlgorithmSelection instead of AsymmetricAlgorithm to allow ecc521 to properly map to P-521 curve instead of defaulting to P-256
- Change ECC256 algorithm display from generic "ecc" to specific "ecc256"

These changes are required for enabling ECC attestation with the NIST curves.


ansasaki (approved):
LGTM!

PR #1134, opened 2025-09-26
Fix RSA2048 algorithm reporting in keylime agent

sergio-correia:
Change EncryptionAlgorithm::Rsa2048 display from "rsa" to "rsa2048" to be consistent with other specific algorithm reporting (similar to ECC fix in f2979c438).

### DIFF
--- a/keylime-agent/src/agent_handler.rs
+++ b/keylime-agent/src/agent_handler.rs
@@ -109,7 +109,7 @@ mod tests {
         let result: JsonWrapper<AgentInfo> = test::read_body_json(resp).await;
         assert_eq!(result.results.agent_uuid.as_str(), "DEADBEEF");
         assert_eq!(result.results.tpm_hash_alg.as_str(), "sha256");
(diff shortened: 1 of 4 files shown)


ansasaki:
The tests are failing with a tenant error like:
```
2025-09-26 07:14:00.963 - keylime.tenant - ERROR - TPM Quote from Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 (127.0.0.1:9002) is using an unaccepted encryption algorithm: rsa2048
```
Probably we'll need to fix the tenant first to accept both `rsa` and `rsa2048` as equivalent

sergio-correia:
> > Probably we'll need to fix the tenant first to accept both rsa and rsa2048 as equivalent
> 
> Yes the reason I kept the names when implementing this, was that changing those values breaks the the current tenant configuration. So we would add backward compatibility to the tenant first and then change it.

Right, https://github.com/keylime/keylime/pull/1803 should go first indeed.

sergio-correia:
> Some test cases need to be updated:
> 


Thanks, they should be updated now.

PR #1140, opened 2025-10-01
Use simple algorithm specifiers in certification_keys object

sergio-correia:
Update key_algorithm reporting in the certification_keys object to use simple specifiers ("rsa" or "ecc") instead of full algorithm names ("rsa2048", "ecc256", etc.) to comply with the push model protocol specification.

This is not an issue for algorithm identification because the certification_keys object also includes the key_size field, allowing complete identification of the exact algorithm in use.


sarroutbi:
Thanks. I can confirm this change fixes Push Mode attestation:

image

sarroutbi (approved):
LGTM

Issue #1136, opened 2025-09-29
agent fails attestation after a restart


kkaarreell:
When a registered and activated agent is restarted, it fails attestation after the service restart.

The issue is visible in this test log
https://artifacts.dev.testing-farm.io/431d802b-f0a2-41a6-bfaa-64b0b83e044d/work-upstream-keylime-all-tests_uo7oesh/plans/upstream-keylime-all-tests/execute/data/guest/default-0/update/basic-attestation-on-localhost/all-46/output.txt

```
:: [ 12:34:21 ] :: [   PASS   ] :: Command 'limeWaitForAgentRegistration d432fbb3-d2f1-4a97-9ef7-75bd81c00000' (Expected 0, got 0)
:: [ 12:34:26 ] :: [  BEGIN   ] :: Running 'limeWaitForAgentStatus d432fbb3-d2f1-4a97-9ef7-75bd81c00000 'Get Quote''
INFO:keylime.config:Reading configuration from ['/etc/keylime/logging.conf']
2025-09-26 12:34:46.125 - keylime.tenant - WARNING - The configuration upgrade templates path /usr/share/keylime/templates does not exist
2025-09-26 12:34:46.126 - keylime.config - INFO - Reading configuration from ['/etc/keylime/tenant.conf']
2025-09-26 12:34:46.126 - keylime.tenant - INFO - Setting up client TLS...
2025-09-26 12:34:46.126 - keylime.tenant - INFO - Using default client_cert option for tenant
2025-09-26 12:34:46.126 - keylime.tenant - INFO - Using default client_key option for tenant
2025-09-26 12:34:46.126 - keylime.tenant - INFO - No value provided in client_key_password option for tenant, assuming the key is unencrypted
2025-09-26 12:34:46.133 - keylime.tenant - INFO - TLS is enabled.
2025-09-26 12:34:46.150 - keylime.tenant - INFO - Status from Registrar (127.0.0.1:8891): Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 exists on registrar 127.0.0.1 port 8891.
[...]

PR #1141, opened 2025-10-01
Persist the payload key to avoid failing attestation on restart

ansasaki:
Previously, when the agent was restarted while being monitored by the verifier, it would fail attestation because the PCR#16 would be extended with the new ephemeral payload public key.

With this change, the payload key is persisted, making the agent to reuse the previously generated payload key instead of generating a new ephemeral key on restart.

Resolves: #1136

### DIFF
--- a/keylime-agent.conf
+++ b/keylime-agent.conf
@@ -87,6 +87,28 @@ server_key = "default"
 # environment variable.
 server_key_password = ""
 
+# The name of the file containing the payload encryption private key.
+# This private key is used to decrypt U and V keys sent by the verifier/tenant
+# for secure payload transmission.
+# A new RSA 2048 private key is generated in case it is not found.
+# If an existing key is provided, it MUST be RSA 2048, otherwise the agent will
+# fail to start.
+# If set as "default", the "payload-private.pem" value is used.
+# If a relative path is set, it will be considered relative from the keylime_dir.
+# If an absolute path is set, it is used without change.
+#
+# To override payload_key, set KEYLIME_AGENT_PAYLOAD_KEY environment variable.
+payload_key = "default"
+
+# Set the password used to encrypt the payload private key file.
+# This password will also be used to protect the generated private key used for
+# payload encryption.
+# If left empty, the private key will not be encrypted.
+#
+# To override payload_key_password, set KEYLIME_AGENT_PAYLOAD_KEY_PASSWORD
+# environment variable.
+payload_key_password = ""
+
 # The name of the file containing the X509 certificate used as the Keylime agent
 # server TLS certificate.
 # This certificate must be self signed.
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -490,47 +490,38 @@ async fn main() -> Result<()> {
         (None, None)
     };
 
-    // Generate ephemeral RSA key pair for secure transmission of u, v keys.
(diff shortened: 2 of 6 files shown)


sarroutbi (commented):
Suggestions for Improvement:

- Inconsistent Validation on Keys:
    In keylime-agent/src/main.rs, the call to load_or_generate_key for the payload key correctly enables validation (validate_algorithm: true), which is excellent. However, the call for the mTLS key explicitly disables it (validate_algorithm: false) with a comment about "backward compatibility".
  - Recommendation: While backward compatibility is important, this introduces a potential security weakness. An administrator could mistakenly configure a weak mTLS key without the agent flagging the error. It would be better to make validation the default behavior. A path forward could be to log a prominent warning if an existing mTLS key is loaded and fails validation, while enforcing validation for all newly generated keys. This would encourage users to upgrade their keys without breaking existing setups immediately.

- Potential for Unencrypted Keys:
    The configuration allows for an empty password for the payload_key_password, resulting in an unencrypted private key on disk.

  - Recommendation: While this provides flexibility, it's a less secure default. Consider adding a warning log message during startup if the agent detects it is writing or loading an unencrypted payload key. This would alert the administrator to a potentially insecure configuration without forcing them to use a password. For example: WARN: Payload key '/path/to/key' is not encrypted. It is highly recommended to set 'payload_key_password' in the configuration.

ansasaki:
> Suggestions for Improvement:
> 
>     * Inconsistent Validation on Keys:
>       In keylime-agent/src/main.rs, the call to load_or_generate_key for the payload key correctly enables validation (validate_algorithm: true), which is excellent. However, the call for the mTLS key explicitly disables it (validate_algorithm: false) with a comment about "backward compatibility".
>       
>       * Recommendation: While backward compatibility is important, this introduces a potential security weakness. An administrator could mistakenly configure a weak mTLS key without the agent flagging the error. It would be better to make validation the default behavior. A path forward could be to log a prominent warning if an existing mTLS key is loaded and fails validation, while enforcing validation for all newly generated keys. This would encourage users to upgrade their keys without breaking existing setups immediately.

I disagree. The mTLS keys can use any supported algorithm, while the payload key MUST be RSA 2048. The validation is to check if the key provided in the file is RSA 2048 for the payload key. There is no point in restricting the algorithms that can be used for mTLS: actually it is even counter-productive as in future the algorithms supported by the underlying library (OpenSSL) can be extended, meaning that new algorithms would be automatically supported without changes to the code.

>     * Potential for Unencrypted Keys:
>       The configuration allows for an empty password for the payload_key_password, resulting in an unencrypted private key on disk.
>       
>       * Recommendation: While this provides flexibility, it's a less secure default. Consider adding a warning log message during startup if the agent detects it is writing or loading an unencrypted payload key. This would alert the administrator to a potentially insecure configuration without forcing them to use a password. For example: WARN: Payload key '/path/to/key' is not encrypted. It is highly recommended to set 'payload_key_password' in the configuration.

I think this simply gives a false sense of security: when the payload script is encrypted, the password is basically set in the configuration file. It would be as hard for the attacker to get access to the unencrypted key or to the plaintext configuration file that contains the password, since both rely on the file access permissions to protect their content. And I don't see how encrypting with a default password would be different from not encrypting it at all. What would actually improve security would be to store the keys sealed against the PCR state or, even better, store outside of the disk (e.g. PKCS#11 device).

sarroutbi (approved):

PR #1154, opened 2025-12-01
Fix agent handling of 403 Forbidden registration responses

sarroutbi:
The agent was incorrectly interpreting 403 Forbidden responses from the registrar as API version incompatibility errors. This caused two problems:

1. The agent would try all enabled API versions, even though 403 indicates a permanent security rejection (e.g., TPM identity mismatch during re-registration)

2. The agent would continue running after registration failure, making it appear operational when it was not properly registered

This issue became apparent with the Python keylime registrar security fix for CVE-2025-13609 (duplicate UUID vulnerability), which returns 403 Forbidden when an agent attempts to re-register with a different TPM identity.

The agent will now correctly fail fast when the registrar rejects registration for security reasons, allowing proper error detection in tests and production deployments.

Related: keylime/keylime#1820 (Python registrar UUID spoofing fix)

### DIFF
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -623,10 +623,24 @@ async fn main() -> Result<()> {
         ak_handle,
         retry_config: None,
     };
+    use keylime::error::Error;
+    use keylime::registrar_client::RegistrarClientError;
+
     match keylime::agent_registration::register_agent(aa, &mut ctx).await {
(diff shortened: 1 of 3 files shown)


sergio-correia (approved):
Thanks, this looks good to me. In the future we can add some testing to cover the new `RegistrationForbidden` path.

Issue #1226, opened 2026-04-21
TPM2_Certify fails with TPM_RC_SIZE when agent ID exceeds TPM2B_DATA limit


ansasaki:
## Summary

The agent passes the agent ID string directly as qualifying data to `TPM2_Certify` during IAK-based AK certification. When the agent ID exceeds the TPM's `TPM2B_DATA` maximum size, the TPM rejects the command with `TPM_RC_SIZE`.

## Problem

In `keylime-agent/src/main.rs`, the agent uses the raw agent ID bytes as qualifying data:

```rust
let qualifying_data = Data::try_from(agent_uuid.as_bytes())?;
let (attest, signature) =
    dev_id.certify(qualifying_data, ak_handle, &mut ctx)?;
```

`TPM2B_DATA` has a maximum size of `sizeof(TPMT_HA)` = `2 + largest_hash_digest` as defined by the specific TPM's supported algorithms (TPM 2.0 spec, Part 2, Section 10.4.2). For a TPM that only supports up to SHA-256, this limit is `2 + 32 = 34 bytes`.

A standard UUID string like `d432fbb3-d2f1-4a97-9ef7-75bd81c00000` is 36 bytes, which exceeds the 34-byte limit. Keylime also supports arbitrary strings as agent IDs, which can be even longer.

Note: the tss-esapi library on the host compiles `TPMU_HA` with SHA-512 support (64 bytes), so `Data::try_from()` succeeds on the software side. The error only surfaces when the TPM hardware processes the command.

## Error

```
WARNING:esys:src/tss2-esys/api/Esys_Certify.c:331:Esys_Certify_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_Certify.c:107:Esys_Certify() Esys Finish ErrorCode (0x000001d5)
Error: DeviceID(Certify(Tss2 { err: Tss2Error(FormatOne(FormatOneResponseCode {
    .0: 469, error_number: 21, parameter: true, format_selector: true, number: 1
})), kind: Some(Size), message: "structure is the wrong size (associated with parameter number 1)" }))
```

Error code `0x000001d5` = `TPM_RC_SIZE` on parameter 1 (`qualifyingData`).

## Environment

- Machine: NVIDIA Grace Hopper (aarch64)
- TPM: hardware TPM (likely SHA-256-only, giving a 34-byte `TPM2B_DATA` limit)
- Agent ID: standard UUID format (36 bytes as string)

## Proposed fix

[...]

PR #1239, opened 2026-05-12
tpm: Hash qualifying data in TPM2_Certify for IAK-based AK certification

ansasaki:
tpm: Hash qualifying data in TPM2_Certify for IAK-based AK certification

## Type of Change
*(Select all that apply)*
- [x] Bug fix (non-breaking change)
- [ ] New feature (non-breaking change)
- [ ] Breaking change (fix/feature causing existing behavior to change)
- [ ] Documentation update (standalone)
- [ ] Code refactor (no functional changes)
- [x] Test cases (added/modified)
- [ ] CI/CD changes
- [ ] Other (please specify: ______)

## Related Issues
**Resolves:** https://github.com/keylime/rust-keylime/issues/1226
**See also:** https://github.com/keylime/keylime/issues/1892

### Concise Summary
Hash agent ID (SHA-256) before passing it as qualifying data to `TPM2_Certify`.

### Technical Details

When the agent certifies the AK with the IAK using `TPM2_Certify`, it passes the agent ID string directly as `qualifyingData` (`TPM2B_DATA`). The TPM 2.0 spec limits `TPM2B_DATA` to `sizeof(TPMT_HA)` bytes, which depends on the TPM's supported hash algorithms. On SHA-256-only TPMs (e.g., NVIDIA Grace Hopper aarch64), this limit is 34 bytes — smaller than a standard UUID string (36 bytes). Since Keylime supports arbitrary strings as agent IDs, the qualifying data can be even longer.

This PR hashes the agent ID with SHA-256 before using it as qualifying data, producing a fixed 32-byte value that fits within `TPM2B_DATA` on any TPM. The qualifying data is arbitrary per the TPM 2.0 spec (Part 3, Section 18.2) — hashing it is fully compliant.

**Changes:**

- **Hash agent ID in `main.rs`**: Use `keylime::crypto::hash` with `MessageDigest::sha256()` to hash the agent ID before passing it to `TPM2_Certify`. No new dependencies — uses the existing `openssl` crate via the keylime crypto module.

- **Update test in `tpm.rs`**: `test_certify_credential_with_iak` now hashes the qualifying data to match the production code path.

- **Bump pull-mode API version to 2.6** in `config/base.rs`.

## Documentation Updates Required
*(Check all that apply)*
- [ ] Updated markdown docs (file path: ______)
- [ ] Updated code comments/docstrings
- [ ] Needs user guide updates (`docs/`)
- [ ] Needs ReadTheDocs updates
- [x] No docs needed (requires maintainer approval)

Documentation updates are included in the companion Python PR.

## Verification Process
1. Run `cargo fmt && cargo clippy --all-targets` — no new warnings
2. Run `cargo test` — existing tests pass
3. Integration test with swtpm:
   - Verify that a new agent (2.6) registers successfully with the updated registrar
   - Verify that registration works with agent IDs longer than 34 bytes (e.g., non-UUID strings)

## Checklist
- [x] Code follows project style guidelines
- [x] Unit/integration tests added/updated
- [x] Documentation updated (per above section)
- [x] Commit messages follow [Chris Beams' How to Write a Git Commit Message article](https://chris.beams.io/posts/git-commit/)
- [ ] CHANGELOG updated (if applicable)
- [x] All tests pass (local & CI)

## Additional Context

- **Deployment ordering**: The companion Python PR (https://github.com/keylime/keylime/issues/1892) **must be merged first**. It adds backward-compatible "try both" support to the registrar, accepting both hashed (new agents) and raw (old agents) qualifying data. If this Rust PR is merged first, agents on SHA-256-only TPMs with long agent IDs will send hashed qualifying data to a registrar that cannot verify it.
- No new dependencies were added. The hash is computed using the existing `keylime::crypto::hash` function backed by `openssl`.

> This PR was created with the assistance of Claude Opus 4.6 (Anthropic).

## Summary by CodeRabbit

* **New Features**
  * Added support for API version 2.6.

* **Updates**
  * Agent registration now computes IAK/IDevID attestations at registration time and selects qualifying data based on registrar API version (SHA-256 for 2.6+).

* **Tests / CI**
  * Unit tests adjusted for the new qualifying-data behavior; e2e scenario now pins upstream source/branch used by CI.

[Review Change Stack](https://app.coderabbit.ai/change-stack/keylime/rust-keylime/pull/1239)

### DIFF
--- a/keylime-agent/src/api.rs
+++ b/keylime-agent/src/api.rs
@@ -188,6 +188,25 @@ fn configure_api_v2_5(cfg: &mut web::ServiceConfig) {
     )
 }
 
+/// Configure the endpoints supported by API version 2.6
+///
+/// Version 2.6 has the same agent-side endpoints as 2.5. The version bump
+/// reflects a protocol change: the agent now SHA-256 hashes the agent ID
+/// before using it as qualifying data in TPM2_Certify (IAK-based AK
+/// certification). This ensures the qualifying data fits within the
+/// TPM2B_DATA size limit on SHA-256-only TPMs (34 bytes). The certify
+/// operation is deferred until after API version negotiation so the agent
+/// can fall back to raw UUID bytes when the registrar only supports 2.5.
+fn configure_api_v2_6(cfg: &mut web::ServiceConfig) {
+    let version = Version::from_str("2.6").expect("Invalid API version");
+    _ = cfg.app_data(web::Data::new(version));
+    configure_base_endpoints(cfg);
+    _ = cfg.service(
+        web::scope("/agent")
+            .configure(agent_handler::configure_agent_endpoints),
+    )
+}
+
 /// Get a scope configured for the given API version
 pub(crate) fn get_api_scope(version: &str) -> Result<Scope, APIError> {
     match version {
@@ -201,6 +220,8 @@ pub(crate) fn get_api_scope(version: &str) -> Result<Scope, APIError> {
             .configure(configure_api_v2_4)),
         "2.5" => Ok(web::scope(format!("v{version}").as_ref())
             .configure(configure_api_v2_5)),
+        "2.6" => Ok(web::scope(format!("v{version}").as_ref())
+            .configure(configure_api_v2_6)),
         _ => Err(APIError::UnsupportedVersion(version.into())),
     }
 }
--- a/keylime-agent/src/main.rs
+++ b/keylime-agent/src/main.rs
@@ -87,7 +87,7 @@ use tss_esapi::{
     handles::KeyHandle,
     interface_types::algorithm::{AsymmetricAlgorithm, HashingAlgorithm},
     interface_types::resource_handles::Hierarchy,
-    structures::{Auth, Data, Digest, MaxBuffer, PublicBuffer},
+    structures::{Auth, Digest, MaxBuffer, PublicBuffer},
     traits::Marshall,
     Context,
 };
@@ -494,24 +494,6 @@ async fn main() -> Result<()> {
         None
     };
 
-    let (attest, signature) = if let Some(dev_id) = &mut device_id {
-        let qualifying_data = Data::try_from(agent_uuid.as_bytes())?;
-        let (attest, signature) =
-            dev_id.certify(qualifying_data, ak_handle, &mut ctx)?;
-
-        info!("AK certified with IAK.");
-
-        // // For debugging certify(), the following checks the generated signature
-        // let max_b = MaxBuffer::try_from(attest.clone().marshall()?)?;
-        // let (hashed_attest, _) = ctx.inner.hash(max_b, HashingAlgorithm::Sha256, Hierarchy::Endorsement,)?;
-        // println!("{:?}", hashed_attest);
-        // println!("{:?}", signature);
-        // println!("{:?}", ctx.inner.verify_signature(iak.as_ref().unwrap().handle, hashed_attest, signature.clone())?); //#[allow_ci]
-        (Some(attest), Some(signature))
-    } else {
-        (None, None)
-    };
-
     // Load or generate RSA key pair for secure transmission of u, v keys.
     // The u, v keys are two halves of the key used to decrypt the workload after
     // the Identity and Integrity Quotes sent by the agent are validated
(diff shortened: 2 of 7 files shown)


keylime-agent/src/main.rs
```diff
@@ -495,7 +495,11 @@ async fn main() -> Result<()> {
     };
 
     let (attest, signature) = if let Some(dev_id) = &mut device_id {
-        let qualifying_data = Data::try_from(agent_uuid.as_bytes())?;
+        let hash = crypto::hash(
+            agent_uuid.as_bytes(),
+            openssl::hash::MessageDigest::sha256(),
+        )?;
+        let qualifying_data = Data::try_from(hash.as_slice())?;
```
sergio-correia: The hashing looks correct, but it runs unconditionally here. Later on there is the API version negotiation with the registrar, and it seems we woud have an issue if e.g. the registrar only supports v2.5: the agent would still send `extra_data = SHA-256(UUID)`, which would then be rejected by the registrar expecting raw UUID bytes. Shouldn't we enforce `API >= 2.6` when `enable_iak_idevid` is enabled?

keylime-agent/src/api.rs
```diff
@@ -188,6 +188,16 @@ fn configure_api_v2_5(cfg: &mut web::ServiceConfig) {
     )
 }
 
+fn configure_api_v2_6(cfg: &mut web::ServiceConfig) {
```
sergio-correia, line 200: Every other configure_api_v2_X function has a doc comment explaining what changed in that version. We should probably add something here as well mentioning the protocol change on the hashed qualifying data.

keylime/src/tpm.rs
```diff
@@ -3291,10 +3291,14 @@ pub mod tests {
...
-        let qualifying_data = "some_uuid".as_bytes();
+        let qualifying_data = crate::crypto::hash(
+            "some_uuid".as_bytes(),
+            openssl::hash::MessageDigest::sha256(),
+        )
+        .unwrap(); //#[allow_ci]
 
         let r = ctx.certify_credential_with_iak(
-            Data::try_from(qualifying_data).unwrap(), //#[allow_ci]
+            Data::try_from(qualifying_data.as_slice()).unwrap(), //#[allow_ci]
```
sergio-correia: The test checks that certify succeeds, but does not verify that the returned `Attest` structure's `extra_data` actually contains the expected hash. Perhaps we can add something like `assert_eq!(attest.extra_data().value(), qualifying_data.as_slice()`?

keylime/src/tpm.rs
```diff
@@ -3291,10 +3291,14 @@ pub mod tests {
             .expect("failed to create IAK")
             .handle;
 
-        let qualifying_data = "some_uuid".as_bytes();
+        let qualifying_data = crate::crypto::hash(
+            "some_uuid".as_bytes(),
```
sergio-correia: Just a nit here: `"some_uuid"` is 9 bytes. It might be worth using a realistic UUID string (36 bytes) here since that is the exact input size that triggers the bug this fixes.

keylime-agent/src/main.rs
ansasaki: Good point, we have to guard this to not affect older API versions.

keylime-agent/src/api.rs
ansasaki: I'll add the documentation comment

keylime/src/tpm.rs
ansasaki: Yes, this makes sense, we should check if the value is the expected one

keylime/src/tpm.rs
ansasaki: Right, better to use more realistic input

keylime-agent/src/main.rs
ansasaki: Done!

sarroutbi (approved):
Changes LGTM. Please, remember to drop `[DO NOT MERGE]`(https://github.com/keylime/rust-keylime/pull/1239/changes/81a9b9bb76f2655dda446b61116f573436804bde) commit before pushing

sergio-correia (approved):
Thanks for the updates, LGTM

PR #1259, opened 2026-07-03
fix: use runtime OpenSSL algorithm detection to handle missing SM3 support

ansasaki:
## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor (no functional changes)
- [ ] Tests
- [ ] CI/CD
- [ ] Other

## Related Issues

See also: Fedora COPR build failure — `openssl::hash::MessageDigest::sm3()` not found when building against Fedora's OpenSSL (compiled with `OPENSSL_NO_SM3`).

### Concise Summary

Replace all `MessageDigest::sha*()` / `MessageDigest::sm3()` associated function calls with `MessageDigest::from_name()` runtime lookups, making the code resilient to any hash algorithm being removed from OpenSSL without requiring code changes.

### Technical Details

Fedora's OpenSSL is now built with `OPENSSL_NO_SM3`, which removes `MessageDigest::sm3()` from the `openssl` crate at compile time. This caused two build errors in `algorithms.rs` and `tpm.rs`.

Rather than using compile-time `cfg` gates (which don't propagate transitively from the `openssl` crate), the fix uses `MessageDigest::from_name()` — a runtime lookup via `EVP_get_digestbyname` — for **all** hash algorithms. This is future-proof: if OpenSSL drops another algorithm, the code handles it gracefully without any changes.

**Changes:**

- **`algorithms.rs`**: `From<HashAlgorithm> for MessageDigest` becomes `TryFrom`, using `from_name()` for all algorithms. Adds `digest_size()` (compile-time constant sizes), `openssl_name()`, and `is_supported()` methods to `HashAlgorithm`.
- **`tpm.rs`**: `hash_alg_to_message_digest` now delegates to the `TryFrom` impl instead of duplicating the conversion logic. `get_supported_hash_algorithms` filters out algorithms unavailable in the current OpenSSL build, preventing the agent from advertising capabilities it cannot fulfill.
- **`entry.rs`**: `Digest::start()` and `Digest::ff()` use `digest_size()` instead of going through OpenSSL — they only need byte lengths, not digest objects. `Digest::new()` also uses `digest_size()` for validation. This removes the OpenSSL dependency from `entry.rs` entirely.
- **`context_info.rs`**, **`keylime-ima-emulator/main.rs`**: Call sites updated from infallible `.into()` to `TryFrom`/`try_from`.

### What is NOT changed

- No `Cargo.toml` changes (no new dependencies, no lint configs)
- No RPM spec or packit changes
- No `cfg` gates or `build.rs` files
- `HashAlgorithm::Sm3_256` enum variant remains (TPM protocol compatibility)
- `PcrBanks.sm3_256` field remains (wire format / deserialization compatibility)
- `capabilities_negotiation.rs` and `struct_filler.rs` are unchanged — they are pure data structures that don't use OpenSSL

### Alternatives Considered

1. **Compile-time `cfg(osslconf = "OPENSSL_NO_SM3")`**: The openssl crate's `osslconf` cfg doesn't propagate to downstream crates, requiring `build.rs` + `openssl-sys` dependency additions — more complex for no benefit.
2. **Cargo feature flag**: Would require manual coordination between spec files, packit, and Cargo.toml on every platform.

## Documentation Updates Required

- [ ] README
- [ ] Configuration documentation
- [ ] API documentation
- [x] None required

## Verification Process

1. Build on Fedora (OpenSSL with `OPENSSL_NO_SM3`): `cargo build` — compiles cleanly
2. Build on systems with SM3 support: `cargo build` — compiles cleanly
3. Run test suite: `bash tests/run.sh` — all tests pass
4. Verify no `MessageDigest::sm3()` references remain: `grep -r "MessageDigest::sm3" --include="*.rs" | grep -v target/`

## Checklist

- [x] Code follows the project style guidelines
- [x] Tests have been added/updated
- [ ] Documentation has been updated
- [x] Commit messages follow the project conventions
- [ ] CHANGELOG has been updated
- [x] All tests pass locally

## Additional Context

> **Note:** This PR was developed with AI assistance (Claude).

## Summary by CodeRabbit

* **New Features**
  * Added hash-algorithm availability checks so only algorithms supported by the current OpenSSL build are used.
* **Bug Fixes**
  * Digest creation now returns errors when an algorithm isn’t supported, eliminating implicit conversion/silent fallback behavior.
  * Improved IMA digest initialization and validation; `start`/`ff` now return `Result` to reflect potential failures.
  * Updated TPM and context initialization to surface unsupported-algorithm errors consistently.
* **Tests**
  * Updated hashing-to-digest tests to reflect the new supported/unsupported behavior (including SM3).

### DIFF
--- a/keylime/src/ima/entry.rs
+++ b/keylime/src/ima/entry.rs
@@ -32,7 +32,8 @@ pub struct Digest {
 impl Digest {
     /// Creates a new `Digest` with `algorithm` and `value`.
     pub fn new(algorithm: HashAlgorithm, value: &[u8]) -> Result<Self> {
-        let digest: MessageDigest = algorithm.into();
+        let digest: MessageDigest =
+            MessageDigest::try_from(algorithm).map_err(Error::other)?;
         if value.len() != digest.size() {
             return Err(Error::new(
                 ErrorKind::InvalidInput,
@@ -54,22 +55,24 @@ impl Digest {
 
     /// Returns a pre-defined digest value used to indicate the start
     /// of the IMA measurement list.
-    pub fn start(algorithm: HashAlgorithm) -> Self {
-        let digest: MessageDigest = algorithm.into();
-        Self {
+    pub fn start(algorithm: HashAlgorithm) -> Result<Self> {
+        let digest: MessageDigest =
+            MessageDigest::try_from(algorithm).map_err(Error::other)?;
+        Ok(Self {
             algorithm,
             value: vec![0x00u8; digest.size()],
-        }
+        })
     }
 
     /// Returns a pre-defined digest value used to indicate the ToMToU
     /// error in the IMA measurement list.
-    pub fn ff(algorithm: HashAlgorithm) -> Self {
-        let digest: MessageDigest = algorithm.into();
-        Self {
+    pub fn ff(algorithm: HashAlgorithm) -> Result<Self> {
+        let digest: MessageDigest =
+            MessageDigest::try_from(algorithm).map_err(Error::other)?;
+        Ok(Self {
             algorithm,
             value: vec![0xffu8; digest.size()],
-        }
+        })
     }
 }
 
--- a/keylime/src/tpm.rs
+++ b/keylime/src/tpm.rs
@@ -1818,6 +1818,16 @@ impl Context<'_> {
                 }
             }
         }
+        usable_algs.retain(|alg| {
+            let supported = alg.is_supported();
+            if !supported {
+                log::warn!(
+                    "Hash algorithm {alg} supported by TPM but not by \
+                     this OpenSSL build; excluding"
+                );
+            }
+            supported
+        });
         Ok(usable_algs)
     }
 
@@ -2329,14 +2339,14 @@ fn make_pcr_blob(
 fn hash_alg_to_message_digest(
     hash_alg: HashingAlgorithm,
 ) -> Result<MessageDigest> {
-    match hash_alg {
-        HashingAlgorithm::Sha256 => Ok(MessageDigest::sha256()),
-        HashingAlgorithm::Sha1 => Ok(MessageDigest::sha1()),
-        HashingAlgorithm::Sha384 => Ok(MessageDigest::sha384()),
-        HashingAlgorithm::Sha512 => Ok(MessageDigest::sha512()),
-        HashingAlgorithm::Sm3_256 => Ok(MessageDigest::sm3()),
-        other => Err(TpmError::UnsupportedHashingAlgorithm { alg: other }),
-    }
+    let keylime_alg = KeylimeInternalHashAlgorithm::try_from(hash_alg)
+        .map_err(|_| TpmError::UnsupportedHashingAlgorithm {
+            alg: hash_alg,
+        })?;
+    MessageDigest::try_from(keylime_alg).map_err(|e| {
+        log::debug!("OpenSSL digest lookup failed for {hash_alg:?}: {e}");
+        TpmError::UnsupportedHashingAlgorithm { alg: hash_alg }
+    })
 }
 
 /// Check if the data attested in the quote matches the data read from the TPM PCRs
--- a/keylime-ima-emulator/src/main.rs
+++ b/keylime-ima-emulator/src/main.rs
@@ -62,11 +62,11 @@ fn ml_extend(
 ) -> Result<usize> {
     let f = File::open(ml)?;
     let mut reader = BufReader::new(f);
-    let ima_digest: MessageDigest = ima_hash_alg.into();
-    let ima_start_hash = ima::Digest::start(ima_hash_alg);
-    let pcr_digest: MessageDigest = pcr_hash_alg.into();
-    let mut running_hash = ima::Digest::start(pcr_hash_alg);
-    let ff_hash = ima::Digest::ff(pcr_hash_alg);
+    let ima_digest: MessageDigest = MessageDigest::try_from(ima_hash_alg)?;
+    let ima_start_hash = ima::Digest::start(ima_hash_alg)?;
+    let pcr_digest: MessageDigest = MessageDigest::try_from(pcr_hash_alg)?;
+    let mut running_hash = ima::Digest::start(pcr_hash_alg)?;
+    let ff_hash = ima::Digest::ff(pcr_hash_alg)?;
     for line in reader.by_ref().lines().skip(position) {
         let line = line?;
         if line.is_empty() {
@@ -198,7 +198,8 @@ fn main() -> std::result::Result<(), ImaEmulatorError> {
                 )
             })?;
 
-        let pcr_digest: MessageDigest = (*pcr_hash_alg).into();
+        let pcr_digest: MessageDigest =
+            MessageDigest::try_from(*pcr_hash_alg)?;
         let pcr_start_hash = vec![0x00u8; pcr_digest.size()];
         if digest.value() != pcr_start_hash {
             log::warn!("IMA PCR is not empty, trying to find the last updated file in the measurement list...");
--- a/keylime/src/algorithms.rs
+++ b/keylime/src/algorithms.rs
@@ -113,15 +113,32 @@ impl From<HashAlgorithm> for HashingAlgorithm {
     }
 }
 
-impl From<HashAlgorithm> for MessageDigest {
-    fn from(hash_algorithm: HashAlgorithm) -> Self {
-        match hash_algorithm {
-            HashAlgorithm::Sha1 => MessageDigest::sha1(),
-            HashAlgorithm::Sha256 => MessageDigest::sha256(),
-            HashAlgorithm::Sha384 => MessageDigest::sha384(),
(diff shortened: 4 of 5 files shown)


keylime/src/ima/entry.rs
```diff
@@ -55,7 +56,8 @@ impl Digest {
     /// Returns a pre-defined digest value used to indicate the start
     /// of the IMA measurement list.
     pub fn start(algorithm: HashAlgorithm) -> Self {
-        let digest: MessageDigest = algorithm.into();
+        let digest: MessageDigest = MessageDigest::try_from(algorithm)
```
sergio-correia: Before this commit, `From<HashAlgorithm> for MessageDigest` was infallible so this `.expect()` could never fire. Now that it's `TryFrom`, calling `Digest::start(HashAlgorithm::Sm3_256)` on a no-SM3 build will panic. The current caller in the IMA emulator happens to be protected by a preceding `try_from` with `?`, but these are `pub` methods.

`Digest::new()` already returns `Result` and handles the same conversion with `.map_err(Error::other)?`. Would it make sense to change `start()` and `ff()` to match that pattern?

```rust
pub fn start(algorithm: HashAlgorithm) -> Result<Self> {
    let digest: MessageDigest =
        MessageDigest::try_from(algorithm).map_err(Error::other)?;
    Ok(Self {
        algorithm,
        value: vec![0x00u8; digest.size()],
    })
}
```

keylime/src/tpm.rs
```diff
@@ -2334,7 +2334,12 @@ fn hash_alg_to_message_digest(
         HashingAlgorithm::Sha1 => Ok(MessageDigest::sha1()),
         HashingAlgorithm::Sha384 => Ok(MessageDigest::sha384()),
         HashingAlgorithm::Sha512 => Ok(MessageDigest::sha512()),
+        #[cfg(not(osslconf = "OPENSSL_NO_SM3"))]
```
sergio-correia: This SM3 cfg-gating logic is duplicated between here and `TryFrom<HashAlgorithm> for MessageDigest` in `algorithms.rs`. Both have identical `#[cfg(not(osslconf = "OPENSSL_NO_SM3"))]` / `#[cfg(osslconf = "OPENSSL_NO_SM3")]` pairs. If one is updated and the other forgotten, the guard breaks for a subset of code paths.

Would it be cleaner to have this function delegate to the `TryFrom` implementation?

```rust
fn hash_alg_to_message_digest(hash_alg: HashingAlgorithm) -> Result<MessageDigest> {
    let keylime_alg = KeylimeInternalHashAlgorithm::try_from(hash_alg)
        .map_err(|_| TpmError::UnsupportedHashingAlgorithm { alg: hash_alg })?;
    MessageDigest::try_from(keylime_alg)
        .map_err(|_| TpmError::UnsupportedHashingAlgorithm { alg: hash_alg })
}
```

keylime/src/structures/capabilities_negotiation.rs
```diff
@@ -168,6 +168,7 @@ impl PcrBanks {
         if let Some(v) = &self.sha512 {
             map.insert("sha512".to_string(), v.clone());
         }
+        #[cfg(not(osslconf = "OPENSSL_NO_SM3"))]
```
sergio-correia: When `OPENSSL_NO_SM3` is set and `self.sm3_256` is `Some(...)` (e.g., deserialized from a verifier response), this silently drops the data with no log message. The `struct_filler.rs` code already logs `warn!("Ignoring SM3_256 PCR bank...")` for the same situation. Maybe add a similar warning here?

```rust
#[cfg(osslconf = "OPENSSL_NO_SM3")]
if self.sm3_256.is_some() {
    log::warn!("Dropping SM3_256 PCR bank data: not supported by this OpenSSL build");
}
```

sergio-correia (commented):
A few issues to address, mostly around completeness of the SM3 gating.

Not in this diff, but related: `get_supported_hash_algorithms()` in `keylime/src/tpm.rs:1768-1822` returns SM3 from the TPM unfiltered. The PCR banks in `available_subjects` are correctly filtered (the new cfg guard in `struct_filler.rs`), and `PcrBanks::to_map()` also drops SM3. But the `hash_algorithms` field at `struct_filler.rs:249` calls `get_supported_hash_algorithms()` with no SM3 filtering. If a verifier selects SM3 from the advertised list, the agent fails at `hash_alg_to_message_digest()`.

The simplest fix would be to filter SM3 from the returned list in `get_supported_hash_algorithms()`:

```rust
// After building usable_algs, before returning:
#[cfg(osslconf = "OPENSSL_NO_SM3")]
usable_algs.retain(|alg| !matches!(alg, KeylimeInternalHashAlgorithm::Sm3_256));
```

keylime/src/ima/entry.rs
ansasaki: Thank you! I followed the suggestion and modified them to return `Result`

keylime/src/tpm.rs
ansasaki: Dropped the configuration guards that didn't work, replace them with runtime detection using `MessageDigest::from_name()`

keylime/src/structures/capabilities_negotiation.rs
ansasaki: This warning is no longer necessary as the algorithms are detected in runtime, not hidden behind the configuration guards

sergio-correia (commented):
A few things to address, mostly around diagnostic logging and fail-fast behavior.

Not in this diff, but related: `ContextInfo::new()` in `keylime/src/context_info.rs:160` accepts the configured `tpm_hash_alg` and immediately proceeds to create EK/AK objects in the TPM without checking `is_supported()`. OpenSSL support is only validated much later when `get_ak_local_identifier_str()` or quote operations invoke `MessageDigest::try_from()`.

If an operator configures `tpm_hash_alg = "sm3_256"` on a system where OpenSSL lacks SM3, the agent starts, creates EK/AK handles, registers with the verifier, then fails at the first attestation attempt. Would it make sense to add an early `is_supported()` check before the TPM operations?

```rust
let tpm_hash_alg = config.tpm_hash_alg;
if !tpm_hash_alg.is_supported() {
    return Err(ContextInfoError::InvalidAlgorithm(
        AlgorithmError::UnsupportedHashingAlgorithm(format!(
            "{tpm_hash_alg} (not supported by this OpenSSL build)"
        ))
    ));
}
```

keylime/src/tpm.rs
```diff
@@ -1818,6 +1818,7 @@ impl Context<'_> {
                 }
             }
         }
+        usable_algs.retain(|alg| alg.is_supported());
```
sergio-correia: This silently drops algorithms that the TPM supports but OpenSSL doesn't. On a system where SM3 is the only TPM-supported hash (e.g., Chinese regulation compliance), the returned list could be empty with no explanation, making field debugging difficult.

Maybe add a `log::warn!` for each dropped algorithm?

```rust
usable_algs.retain(|alg| {
    let supported = alg.is_supported();
    if !supported {
        log::warn!(
            "Hash algorithm {alg} supported by TPM but not by \
             this OpenSSL build; excluding"
        );
    }
    supported
});
```

keylime/src/tpm.rs
```diff
@@ -2329,14 +2330,12 @@ fn make_pcr_blob(
...
-        HashingAlgorithm::Sha512 => Ok(MessageDigest::sha512()),
-        HashingAlgorithm::Sm3_256 => Ok(MessageDigest::sm3()),
-        other => Err(TpmError::UnsupportedHashingAlgorithm { alg: other }),
-    }
+    let keylime_alg = KeylimeInternalHashAlgorithm::try_from(hash_alg)
+        .map_err(|_| TpmError::UnsupportedHashingAlgorithm {
+            alg: hash_alg,
+        })?;
+    MessageDigest::try_from(keylime_alg)
+        .map_err(|_| TpmError::UnsupportedHashingAlgorithm { alg: hash_alg })
```
sergio-correia: Both `.map_err(|_| ...)` closures discard the inner error. The `AlgorithmError` from `TryFrom` contains `"sm3 (not supported by this OpenSSL build)"`, but the `TpmError` only shows `"Unsupported hashing algorithm : Sm3_256"`. The user has to guess whether the issue is keylime, the TPM, or OpenSSL.

Would it help to at least log the inner error before discarding it?

```rust
MessageDigest::try_from(keylime_alg).map_err(|e| {
    log::debug!("OpenSSL digest lookup failed for {hash_alg:?}: {e}");
    TpmError::UnsupportedHashingAlgorithm { alg: hash_alg }
})
```

ansasaki: Done, thanks!

keylime/src/tpm.rs
ansasaki: Done, thanks!

sergio-correia (approved):
LGTM, thanks!

PR #1272, opened 2026-08-20
Unwrap OCTET STRING wrapped EK certificates

### LINKED ISSUE #1225: Agent sends OCTET STRING-wrapped EK certificate, breaking older registrars
## Summary

When a TPM stores its EK certificate in NV RAM wrapped in an ASN.1 OCTET STRING (tag `0x04`), the Rust agent sends this wrapped data to the registrar as-is. Older registrars (and the current Python registrar's `cert_utils.x509_der_cert`) cannot parse this format, resulting in a 400 error during agent registration.

## Problem

The `check_ek_cert` method in `keylime/src/tpm.rs` performs a generic ASN.1 roundtrip using `picky_asn1_der::Asn1RawDer`:

```rust
fn check_ek_cert(&mut self, cert: &[u8]) -> Result<Vec<u8>> {
    let parsed_cert: picky_asn1_der::Asn1RawDer =
        picky_asn1_der::from_bytes(cert)?;
    Ok(picky_asn1_der::to_vec(&parsed_cert)?)
}
```

This strips trailing padding but does **not** detect or unwrap an OCTET STRING wrapper around the certificate. An X.509 certificate must start with a SEQUENCE tag (`0x30`), but the data sent by the agent starts with OCTET STRING (`0x04`).

Some TPMs store EK certificates in NV RAM wrapped in an OCTET STRING, which is a known convention per the TCG EK Credential Profile. The agent should unwrap this before sending the certificate to the registrar.

## Error from registrar

```
ValueError: error parsing asn1 value: ParseError { kind: UnexpectedTag { actual: 4 } }
```

Followed by the pyasn1 fallback also failing:

```
pyasn1.error.PyAsn1Error: <TagSet object, tags 0:0:4> not in asn1Spec: <Certificate schema object, ...>
```

## Proposed fix

In `check_ek_cert`, after decoding the raw ASN.1, check whether the outer tag is OCTET STRING (`0x04`). If so, unwrap it and use the inner bytes as the certificate data. This ensures backward compatibility with all registrar versions.

Additionally, it may be worth validating that the final result starts with a SEQUENCE tag (`0x30`) before returning, to catch other unexpected wrapping formats early.

## Environment

- Rust agent: latest version
- Registrar: older version (Python 3.9, keylime installed from packages)
- TPM: hardware TPM that stores EK cert wrapped in OCTET STRING in NV RAM

## Related

A companion issue will be filed against [keylime/keylime](https://github.com/keylime/keylime) to add defensive handling on the registrar side as well.

> This issue was generated with the assistance of Claude Opus 4.6 (Anthropic).

msafarik:
Resolves #1225

## What
Some TPMs store the EK certificate in NVRAM wrapped in an ASN.1 OCTET STRING (tag `0x04`) instead of raw DER (tag`0x30`). `check_ek_cert()` didn't detect or unwrap this, only stripped trailing padding, so the wrapped cert was sent to the registrar as-is and rejected.

## Fix
Detect the OCTET STRING tag before the existing parse/re-encode step, unwrap it, and verify the inner content starts with a SEQUENCE tag. Falls back to the original bytes (with a warning) if anything doesn't look like a validly wrapped cert.

## Testing
Added unit tests covering: wrapped cert gets unwrapped, non-wrapped cert is unaffected, and wrapped-but-invalid-inner-content is passed through as-is.

Companion registrar-side fix: https://github.com/keylime/keylime/pull/1946

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TPM EK certificate handling for certificates stored in ASN.1 OCTET STRING format.
  * Preserved compatibility with certificates provided in the existing raw DER format.
  * Added safeguards for invalid or malformed wrapped certificate data.

### DIFF
--- a/keylime/src/tpm.rs
+++ b/keylime/src/tpm.rs
@@ -611,6 +611,34 @@ fn hash_alg_to_string(hash_alg: TssEsapiHashingAlgorithm) -> Result<String> {
 
 static TPM_CTX: OnceLock<Arc<Mutex<tss_esapi::Context>>> = OnceLock::new();
 
+// Some TPMs store the EK cert in NVRAM wrapped in an ASN.1 OCTET STRING
+// (tag 0x04) instead of raw DER (tag 0x30). Unwrap it here so it can be
+// parsed as a normal certificate.
+fn unwrap_octet_string_ek_cert(cert: &[u8]) -> Vec<u8> {
+    if cert.first() != Some(&0x04) {
+        return cert.to_vec();
+    }
+
+    match picky_asn1_der::from_bytes::<serde_bytes::ByteBuf>(cert) {
+        Ok(inner) if inner.first() == Some(&0x30) => inner.into_vec(),
+        Ok(_) => {
+            warn!(
+                "EK certificate is OCTET-STRING-wrapped, but the inner \
+                 content does not start with a SEQUENCE tag; using it \
+                 as-is"
+            );
+            cert.to_vec()
+        }
+        Err(e) => {
+            warn!(
+                "EK certificate looks OCTET-STRING-wrapped, but could \
+                 not be unwrapped: {e}"
+            );
+            cert.to_vec()
+        }
+    }
+}
+
 impl Context<'_> {
     /// Creates a connection context.
     pub fn new() -> Result<Self> {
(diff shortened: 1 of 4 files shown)


kkaarreell:
I have implemented a functional test in
https://github.com/keylime/rust-keylime/pull/1272
but pls consider adding unit tests as well.

keylime/src/tpm.rs
```diff
@@ -656,8 +684,9 @@ impl Context<'_> {
 
     // Tries to parse the EK certificate and re-encodes it to remove potential padding
     fn check_ek_cert(&mut self, cert: &[u8]) -> Result<Vec<u8>> {
+        let cert = unwrap_octet_string_ek_cert(cert);
```
sergio-correia: Placing the unwrap inside `check_ek_cert` means the unwrapped result is lost on the error path. In `create_ek` (line 754), when `check_ek_cert` returns `Err`, the fallback at line 758 uses `Some(cert)` which is the original raw NVRAM bytes, still OCTET STRING-wrapped.

Would it work better to move the unwrapping to the call site? Remove it from here and instead do it in create_ek around line 753:

```rust
Ok(cert) => {
    let cert = unwrap_octet_string_ek_cert(&cert);
    match self.check_ek_cert(&cert) {
        Ok(cert_checked) => Some(cert_checked),
        Err(_) => {
            warn!("EK certificate in TPM NVRAM is not ASN.1 DER encoded");
            Some(cert)  // now uses the unwrapped cert
        }
    }
}
```

That way `check_ek_cert` stays focused on the DER re-encoding, and the fallback also benefits from the unwrapping.

sergio-correia (commented):
Looks good overall, clean fix. One suggestion about where to place the unwrapping call.

Not in this diff, but related: `split_der_certificates` breaks on any non-`0x30` first byte, so if the same TPM vendors that wrap the EK cert also wrap CA chain certs at NV indices `0x01c00100`-`0x01c001ff`, they would be silently dropped by `read_ek_ca_chain`. Might be worth applying the same `unwrap_octet_string_ek_cert` to each NV index's data as a follow-up.

ansasaki (approved):
LGTM!

PR #1277, opened 2026-09-16
Limit actix-web workers

### LINKED ISSUE #1276: "Too many open files" crash in keylime_agent (pull mode)
Test that failed

  /Tests with IMA sign template/functional/agent-resilience-and-reattestation/pull — all phases after setup fail because the agent crashes immediately on startup.

  What was expected

  The keylime agent starts, listens on port 9002, and accepts attestation requests from the verifier.

  What actually happened

  The agent registers with the registrar, activates, prints "Listening on https://127.0.0.1:9002", then immediately panics with:
  thread 'actix-rt|system:0|arbiter:252' panicked at actix-server/src/worker.rs:428:30:
  called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
  This happens on all 3 restart attempts. The tenant then gets "Connection refused" on port 9002.

  Root cause analysis

  The test machine is an Ampere Altra Max (ampere-mtjade-altramax-01) with 2 sockets × 128 cores = 256 total cores.

  The rust keylime agent (v0.2.10) uses actix-web's HttpServer, which defaults to one worker thread per CPU core (std::thread::available_parallelism()). On this 256-core machine, actix spawns 256 worker threads.
  Each worker creates its own tokio runtime, needing ~5 file descriptors (epoll fd, eventfd, timer fd, etc.). That's ~1280 FDs minimum — exceeding the default soft ulimit of 1024.

  Around worker ~252, the tokio::runtime::Builder::new_current_thread().build() call at worker.rs:428 fails with EMFILE. The .unwrap() panics, which cascades to the main thread and kills the process.

  Why the push variant passes

  The push model agent (keylime_push_model_agent) doesn't create an HttpServer — it pushes attestation to the verifier as a client. No actix workers, no fd exhaustion.

  Not a v0.2.9 → v0.2.10 regression

  The actix-server version (2.6.0), worker spawning code, and systemd unit file are identical between v0.2.9 and v0.2.10. The .workers() method to limit worker count has never been called in the keylime agent
  codebase. This issue would affect any version of rust-keylime on a machine with ~200+ cores.

  No existing upstream issue

  No issue was found in keylime/rust-keylime tracking this problem. The underlying actix bug is documented in actix/actix-net#703 (graceful handling instead of panic), fixed by actix/actix-net#944, but that fix
  only prevents the panic — it doesn't prevent the fd exhaustion.

 The fix would be adding .workers() at line 775, right before .disable_signals():

```
  // line 771 (current)
      .default_service(web::to(errors_handler::app_default))
      })
      // Disable default signal handlers.  See:
      // https://github.com/actix/actix-web/issues/2739
      // for details.
      .disable_signals();
``

  The change:

```
      .default_service(web::to(errors_handler::app_default))
      })
      .workers(4)
      // Disable default signal handlers.  See:
      // https://github.com/actix/actix-web/issues/2739
      // for details.
      .disable_signals();
``

  File: keylime-agent/src/main.rs, line 775 — add .workers(4) in the HttpServer builder chain, before .disable_signals(). The agent only ever serves the verifier pulling quotes, so 4 workers is more than enough.
  You could even argue for 1 or 2, but 4 gives a comfortable margin without wasting resources on a 256-core machine.

  Package versions

  - keylime-agent-rust-0.2.10-2.el10.aarch64
  - keylime-7.14.3-3.el10.aarch64
  - Test branch: rhel-10-main
  - Platform: RHEL-10.4-Nightly / aarch64

kkaarreell:
On machines with many cores (e.g., 256-core Ampere Altra Max), actix-web defaults to one worker per core, exhausting the default 1024 file descriptor ulimit and crashing the agent at startup.

Cap workers at 4, which is more than sufficient for serving verifier quote requests.

Closes: #1276

## Summary by CodeRabbit

* **Performance**
  * Limited the agent’s server worker threads to improve resource usage and predictability.


ansasaki (approved):
LGTM! I think 4 is OK, maybe even too many. Maybe we can improve this in future by auto-detecting if the machine has a low number of CPUs and make it `min(num_cpus, 4)` or configurable, but at the moment I feel it would be an overkill. In general I'd prefer if there was a constant defined instead of using magic numbers, but I think that it wouldn't be used anywhere else, the intention here is clear, and it is well documented.

